PR #261, opened 2017-05-04
Feature #7496

lde-axelor:
Adding Company and Code of year field in kilometricLog grid view


axelor-human-resource/src/main/resources/views/KilometricLog.xml
```diff
@@ -5,6 +5,8 @@
     
     <grid name="kilometric-log-grid" title="Kilometrice log" model="com.axelor.apps.hr.db.KilometricLog">
         <field name="year"/>
```
gdu-axelor: Please, in this case, remove the field year to avoid duplicate informations

PR #594, opened 2017-10-12
fix issue #8417 : charactere * is not authorised in generation of file in Windows

fsc-axelor:


beuss:
I've a fix pending that use a sequence for drafts numbering, is it worth doing a pull request with it or will it be rejected?

Issue #455, opened 2017-08-01
Invalid sales document numbering


beuss:
With bare 4.0.0-RC4 install (tried manual install & bundled installer, same results), creating a new sales order cause it to be numbered as *12 instead of SO00012 according to configuration.

beuss:
Seem to be the expected behavior, draft documents are numbered using *id (soon #id). Will open a PR that use a specific sequence for drafts

PR #616, opened 2017-10-18
Use sequence for drafts instead of hard coded numberin

beuss:
This patch series add a new kind of sequence for draft documents numbering (any kind). This solves issues #455 and all issues related to invalid character. This a different approach from #594 as it does not hardcode anything.


beuss:
Fixed by Axelor in another way

PR #715, opened 2017-11-09
8861 : Removing some fields in sale order / sale order lines template form and set creation date / end of validity date at the creation of the sale order from the template

lde-axelor:


axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -417,6 +412,10 @@
 		<attribute name="hidden" for="toInvoice,allPanelTabs,margin" if="eval: __user__.group.code == 'CLT'" expr="true"/>
```
gdu-axelor: Please, don't use hard coded value.

axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -417,6 +412,10 @@
...
 	</action-attrs>
 
+	<action-attrs name="action-sale-order-line-template-attrs-hidden-fields-for-client">
+		<attribute name="hidden" for="allPanelTabs" if="eval: __user__.group.code == 'CLT'" expr="true"/>
```
gdu-axelor: Please, don't use hard coded value.

PR #973, opened 2017-12-14
RM#9209: Review of address filling

fsc-axelor:
* Add new field ZIP in address
* Add filter on city by ZIP and Country


axelor-base/src/main/resources/views/Address.xml
```diff
@@ -80,6 +84,7 @@
 		<field name="addressL6" expr="eval:city.hasZipOnRight ? city.name+' '+city.zip :city.zip+' '+city.name"/>
 <!--   		<field name="addressL7Country" expr="eval:city.department==null?addressL7Country:city.department.region!=null?city.department.region.country:addressL7Country"/> -->
   		<field name="addressL7Country" expr="eval:city?.country"/>
+		<field name="zip" expr="eval: city.zip"/>
```
gdu-axelor: potential NPE

PR #994, opened 2017-12-18
 RM#9173: Fix ventilate invoice with Company account 

fsc-axelor:


gdu-axelor:
Better way here #1502

PR #1088, opened 2018-01-05
RM9705: payroll-leave-form is now readonly

akh-axelor:


axelor-human-resource/src/main/resources/views/PayrollLeave.xml
```diff
@@ -11,12 +11,12 @@
 	</grid>
 
 	<form name="payroll-leave-form" title="Payroll Leave" model="com.axelor.apps.hr.db.PayrollLeave" cols="12" width="large">
-		<panel name="main">
+		<panel name="main" readonly="true">
 			<field name="fromDate" colSpan="4"/>
 			<field name="toDate" colSpan="4"/>
 			<field name="duration" colSpan="4"/>
 			<field name="leaveReason"/>
```
gdu-axelor: grid-view and form-view should be specified

PR #1108, opened 2018-01-09
Dev #9764 added domain on employeeSet of HRBatch

dda-axelor:


axelor-human-resource/src/main/resources/views/HRBatch.xml
```diff
@@ -11,11 +11,11 @@
 		<field name="description"/>
     </grid>
  
-	<form name="hr-batch-form" title="HR Batch" model="com.axelor.apps.hr.db.HrBatch" width="large" onNew="action-hr-batch-record-default">
+	<form name="hr-batch-form" title="HR Batch" model="com.axelor.apps.hr.db.HrBatch" width="large" onNew="action-hr-batch-record-default,action-hr-batch-attrs-employee-set-domain" onLoad="action-hr-batch-attrs-employee-set-domain">
 	  	<panel name="main">
 		    <field name="actionSelect"/>
 		    <field name="code"/>
-		    <field name="company" canEdit="false"/>
+		    <field name="company" canEdit="false" onChange="action-hr-batch-attrs-employee-set-domain"/>
```
gdu-axelor: Domain should be apply on OnSelect event on the concerned field, here, employeeSet

PR #1197, opened 2018-01-17
RM9949: check for all unique fields on User.

pdo-axelor:


axelor-base/src/main/resources/domains/User.xml
```diff
@@ -35,6 +35,8 @@
    	<boolean name="noHelp" help="Whether to show help messages." massUpdate="true"/>
    	<boolean name="blocked" default="false" help="Specify whether to block the user for an indeterminate times." massUpdate="true"/>
     <string name="homeAction" massUpdate="true"/>
+
+    <finder-method name="findByCode" using="code" filter="LOWER(self.code) = LOWER(:code)"/>
```
beuss: Shouldn't you add an unique key to avoid concurrency issues?

pdo-axelor: What do you mean? The User code field is already defined as unique.

PR #1269, opened 2018-01-24
RM10007: Prevent timesheet closing when we click on 'Confirm & contin…

dtr-axelor:
…ue' while there is an errror.

### DIFF
--- a/axelor-human-resource/src/main/java/com/axelor/apps/hr/web/timesheet/TimesheetController.java
+++ b/axelor-human-resource/src/main/java/com/axelor/apps/hr/web/timesheet/TimesheetController.java
@@ -245,30 +245,31 @@ public class TimesheetController {
 	}
 	
 	//action called when confirming a timesheet. Changing status + Sending mail to Manager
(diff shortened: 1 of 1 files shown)


axelor-human-resource/src/main/java/com/axelor/apps/hr/web/timesheet/TimesheetController.java
```diff
@@ -245,30 +245,31 @@ public void cancel(ActionRequest request, ActionResponse response)  {
...
-			if(message != null && message.getStatusSelect() == MessageRepository.STATUS_SENT)  {
+			if (message != null && message.getStatusSelect() == MessageRepository.STATUS_SENT) {
 				response.setFlash(String.format(I18n.get("Email sent to %s"), Beans.get(MessageServiceBaseImpl.class).getToRecipients(message)));
-			} 
-			
-		}  catch(Exception e)  {
-			TraceBackService.trace(response, e);
-		}
-		finally {
+			} else {
```
gdu-axelor: Seems be strange change. If no email to send mode is configured, we will alert wrongly.

PR #1351, opened 2018-02-05
currencyService: go back as far as one week for conversion rates

beuss:
Sometime, the webservice has no data for a long time (especially after
week ends since stock exchange is closed). Since processing code takes
the last rate, it shouldn' hurt to go back more than one day in the
past.

### DIFF
--- a/axelor-base/src/main/java/com/axelor/apps/base/service/CurrencyConversionService.java
+++ b/axelor-base/src/main/java/com/axelor/apps/base/service/CurrencyConversionService.java
@@ -21,22 +21,12 @@ import java.lang.invoke.MethodHandles;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.net.URL;
-
 import java.time.LocalDate;
-import java.time.Period;
 import java.util.HashMap;
 import java.util.Map;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


beuss:
What would be the benefit ? You'll get the same result by going back directly to one week, you always take the most recent rate anyway

sos-axelor:
Currency rate changes on daily basis. If we jump off by one week we may not get latest rate. So better approach is go back 1 day.

beuss:
From what I saw from my dump it seems that all rates up to the current day are returned.
eg, with "http://sdw-wsrest.ecb.europa.eu/service/data/EXR/D.USD+EUR.EUR.SP00.A?startPeriod=2018-01-20" I got  12 observations with dataSets[0].series["0:0:0:0:0"].observations[11][0] → 1.2329
 with "http://sdw-wsrest.ecb.europa.eu/service/data/EXR/D.USD+EUR.EUR.SP00.A?startPeriod=2018-02-05" I got 2 observations with observations with dataSets[0].series["0:0:0:0:0"].observations[2][0] → 1.2329
Since at a quick glance parsing code always takes the last value this makes absolutely no difference
Did I miss something?
Regards

sos-axelor:
Yes it required endPeriod too. I will create a ticket for that with limit of checking upto one week.

PR #1357, opened 2018-02-07
 Feature #10295: INVOICING PROJECT : form view

cch-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/web/InvoicingProjectController.java
```diff
@@ -52,6 +52,7 @@
 	public void generateInvoice(ActionRequest request, ActionResponse response) throws AxelorException{
 		InvoicingProject invoicingProject = request.getContext().asType(InvoicingProject.class);
 		invoicingProject = invoicingProjectRepo.find(invoicingProject.getId());
+		invoicingProject.setStatusSelect(4);
```
gdu-axelor: Please, never use hard coded value.

sos-axelor: we are already using hard coded values (mostly for select items) on most of xml views.

PR #1380, opened 2018-02-09
account: fix NPE on payer quality generation

beuss:
This should fix an NPE when clicking on “Compute payers quality” in app account configuration

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/web/AppAccountController.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/web/AppAccountController.java
@@ -21,31 +21,28 @@ package com.axelor.apps.account.web;
 import com.axelor.apps.account.service.app.AppAccountService;
 import com.axelor.apps.account.service.debtrecovery.PayerQualityService;
 import com.axelor.exception.service.TraceBackService;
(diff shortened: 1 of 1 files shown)


gdu-axelor:
Better to do conditionnal injection in the Controller to avoid to inject useless class. That's can be done using provider or injector.

We can easily use injector.getInstance() with Beans.get(PayerQualityService.class).process(...)

beuss:
Well, if we're at that level of "optimisation", the first thing to do would be to @Singelotin-ize everything that can be in the whole application :-)

gdu-axelor:
Yes, I agree, but all changes take time. :)

In fact, we know that we can really bad degradation of performance if in the controller we have a simple action and we have to inject some useless services. Injection is done in cascade so, as we have some dependencies on services, that can be boring. Maybe not the case here, but as you are fixing the issue, better to avoid another later.

beuss:
Entirely agree and added a note on my screen with this tip.
Glad to see that devs are able to take some time to look at third-parties pull requests :-)
I'm almost sure @Singleton would solve most of the performance issues (delta between a non-singleton vs singleton SequenceServer was bigger than huge)
Tanks for merging anyway :)

PR #1389, opened 2018-02-12
user: hide contact info pane if new

beuss:
This fixes a "Name cannot be null" error when attempting to fill details upon creation


beuss:
Hi,

Could someone please take a look and merge or close?

Regards

gdu-axelor:
Hi,
I merge it for now but we have to add new way to create user (with a wizard) because the current way is not good in particular for contact informations.
Regards

PR #1431, opened 2018-02-24
Fix StockMove cancellation

beuss:
Stock move were not cancellable because of wrong filtering on cancel reasons due to class name mangling.


gdu-axelor:
Closed in favor of e37fc287926cc1f0c6672fc951452ad081eb0f0c. Better to use _model instead of getClass() method.

PR #1499, opened 2018-03-12
Redmine 10922: PurchaseOrder, add boolean in Supplychain config, and add button to finish manually the purchase order when status is on validated

man-axelor:


axelor-purchase/src/main/resources/i18n/messages_fr.csv
```diff
@@ -65,6 +65,7 @@
 "Draft","Brouillon",,
 "Estim. delivery date","Date de livraison estimée",,
 "Financial terms","Termes financiers",,
+"Finish manually purchase order","Terminer manuellement les commandes fournisseurs",,
```
gdu-axelor: This term is not good. This message appears on the button on each purchase order, so just put "Finish" (Terminer) or "Finish the order" (Terminer la commande)

PR #1502, opened 2018-03-13
accounting: correctly fallback to global params if partner is empty

beuss:
AccountingSituation::get{Customer,Supplier,Employee}Account should never
be used directly except in AccountingSituationService, in order to
correctly fallback to accounts defined on company level if there's no
account defined on specific partner.


gdu-axelor:
Thanks for you analysis and your contribution !

PR #1515, opened 2018-03-15
Full rewrite of PrestaShop module

beuss:
Full rewrite of existing module. Currently work in progress (only order import left to implement).
Will need some polishing but I'd like reviews to start right now if needed


beuss:
Woutch, as I said it wasn't ready to be merged :D

gdu-axelor:
Will be moved to the new repository "axelor-addons" very soon. So I was thinking it was better to merge it before extract this module.

beuss:
Oh good point.
How will you handle third party modules that needs to modify core views ? eg. I was thinking about adding a mention on products/customers/orders imported from PrestaShop. This will need some synchronization between core repository and third party

gdu-axelor:
Depends of cases. We can override the view (and use panel-include) in addons modules but with a limit : we can't have multiple inheritence of view.
For your case (mention on some objects), perhaps you can fill the both field dedicated for import (importId and importOrigin).

beuss:
Generaly speaking, would you accept patches for core views with appropriate `if`/`if-module`?

PR #1535, opened 2018-03-20
RM11187: Improve StockMove UI.

akh-axelor:


axelor-base/src/main/java/com/axelor/apps/base/exceptions/IExceptionMessage.java
```diff
@@ -50,7 +50,7 @@
 	 */
 
 	static final String UNIT_CONVERSION_1 = /*$$(*/ "Please configure unit conversion from '%s' to '%s'." /*)*/ ;
-	static final String UNIT_CONVERSION_2 = /*$$(*/ "Please configure units." /*)*/ ;
+	static final String UNIT_CONVERSION_2 = /*$$(*/ "Please configure weight units for this product packaging." /*)*/ ;
```
gdu-axelor: This exception is used for any unit conversion, so shouln't be changed. You can define a new exception for specific issue before call the unit conversion class.

PR #1549, opened 2018-03-22
Anomaly #11194: TICKET : demo data

cch-axelor:


axelor-crm/src/main/resources/data-init/input/crm_ticketCategory.csv
```diff
@@ -1,8 +1 @@
 importId;name
```
gdu-axelor: CRM ticket doesn"t exist anymore, so we can remove this file on crm module and move it to helpdesk TicketType.

PR #1584, opened 2018-04-03
rm11374: add message panel dashlet in DebtRecovery

fsc-axelor:


axelor-message/src/main/resources/views/Message.xml
```diff
@@ -131,5 +131,13 @@
 	<action-validate name="action-validate-message-content-onsave">
 		<alert message="No Mail content entered." if="mediaTypeSelect == 1 &amp;&amp; content== null"/>
 	</action-validate>
-	
+
+	<action-view name="action.message.view.related.to.first" model="com.axelor.apps.message.db.Message" title="Messages">
```
gdu-axelor: Should be factored with :
https://github.com/axelor/axelor-business-suite/blob/wip/axelor-account/src/main/resources/views/Invoice.xml#L1063

PR #1598, opened 2018-04-04
Redmine 11387: Batch, Fix batch null unknown when no code entered in …

man-axelor:
…MailBatch, remove BaseBatch and its entry menu, because it is not used and the batch called an undefined method in BaseBatchController, and what was wanted in BaseBatch is in CRM batch


axelor-base/src/main/java/com/axelor/apps/base/web/MailBatchController.java
```diff
@@ -41,16 +42,19 @@
...
 
 	public void remindTimesheet(ActionRequest request, ActionResponse response) throws AxelorException{
+	    try{
+	        MailBatch mailBatch = request.getContext().asType(MailBatch.class);
 
-		MailBatch mailBatch = request.getContext().asType(MailBatch.class);
+	        Batch batch = null;
 
-		Batch batch = null;
+	        batch = mailBatchService.remindMail(mailBatchRepo.find(mailBatch.getId()));
```
gdu-axelor: Why use two lines for assign batch ?

axelor-base/src/main/resources/domains/BaseBatch.xml
```diff
@@ -1,21 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<domain-models xmlns="http://axelor.com/xml/ns/domain-models"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://axelor.com/xml/ns/domain-models http://axelor.com/xml/ns/domain-models/domain-models_5.0.xsd">
-
-  <module name="base" package="com.axelor.apps.base.db"/>
-
-  <entity name="BaseBatch" lang="java">
```
gdu-axelor: Why remove this object ?

axelor-base/src/main/resources/domains/MailBatch.xml
```diff
@@ -8,7 +8,7 @@
   <entity name="MailBatch" lang="java">
   	
   	<!-- HEADER -->
-  	<string name="code" title="Code" namecolumn="true" unique="true"/>
+    <string name="code" title="Code" namecolumn="true" unique="true" required="true"/>
```
gdu-axelor: Code should be able to be null if the batch is run from UI. Only required if the batch is called per WS from a scheduler.

axelor-base/src/main/resources/domains/BaseBatch.xml
man-axelor: Changes were made, now we don't remove the object, the undefined method actionTarget is now defined, code is not required.

PR #1613, opened 2018-04-07
sales: don't use Exception on workflow method

beuss:
Remove throws Exception as it is actually never used, only AxelorException is thrown. Avoid
having a too wide throws declaration to leverage calling code structure (needed for axelor/axelor-addons#7)

### DIFF
--- a/axelor-production/src/main/java/com/axelor/apps/production/service/SaleOrderWorkflowServiceProductionImpl.java
+++ b/axelor-production/src/main/java/com/axelor/apps/production/service/SaleOrderWorkflowServiceProductionImpl.java
@@ -34,36 +34,35 @@ import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 
 public class SaleOrderWorkflowServiceProductionImpl extends SaleOrderWorkflowServiceSupplychainImpl {
-	
+
 	protected ProductionOrderSaleOrderService productionOrderSaleOrderService;
 	protected AppProductionService appProductionService;
-	
+
 	@Inject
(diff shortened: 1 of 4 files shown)


gdu-axelor:
I'm not sure about Rollback on transationnal method.
If we specify a rollbackOn AxelorException, so only this kind of exception will call the rollback.

https://github.com/google/guice/wiki/Transactions#responding-to-exceptions

> **If a transactional method encounters an unchecked exception (any kind of RuntimeException), the transaction will be rolled back. Checked exceptions are ignored and the transaction will be committed anyway.**
> 
> To change this behavior, you can specify your own exceptions (checked or unchecked) on a per-method basis:
> 
>   @Transactional(rollbackOn = IOException.class)
>   public void myMethod() throws IOException { ... }
> **Once you specify a rollbackOn clause, only the given exceptions and their subclasses will be considered for rollback. Everything else will be committed.** Note that you can specify any combination of exceptions using array literal syntax:
> 
>   @Transactional(rollbackOn = { IOException.class, RuntimeException.class, ... })
> It is sometimes necessary to have some general exception types you want to rollback but particular subtypes that are still allowed. This is also possible using the ignore clause:
> 
>   @Transactional(rollbackOn = IOException.class, ignore = FileNotFoundException.class)
> In the above case, any IOException (and any of its subtypes) except FileNotFoundException will trigger a rollback. In the case of FileNotFoundException, a commit will be performed and the exception propagated to the caller anyway.
> 
> Note that you can specify any combination of checked or unchecked exceptions.
> 

So we such have two cases : 
- No checked exception : @Transactionnal
- There is a checked exception : @Transactionnal(rollbackOn = { all checked exceptions , Exception.class })

beuss:
Pushed fixed patch

PR #1614, opened 2018-04-07
account: add method to get CompanyBankDetails from Partner

beuss:
This handle defaulting to company one if needed (needed for axelor/axelor-addons#7)


gdu-axelor:
We will refactor the action-record called on onNew for better maintenability.

PR #1620, opened 2018-04-09
partner/lead/employee: improve website regex

beuss:
Regex wasn't allowing two parts websites (http://xxx.yy), improve it by
forcing http:// prefix and leveraging overall check.


beuss:
Pushed to dev in order to have a little hope to get this processed

gdu-axelor:
Seems heavy to have to fill the http:// prefix. Semes better to allow two parts websites with http:// prefix and three part with or without http://. What do you think about this?

beuss:
Fixed to allow any number of parts (from one to infinite) and make prefix optional.

PR #1623, opened 2018-04-09
reports: properly handle case where no logo is filled

beuss:
Avoid having text complaining about missing resource, instead, hide it
if logo hasn't been filled in company.


axelor-account/src/main/resources/reports/AccountMove.rptdesign
```diff
@@ -886,6 +886,12 @@ where move.id IN (?);]]></xml-property>
                             <property name="overflow">hidden</property>
                             <property name="height">71px</property>
                             <property name="width">309px</property>
+                            <list-property name="visibility">
+                                <structure>
+                                    <property name="format">all</property>
+                                    <expression name="valueExpr" type="javascript">row["logo_path"]</expression>
```
gdu-axelor: Expression seems strange because the goal is to hide the company logo only if the company logo is null. Can you check ?

beuss:
Nice catch, refreshed branch

PR #1627, opened 2018-04-10
invoice: show client box on sales invoices

beuss:
Honor the clientBox parameter existing in accounting configuration.
Displays is the same as for sale orders

### DIFF
--- a/axelor-account/src/main/resources/reports/Invoice.rptdesign
+++ b/axelor-account/src/main/resources/reports/Invoice.rptdesign
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <report xmlns="http://www.eclipse.org/birt/2005/design" version="3.2.23" id="1">
-    <property name="createdBy">Eclipse BIRT Designer Version 4.6.0.v201606072122</property>
+    <property name="createdBy">Eclipse BIRT Designer Version 4.7.0.v201706222054</property>
     <list-property name="propertyBindings">
         <structure>
             <property name="name">queryText</property>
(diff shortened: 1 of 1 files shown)


axelor-account/src/main/resources/reports/Invoice.rptdesign
```diff
@@ -4870,6 +4870,46 @@ row["partnerFirstName"]
                                 </cell>
                                 <cell id="5115"/>
                             </row>
+                            <row id="5134">
+                                <property name="height">0.6085416666666666cm</property>
+                                <list-property name="visibility">
+                                    <structure>
+                                        <property name="format">all</property>
+                                        <expression name="valueExpr" type="javascript">row["ClientBox"] == null || row["ClientBox"] == "" || row["operation_type_select"]</expression>
```
gdu-axelor: The expression seems strange, can you check ?

axelor-account/src/main/resources/reports/Invoice.rptdesign
```diff
@@ -4870,6 +4870,46 @@ row["partnerFirstName"]
...
+                                <cell id="5140"/>
+                                <cell id="5141"/>
+                                <cell id="5142"/>
+                                <cell id="5143"/>
+                            </row>
+                            <row id="5124">
+                                <list-property name="visibility">
+                                    <structure>
+                                        <property name="format">all</property>
+                                        <expression name="valueExpr" type="javascript">row["ClientBox"] == null || row["ClientBox"] == "" || row["operation_type_select"] != 3</expression>
```
gdu-axelor: Better to display this field for customer invoice and customer refund.

beuss:
Just fixed review
As a side note, I don't know what's the intended use for client box on invoice but it's likely that having it configurable on a per-invoice type basis would be a good thing

PR #1628, opened 2018-04-10
sale order: properly hide "delivery address" title

beuss:
The conditions were rather strange, moreover deliveryAddress seems to be
an empty string rather than a null one when not filled.


beuss:
Just a reminder on this one as I've potential subsequent patches on the template :)

beuss:
Will do another PR with more fixes

PR #1635, opened 2018-04-11
Anomaly #11470 Fix: Hiding total(exTaxTotal or inTaxTotal) based on '…

aga-axelor:
…inAti' on sale and purchase orderline


axelor-purchase/src/main/resources/views/PurchaseOrderLine.xml
```diff
@@ -243,6 +243,8 @@
 	<action-attrs name="action-purchase-order-attrs-hide-ati">
 		<attribute name="hidden" for="exTaxTotal" expr="eval:_parent?.inAti" if="_parent?._model == 'com.axelor.apps.purchase.db.PurchaseOrder'"/>
 		<attribute name="hidden" for="inTaxTotal" expr="eval:!_parent?.inAti" if="_parent?._model == 'com.axelor.apps.purchase.db.PurchaseOrder'"/>
+		<attribute name="hidden" for="exTaxTotal" expr="eval: purchaseOrder?.inAti" if="purchaseOrder != null &amp;&amp; !_parent" />
```
sos-axelor: It will not work if '_parent' is there but its not a purchase order. Fix the condition.

PR #1637, opened 2018-04-12
Anomaly #11332: PRODUCT FAMILY : demo data

ugo-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/script/ImportPurchaseOrder.java
```diff
@@ -0,0 +1,54 @@
...
+	public Object importPurchaseOrder(Object bean, Map<String,Object> values) throws AxelorException {
+		assert bean instanceof PurchaseOrder;
+		
+		PurchaseOrder purchaseOrder = (PurchaseOrder) bean;
+		
+		if (purchaseOrder.getStatusSelect() == 1 || purchaseOrder.getStatusSelect() == 2) {
+			purchaseOrder.setPurchaseOrderSeq(sequenceService.getDraftSequenceNumber(purchaseOrder));
+		}
+		
+		purchaseOrder = purchaseOrderRepository.find(purchaseOrder.getId());
```
sos-axelor: Why you fetched managed instance of PO, and is there meaning of setting sequence in previous line if you  fetch on next line ?

PR #1781, opened 2018-05-07
RM11986: Add a synchronization option on calendars to choose which ty…

akh-axelor:
…pe of event to create when synchronizing with an external calendar.


axelor-base/src/main/resources/domains/ICalendar.xml
```diff
@@ -30,8 +31,12 @@
 			public static final int GCAL = 3;
 			public static final int ZIMBRA = 4;
 			public static final int KMS = 5;
-			public  static final int CGP = 6;
+			public static final int CGP = 6;
 			public static final int CHANDLER = 7;
+
+			// SYNCHRONISATION SELECT
+			public static final String ICAL_ONLY = "ICalEvent";
+			public static final String CRM_SYNCHRO = "Event";
```
gdu-axelor: Should be moved in CRM module

PR #1820, opened 2018-05-14
Anomaly #10150 PeriodService: while loop

ego-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/BudgetController.java
```diff
@@ -45,7 +46,7 @@ public void updateLines(ActionRequest request, ActionResponse response){
 		response.setValue("budgetLineList", budgetLineList);
 	}
 	
-	public void generatePeriods(ActionRequest request, ActionResponse response) {
+	public void generatePeriods(ActionRequest request, ActionResponse response) throws AxelorException {
```
gdu-axelor: Should have a Try catch statement here and store the exception using TraceBackService.

axelor-account/src/main/java/com/axelor/apps/account/web/YearController.java
```diff
@@ -40,7 +41,7 @@ public void close(ActionRequest request, ActionResponse response) {
 		catch(Exception e)  { TraceBackService.trace(response, e); }	
 	}
 	
-	public void generatePeriods(ActionRequest request, ActionResponse response) {
+	public void generatePeriods(ActionRequest request, ActionResponse response) throws AxelorException {
```
gdu-axelor: Should have a Try catch statement here and store the exception using TraceBackService.

PR #1842, opened 2018-05-17
Redmine 12156: Add missing sequence for analytic report in demo data,…

man-axelor:
… add missing trace in save of AccountingReport (transaction is closed, then trace is runInTransaction, then initial transaction is open)


axelor-account/src/main/java/com/axelor/apps/account/db/repo/AccountingReportManagementRepository.java
```diff
@@ -40,6 +42,9 @@ public AccountingReport save(AccountingReport accountingReport) {
 			
 			return super.save(accountingReport);
 		} catch (Exception e) {
+            JPA.em().getTransaction().commit();
+            JPA.runInTransaction(() -> TraceBackService.trace(e));
+            JPA.em().getTransaction().begin();
```
gdu-axelor: Strange

We should start with JPA.em().getTransaction().begin();
and finish per a commit.

Else we will don't manage the rollback

beuss: I guess the goal is to have a separate transaction for the traceback to ensure it is correctly saved, but since this is an exception handler, shouldn't this be rollback() (cancel current), runInTransaction (create and commit a new one), begin (create a new one)?

PR #1870, opened 2018-05-25
Anomaly #12133 SALEORDERLINE : BOM

ego-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -83,24 +85,35 @@ public void updateProductCostPrice(BillOfMaterial billOfMaterial) throws AxelorE
...
+        personalizedBOMSet.add(customizeBillOfMaterial(childBillOfMaterial));
+      }
+      personalizedBOM.setBillOfMaterialSet(personalizedBOMSet);
 
-	}
+      return personalizedBOM;
+    }
+
+    return null;
+  }
```
gdu-axelor: Please add a break condition to avoid to have an infinite loop if there is a cycle (max = 1000)

PR #1880, opened 2018-05-27
purchases: use default product info when not in catalog

beuss:
Use default product name and purchase price on purchase order line when
selected product is not present in supplier's catalog. This improves
workflow for generic "one-shot" products


beuss:
Rebased on dev to get a chance to see this merged

gdu-axelor:
I'm not sure about this change. Perhaps should be an option. When we create a RFQ, we don't want use the default purchase price but we ask supplier to provide it. I think that for "one-shot" products, we have to enter directly the unit price on the line.

beuss:
Didn't thought about that use case indeed.
Let apart the product description that can be kept in any case
For purchase price there in fact more case that I thought of first:

- We're in a RFQ, price should be left empty… do we have a way to distinguish between RFQ & order?
- Product was never bought from this supplier, we use default purchase price (for my understanding, what is its use otherwise?)
- Product was already bought from this supplier, use the last purchase price?

gdu-axelor:
I'm agree with the logic you explain. Also, now, on wip branc, we have added a boolean to filter on supplier catalog. So if we use another product, we can get the general purchase price, code, and name.

I will merge it soon.

beuss:
Seems that the alternate fix has been merged to dev, if so @gdu-axelor could you confirm & close?

PR #1886, opened 2018-05-29
RM12291: Fix and optimize Validate and Ventilate processus

fsc-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -286,7 +286,7 @@
             <field name="amountRemaining" hidden="true"/>
             <field name="hasPendingPayments" hidden="true"/>
             <button name="validate" hidden="true" title="Validate" onClick="save,action-invoice-method-validate"/>
-            <button name="ventilate" hidden="true" title="Ventilate" showIf="statusSelect == 2 &amp;&amp; operationSubTypeSelect != 2" onClick="save,action-invoice-method-ventilate,save" prompt="Are you sure you want to ventilate the invoice ?" readonly="true"/>
+            <button name="ventilate" hidden="true" title="Ventilate" showIf="statusSelect == 2 &amp;&amp; operationSubTypeSelect != 2" onClick="action-invoice-validate-check-invoice-date-for-ventilate,save,action-invoice-method-ventilate,save" prompt="Are you sure you want to ventilate the invoice ?"/>
```
gdu-axelor: Better to manage this case in ventilate java method. If we manage it only in view, we can have an invoice in futur if we ventilate it in mass.

PR #1889, opened 2018-05-30
Feature #12347 SALEORDER : printing

ego-axelor:


axelor-sale/src/main/resources/views/AppSale.xml
```diff
@@ -3,10 +3,11 @@
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://axelor.com/xml/ns/object-views http://axelor.com/xml/ns/object-views/object-views_5.0.xsd">
 	
-	<form name="app-sale-config-form" title="App sale" model="com.axelor.apps.base.db.AppSale" canDelete="false" canNew="false" width="large">
+	<form name="app-sale-config-form" title="App sale" model="com.axelor.apps.base.db.AppSale" canDelete="false" canNew="false" width="large" onLoad="action-app-sale-attrs-readonly-manage-sale-order-version-onchange">
```
gdu-axelor: This action-attrs that update a value shouldn't be called on onLoad.
You have to separate into two actions, one for value, another one for readonly (or better, use a javascript expression)

axelor-sale/src/main/resources/views/AppSale.xml
```diff
@@ -25,5 +26,9 @@
 		<view type="grid" name="sale-config-grid" />
 		<view type="form" name="sale-config-form"/>
 	</action-view>
-	
+
+  <action-attrs name="action-app-sale-attrs-manage-sale-order-version-onchange">
+    <attribute name="value" for="generatePdfDuringSaleOrderFinalization" expr="true"/>
```
gdu-axelor: better to define an expression using manageSaleORderVersion boolean, instead of "true". Else, the boolean generatePdfDuringSaleOrderFinalization will never be reseted. --> printingOnSOFinalization ?

Also, use a shorter word for this new boolean to avoid database index issue or URL issue due to too long term.

ego-axelor: done

PR #1890, opened 2018-05-30
dev 12301 : set opportunity field to be modifiable when the order is …

wde-axelor:
…a draft


gdu-axelor:
ChangeLog file update is missing

PR #1899, opened 2018-05-31
Allow to refresh ventilated invoice printed copy

### LINKED ISSUE #1895: Consistency issue in ventilated invoice printing
When printing a ventilated invoice, if there's an existing copy, it is fetched instead of making a "real" print. I understand that this is to prevent any unwanted modification on the document sent to eg. customer. But the UI still allows to modify some fields (notes, payment mode, …), which is fine since there's no reason to lock those "auxiliary" fields. But updated values can never be printed.

beuss:
Add a menu entry to regenerate PDF as some fields of the invoice can still be modified.

close #1895


beuss:
Rebased on dev to get a chance to see this merged

PR #1902, opened 2018-05-31
expense: fix translation handling

beuss:
All translated fields were checking row['message_key'] != "" whereas
its value is null when no translation is present. This was causing
missing elements in the report, especially the "kilometric allow param",
when in foreign language (eg. french).
Moreover fix a typo in Kilometric Allowance preventing the title to be
properly printed on report.
The kilometric_allow_param can be translated in the report which is
strange since these are user defined strings, but leave it this way as
changing it might trigger https://xkcd.com/1172/


gdu-axelor:
We can translate field values (ex : product name). If the field is not defined like this we will do it.

PR #1908, opened 2018-06-04
Anomaly #12409 EXPENSE : printing

ego-axelor:


axelor-human-resource/src/main/resources/reports/Expense.rptdesign
```diff
@@ -1959,7 +2019,7 @@ WHERE ExpenseLine.kilometric_expense = ?
                             </structure>
                         </list-property>
                         <property name="contentType">html</property>
-                        <text-property name="content"><![CDATA[<VALUE-OF> if (row["translation"] != null){ row["translation"]} else { row._outer["company_cb_select"] }</VALUE-OF>]]></text-property>
+                        <text-property name="content"><![CDATA[<VALUE-OF> if (row["translation"] != ""){ row["translation"]} else { row["message_key"] }</VALUE-OF>]]></text-property>
```
gdu-axelor: Are you sure that this condition is correct, because this condition has been recently updated for fix the value displayed in case of missing translation : https://github.com/axelor/axelor-business-suite/commit/4c337e7990ba92b44e468b3c2315384825c134fd

beuss: It seems the whole patch has been reverted on this one

gdu-axelor: Also, @ego-axelor, please explain your changes in changeLog file and in commit comment. Here, the case is to add the currency, and translate the value of the kilometric log using translatable field.

ego-axelor: done

PR #1924, opened 2018-06-07
RM#11989 Sale order form : Fix display timetable tab

aga-axelor:


sos-axelor:
No need of the change.

PR #1950, opened 2018-06-11
RM#12437: Fix field in project task tree

fsc-axelor:


axelor-project/src/main/resources/views/Project.xml
```diff
@@ -312,7 +312,7 @@
 	<tree name="project-task-tree" title="Tasks">
     	
     	<column name="name" type="string"/>
-    	<column name="assignedTo" type="reference"/>
+    	<column name="assignedTo" type="string"/>
```
gdu-axelor: Please check the ticket  : https://redmine.axelor.com/issues/11089

For reference fields (M2O, O2O), better to let the system use the fullname of the object. The issue has been fixed on ADK so now, no need to specify the dotted field in the node.

PR #1980, opened 2018-06-14
Redmine 12597: Sale order, show stock move instead of stock move line…

man-axelor:
… in CustomerDeliveries


gdu-axelor:
We need to know  the delivered qty per product.

PR #2008, opened 2018-06-18
Redmine 12417: Balance on top of Partner view become configurable for…

man-axelor:
… use specific balance


axelor-account/src/main/resources/domains/AppAccount.xml
```diff
@@ -21,6 +21,7 @@
 
 		<boolean name="printReportOnVentilation" title="Print report on invoice ventilation" default="true" />
 
+        <integer name="balanceConfiguration" title="Balance configuration" selection="account.general.balance.configuration.type.select" default="1"/>
```
gdu-axelor: Selection take the same words like the field name, so here to respect the naming convention and your selection definition, you can define field like : balanceConfigurationTypeSelect

But, this name is not enough clear.

You must indicate that this field refer to the partner balance viewer not general balance configuration.

PR #2025, opened 2018-06-19
Product: show accounting tab only if Accounting app is enabled.

pdo-axelor:


gdu-axelor:
No, We need this tab for define Tax that will applied on Sale order and Purchase order

PR #2061, opened 2018-06-26
RM12487: add mail subject for follower mails

fsc-axelor:
* Fix reference in mail header for google newsfeed.
* Remove block div from google.


axelor-helpdesk/build.gradle
```diff
@@ -19,4 +19,5 @@ axelor {
 
 dependencies {
 	compile project(":modules:axelor-base")
+	compile 'org.jsoup:jsoup:1.11.3'
```
gdu-axelor: jsoup 1.10 is already add on ADK, so no need to add a new dependency.

PR #2084, opened 2018-06-29
Remove useless dependencies in build.gradle files

dtr-axelor:


gdu-axelor:
Should be done on dev branch

gdu-axelor:
change log is not up to date

PR #2090, opened 2018-07-02
RM 12731: added translation and made product required when the prod p…

wde-axelor:
…rocess is only available for the product


gdu-axelor:
change log ?

Issue #2071, opened 2018-06-27
Inconsistent stockLocationLine quantities


beuss:
I do not know how I ended up in this situation but if stockLocationLine quantities are inconsistent (because of late Stock module activation? because of exception thrown at wrong moment?), there is no way to go out of this situation.
Application should provide a batch allowing to recompute some or all cached quantities.

gdu-axelor:
Yes it can be a good things. In the same way we have done a batch to recompute the partner balances using account move lines.

beuss:
OK, I can take over this one (if you can assign it), I'm doing it for the prestashop addon anyway :)

PR #2134, opened 2018-07-09
stock: add batch to recompute StockLocationLine entries

beuss:
Preliminary work, add batch structure and logic for non-detail stock
locations. We'll have to move this into supplychain module to handle
reservedQuantity correctly.

ref #2071


beuss:
Please @gdu-axelor or @pdo-axelor  take a look at this preliminary work as all business rules are extrapolated from source code. I guess I should move it to supplychain module as I can't access to ReservedQty otherwise.
By the way, could someone explain me what reservedQty is meant to be? The whole stock handling code is quite opaque (especially with its two parameters future & current where current is always passed as true from what I saw), how is the moveLine reservedQty value set?

pdo-axelor:
Are you sure you need to use native queries? It might cause portability issues.
Also, there might be memory issues with big result lists. It's better to use limit and/or offset on queries.
Besides, batches should update done and anomaly counts.

beuss:
I'll try to rewrite queries in HQL, native is the result of some tries I had previously, shouldn't be an issue.
The main point is to check that computations are correct before optimizing memory & report anomalies/done.

beuss:
I removed the native queries & used a windowed system. Could you check that I use correct information for recalculation so I could finish it? (I'll add success/anomalies count at the end)

PR #2153, opened 2018-07-11
account: improve per partner account creation

beuss:
Improve UX on partner account creation (employee/supplier/customer) by
filling fileds with suitable defaut values (configured partner account
depending on type, account type, parent, etc.)


gdu-axelor:
Please, update the changelog file with this new improvment on the current version.

beuss:
Done (sorry for the delay)

PR #2192, opened 2018-07-19
RM 13189: supply method of a sale order line generated from a pack is …

wde-axelor:
…now correctly set


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
```diff
@@ -429,4 +432,42 @@ public void checkMultipleQty(SaleOrderLine saleOrderLine, ActionResponse respons
...
+          price = (BigDecimal) discounts.get("price");
+        }
+      }
+      subLine.setPrice(price);
+
+      subLines.add(subLine);
+    }
+
+    return subLines;
+  }
```
gdu-axelor: Please, add a new method to create a saleOrderLine in order to override it in supplychain module to fill the specific fields, instead of override completely the parent method.

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderLineServiceSupplyChainImpl.java
```diff
@@ -211,4 +214,37 @@ public void updateDeliveryState(SaleOrderLine saleOrderLine) {
...
+      throws AxelorException {
+    SaleOrderLine subLine = new SaleOrderLine();
+    Product subProduct = packLine.getProduct();
+    subLine.setProduct(subProduct);
+    subLine.setProductName(subProduct.getName());
+    subLine.setPrice(subProduct.getSalePrice());
+    subLine.setUnit(this.getSaleUnit(subLine));
+    subLine.setQty(new BigDecimal(packLine.getQuantity()));
+    subLine.setCompanyCostPrice(this.getCompanyCostPrice(saleOrder, subLine));
+    subLine.setSaleSupplySelect(subProduct.getSaleSupplySelect());
```
gdu-axelor: Here, I assume that only saleSupplySelect assignation is added, so better to call the super method and add only this assignation to avoid duplicate code.

PR #2194, opened 2018-07-19
dev : added FR marketing demo data

dda-axelor:


gdu-axelor:
Changelog is missing

dda-axelor:
done

PR #2224, opened 2018-07-30
account: slightly improve account creation from partner

beuss:
Addendum to commit 22d3e31, use partner name as default name for created
account.


gdu-axelor:
Please, update the changelog file for any change

beuss:
Hi,
Is this necessary as this is only improving a commit with an existing entry in the changelog?
Regards

gdu-axelor:
Hi,

If there is an existing entry for the same version, no need to update the changlog file. Else, you can add it in improvment part of the current version.

Regards

beuss:
Done :-)

PR #2246, opened 2018-08-04
invoice: set journal & partner account on validation

beuss:
Setting it on ventilation implies that user cannot check if they're
correct or not before unreversibly registering invoice into accounting.

By the way, remove JournalService which is pretty useless and was
(uselessly) misused.


beuss:
Superseded by #2248

PR #2274, opened 2018-08-13
Anomaly #11330 SALEORDERLINE : service

ego-axelor:


axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -384,6 +386,11 @@
     <action-record name="action-sale-order-line-record-progress" model="com.axelor.apps.sale.db.SaleOrderLine">
     	<field name="progress" expr="eval: project?.progress"/>
     </action-record>
+
```
gdu-axelor: Better to use action-attrs directly instead of using action-record and client side expression (js)

PR #2277, opened 2018-08-14
Anomaly #12862 TIMESHEET : empty chart in manuf order mode

ego-axelor:


axelor-business-production/src/main/resources/domains/Employee.xml
```diff
@@ -6,11 +6,8 @@
     <module name="human-resource" package="com.axelor.apps.hr.db"/>
 
     <entity name="Employee" cachable="true">
-        <integer name="timesheetImputationSelect" title="Timesheet imputed on" selection="business.production.employee.timesheet.imputation.select" default="1"/>
-
         <extra-code>
             <![CDATA[
-                public static final int TIMESHEET_PROJECT = 1;
```
gdu-axelor: Why have you moved this value ?

This selection is used only in business-production module.

On human-resources module, we don't have manuf order, so we always take into account the project only

ego-axelor: I put it in the human resource module, when the business-production module is not loaded then the selection only have one value TIMESHEET_PROJECT = 1 and a second value is added when the module is loaded.

PR #2322, opened 2018-08-30
Feature RM#13557 Tracking Number : form view

rch-axelor:


axelor-stock/src/main/resources/views/StockMoveLine.xml
```diff
@@ -36,7 +36,8 @@
         <field name="trackingNumber"
             form-view="tracking-number-stock-move-line-form"
             grid-view="tracking-number-grid"
-            onSelect="action-stock-move-line-attrs-tracking-number-domain"/>
+            onSelect="action-stock-move-line-attrs-tracking-number-domain" 
+            onChange="action-tracking-number-stock-move-line-group-action"/>
```
gdu-axelor: Why action-tracking-number-stock-move-line-group-action is called in StockMoveLine view ?

PR #2324, opened 2018-08-30
Add tool in studio to map relevant widgets per type and vice-versa

dtr-axelor:


dtr-axelor:
I created that one because I thought it may be useful for this issue: https://redmine.axelor.com/issues/13416

It may also be useful for configurator creators in Manufacturing app when we add attributes, and in that case I think this tool should be moved into an other module (axelor-tool ?).

Finally, it may be useful for @wde-axelor for his demo-project.

PR #2339, opened 2018-09-04
Feature #13601 PRODPROCESSLINE : duration

ego-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/OperationOrderWorkflowService.java
```diff
@@ -457,24 +456,19 @@ public LocalDateTime computePlannedEndDateT(OperationOrder operationOrder)
...
     ProdProcessLine prodProcessLine = operationOrder.getProdProcessLine();
     WorkCenter workCenter = prodProcessLine.getWorkCenter();
 
     long duration = 0;
+    
+    BigDecimal maxCapacityPerCycle = prodProcessLine.getMaxCapacityPerCycle();
+
+    BigDecimal nbCycles;
+    if (maxCapacityPerCycle.compareTo(BigDecimal.ZERO) == 0) {
+  	  nbCycles = new BigDecimal(1);
```
gdu-axelor: BigDecimal.ONE

axelor-production/src/main/java/com/axelor/apps/production/service/OperationOrderWorkflowService.java
```diff
@@ -489,44 +483,13 @@ public long computeMachineDuration(OperationOrder operationOrder, BigDecimal qty
...
-        duration += qty.multiply(durationPerCycle).longValue();
-      } else {
-        duration +=
-            (qty.divide(maxCapacityPerCycle, RoundingMode.HALF_UP))
-                .multiply(durationPerCycle)
-                .longValue();
-      }
-
       duration += machine.getEndingDuration();
+      duration += nbCycles.add(new BigDecimal(-1)).multiply(new BigDecimal(machine.getSetupDuration())).longValue();
```
gdu-axelor: better to use subtract method

axelor-production/src/main/java/com/axelor/apps/production/service/OperationOrderWorkflowService.java
ego-axelor: It was in the original code. I guessed that if you put 0 as the maxCapacity it should mean that you can do one product at a time so the number of cycles is the qty (then I think that there is a mistake nbCycles should be equal to qty and not to BigDecimal(1)).

axelor-production/src/main/java/com/axelor/apps/production/service/OperationOrderWorkflowService.java
ego-axelor: done

axelor-production/src/main/java/com/axelor/apps/production/service/OperationOrderWorkflowService.java
gdu-axelor: No, I means that newBigDecimal(1) can be replaced per BigDecimal.ONE

ego-axelor: Ok got it so it is not necessary now

PR #2341, opened 2018-09-05
Feature RM#13583: Stock move: add column available qty on m2o product…

aga-axelor:
… grid displayed in popup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,8 @@
 - Tracking Number Form view : automatically fill fields from stockMoveLine while creating new record from stockMoveLine. 
 - Fix translation typo.
 - Add xml escape in configurator export.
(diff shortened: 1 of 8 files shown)


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/db/repo/ProductSupplychainRepository.java
```diff
@@ -0,0 +1,81 @@
...
+        if (stockLocationId != null) {
+          List<StockLocationLine> stockLocationLineList =
+              stockLocationLineRepo
+                  .all()
+                  .filter(
+                      "self.product.id = ?1 and self.stockLocation.id = ?2",
+                      productId,
+                      stockLocationId)
+                  .fetch();
+
```
gdu-axelor: We should get only one stockLocationLine for same product and stockLocation, so here a fetchOne method is enougth.

PR #2344, opened 2018-09-05
RM13655: Fix NPE when user use relational field in studio

dtr-axelor:


axelor-helpdesk/src/main/resources/domains/Ticket.xml
```diff
@@ -8,7 +8,7 @@
   <entity name="Ticket">
       
 		<string name="ticketSeq" title="Ticket Number" readonly="true" />
-		<string name="subject" title="Subject" required="true"/>
+		<string name="subject" title="Subject" required="true" namecolumn="true" />
 		<many-to-one name="project" ref="com.axelor.apps.project.db.Project" title="Project"/>
 		<many-to-one name="customer" ref="com.axelor.apps.base.db.Partner" title="Customer"/>
 		<many-to-one name="contactPartner" ref="com.axelor.apps.base.db.Partner" title="Customer contact"/>
```
sos-axelor: Is it the only model without name column ?. Have to find some solution on studio side rather than on models, because it will be possible that some models don't have any namecolumn as adk allows it.

dtr-axelor: As discussed, indeed there are other models like that.

Please don't merge this issue, we need to discuss it because it may be issue on ADK

dtr-axelor:
Issue fixed on ADK, no need to merge this PR.

PR #2355, opened 2018-09-07
Anomaly #13557 TRACKING NUMBER : form view

rch-axelor:


axelor-stock/src/main/resources/views/TrackingNumber.xml
```diff
@@ -17,21 +17,21 @@
     	onLoad="action-tracking-number-attrs-hide-fields" >
     	<panel name="main">
 	    	<field name="trackingNumberSeq"/>
-	        <field name="product" hidden="true" onChange="action-tracking-number-attrs-hide-fields" form-view="product-form" grid-view="product-grid"/>
+	        <field name="product" readonly="true" hidden="true" onChange="action-tracking-number-attrs-hide-fields" form-view="product-form" grid-view="product-grid"/>
```
sos-axelor: _Product will be always in readonly if id != null so you can use a client side expression (js)._
Read comments properly from previous commit.

PR #2357, opened 2018-09-07
Feature #13556 STOCK MOVE : Wizard for tracking number

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -844,4 +847,47 @@ public BigDecimal computeNetMass(StockMoveLine stockMoveLine, Company company)
...
+            LocalDate.parse(trackingNumberItem.get("warrantyExpirationDate").toString()));
+      }
+      if (trackingNumberItem.get("perishableExpirationDate") != null) {
+        trackingNumber.setPerishableExpirationDate(
+            LocalDate.parse(trackingNumberItem.get("perishableExpirationDate").toString()));
+      }
+      trackingNumber.setProduct(stockMoveLine.getProduct());
+      Beans.get(TrackingNumberRepository.class).save(trackingNumber);
+
+      StockMoveLine stMoveLine = new StockMoveLine();
```
gdu-axelor: I think it's better to use JPA.Copy method here instead of new, because, else, we will loss some informations.

PR #2366, opened 2018-09-11
RM#13661 Business project : enhancements of task and project view

aga-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/TeamTaskBusinessService.java
```diff
@@ -25,9 +25,15 @@
...
 import java.time.LocalDateTime;
+import java.util.Map;
 
 public interface TeamTaskBusinessService extends TeamTaskService {
+
   TeamTask create(SaleOrderLine saleOrderLine, Project project, User assignedTo);
 
   TeamTask create(TaskTemplate template, Project project, LocalDateTime date, BigDecimal qty);
+
+  Map<String, Object> getDiscount(TeamTask teamTask, BigDecimal unitPrice);
```
sos-axelor: Method should be 'updateDiscount' and it should only take 'teamTask' as parameter as unitPrice comes from TeamTask. and return type should be 'TeamTask' with updated discount values.

PR #2373, opened 2018-09-11
Anomaly #13594 VEHICLE : VehicleState

rch-axelor:


axelor-fleet/src/main/resources/views/Vehicle.xml
```diff
@@ -107,8 +107,7 @@
 			  <div class="span8">
 			    <span>{{plateNo}}<br/>
 			    {{company.name}}<br/>
-          {{driverPartner.name}}<br/>
-			    {{vehicleState.name}}</span>
+			    {{driverContact.fullName}}<br/></span>
```
sos-axelor: Change 'driverContact' to 'driverPartner'.

PR #2384, opened 2018-09-13
RM#13752 Project: Add translation

aga-axelor:


axelor-human-resource/src/main/resources/i18n/messages_fr.csv
```diff
@@ -688,7 +689,7 @@
 "Sunday","Dimanche",,
 "TSTimer","Minuteur feuille de temps",,
 "Task",,,
-"TeamTask",,,
+"TeamTask","Tâche",,
```
sos-axelor: Change title on domain from 'TeamTask' to 'Team task'

Issue #2331, opened 2018-08-30
Unit Conversion in BOM Cost Sheet Line/Cost Price calculation - incorrect unit conversion


dgagliar:
Original summary (with proposed fix) can be found on the forums [here](https://forum.axelor.com/t/unit-conversion/1376).

When calculating a cost price for a BOM, if the units of the BOM lines aren't the same as the unit on a BOM line's product, the unit conversion is applied incorrectly.

For example, if I have a component product CompA, with a unit of Tons and a cost price of $200, and add it as a BOM line with a unit of Pounds and a quantity of 500, the resulting calculated cost is $200,000,000. This is clearly incorrect.

What the `CostSheetLineServiceImpl` is doing in a few methods is this:

```
BigDecimal costPrice = unitConversionService.convert(product.getUnit(), unit,
    product.getCostPrice().multiply(consumptionQty));
```

Effectively, given the previous example, what happens is this:

```
 $200   500 Pound   2000 Pound   $200,000,000 Pound^2
----- X --------- X ---------- = --------------------
1 Ton       1         1 Ton             1 Ton^2
```

What we want to do is this:

```
BigDecimal costPrice = product.getCostPrice().multiply(unitConversionService.convert(unit,
    product.getUnit(), consumptionQty, appProductionService.getNbDecimalDigitForBomQty()));
```

This gives us the following formula:

```
 $200   500 Pound     1 Ton
----- X --------- X ---------- = $50
1 Ton       1       2000 Pound
```

Alternatively, the various calls to `unitConversionService.convert(...)` should reverse the order of their units, to simulate the division, but that doesn't read quite as clearly to me as explicitly converting the quantity units first, and then multiplying by the unit price.

wde-axelor:
Pull request #2385 has been opened to solve this issue. 
Thank you for having reported this behaviour.

PR #2385, opened 2018-09-13
Github issue #2331: fixed unit conversion issues when calculating cos…

wde-axelor:
…t price of bom


wde-axelor:
Fixing issue #2331.

Issue #2330, opened 2018-08-30
Unit Conversion in BOM Cost Sheet Line/Cost Price calculation


dgagliar:
Original summary (with proposed fix) can be found on the forums [here](https://forum.axelor.com/t/unit-conversion/1376).

`UnitConversionService` stops early when searching for an appropriate conversion. This means if you are trying to convert `A -> B`, and have defined conversions from `A -> B` and `B -> A`, it is possible for the `UnitConversionService` to find `B -> A` first, invert it, and use the resulting coefficient, even if a more correct conversion exists.

In the case where the `UnitConversionService` decides to invert a conversion, it incorrectly calculates the scale of the resulting coefficient. When inverting a coefficient of `2000`, (i.e., when converting tons to pounds, or vice versa), the inverse coefficient should be `0.0005`, but is instead calculated as `0.000` because the scale is incorrectly determined to be `3`, when it should be `4`. This is because `UnitConversionService` calculates the necessary scale as `(int) Math.log10(unitConversion.getCoef().intValue())`, when it should be wrapping `Math.log10(...)` in a call to `Math.ceil(...)` before casting to an `int`. We want to round the results of `Math.log10(...)` up before truncating so as not to lose necessary information.

I ran into this issue while trying to debug why my unit conversions weren't working. A second ticket will be filed with that information.

wde-axelor:
A pull request has been opened to solve this issue (#2386).

After thinking about it, I don't there is an issue with the order in which the conversion rules are chosen.
It looks to me like the service searches for a perfect match in conversion rules first (i.e same start unit and end unit), and then only searches for rules with the units the other way around (i.e, start unit as end unit, and end unit as start unit.).

Even if I am mistaken, it shouldn't matter either way since the bug fix will ensure a direct conversion or an inverted conversion will both behave the same.

dgagliar:
It definitely can choose the inverse conversion rule first. The iteration logic is thus:

``` Java
for(UnitConversion uc : UnitConversion.all) {
   if (uc.matches?) {
      // Do forward match
      break;
   } else if (uc.inverse?) {
      // Do inverse match
      break;
   }
}
```

However, this behavior does not need to be fixed (it is not incorrect behavior, unit conversions should be a bijective function). Fixing the scale calculation is sufficient to correct the problem.

PR #2386, opened 2018-09-13
Github issue 2330: fiexed wrong scale on decimal result when invertin…

wde-axelor:
…g the coefficient of unit conversion


wde-axelor:
Fixing issue #2330.

PR #2392, opened 2018-09-17
Feature RM#13667 MANUF ORDER : sale order

rch-axelor:


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -233,6 +233,16 @@
     	<attribute for="isConsProOnOperation" name="readonly" expr="eval: !(prodProcess?.isConsProOnOperation)"/>
     </action-attrs>
     
+    <action-attrs name="action-manuf-order-atts-sale-order-onchange-client-partner">
```
sos-axelor: Domain action-attrs must be applied on onSelect not on onChange. To reset saleOrder you can create another action and that action should only apply if select partner is different then clientPartner from saleOrder.

axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -233,6 +233,16 @@
...
+	  <attribute if="clientPartner!=null" name="domain" expr="self.clientPartner = :clientPartner AND self.statusSelect = 3" for="saleOrder"/>
+	  <attribute if="clientPartner==null" name="domain" expr="self.statusSelect = 3" for="saleOrder"/>
+	  <attribute name="value" expr="" for="saleOrder"/>
```
sos-axelor: Value expression should be null not empty.

PR #2397, opened 2018-09-19
Feature #13667 MANUF ORDER : sale order

rch-axelor:


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -242,6 +242,19 @@
     	<attribute for="isConsProOnOperation" name="readonly" expr="eval: !(prodProcess?.isConsProOnOperation)"/>
     </action-attrs>
     
+    <action-attrs name="action-manuf-order-atts-sale-order-onselect">
+	  <attribute if="clientPartner!=null" name="domain" expr="self.clientPartner = :clientPartner AND self.statusSelect = 3" for="saleOrder"/>
+	  <attribute if="clientPartner==null" name="domain" expr="self.statusSelect = 3" for="saleOrder"/>
```
sos-axelor: There should be space after operator in any expression. So here 'clientPartner!=null' should 'clientPartner != null', follow this for all expression.

PR #2399, opened 2018-09-19
currency conversion: fix readonlyIf for “Get current rate”

beuss:
Current rate can be fetched as soon as we have a start & end currency,
no need for a line id.


axelor-base/src/main/resources/views/CurrencyConversionLine.xml
```diff
@@ -17,7 +17,7 @@
         <panel name="main">
 		  <field name="startCurrency" widget="SuggestBox" form-view="currency-form" grid-view="currency-grid"/>
 		  <field name="endCurrency" widget="SuggestBox" form-view="currency-form" grid-view="currency-grid"/>
-		  <button name="getRate" title="Get rate" readonlyIf="startCurrency == null || endCurrency == null || toDate != null || id == null"   onClick="action-group-base-currencyconversionline-getrate-click"/>
+		  <button name="getRate" title="Get rate" readonlyIf="startCurrency == null || endCurrency == null || toDate != null"   onClick="action-group-base-currencyconversionline-getrate-click"/>
 		</panel>
```
sos-axelor: Previous line rate is also checked for variation calculation. So changes are required on  [CurrencyConversionLineController.java](https://github.com/axelor/axelor-business-suite/blob/dev/axelor-base/src/main/java/com/axelor/apps/base/web/CurrencyConversionLineController.java#L68) too.

beuss: Not if the id is null: https://github.com/axelor/axelor-business-suite/blob/dev/axelor-base/src/main/java/com/axelor/apps/base/web/CurrencyConversionLineController.java#L60

beuss:
Rebased on head to get rid of conflicts, @gdu-axelor, can you merge it as it has already been reviewed?

PR #2431, opened 2018-09-28
RM#13962 Stock move : enable the changing tracking on move form

aga-axelor:


axelor-stock/src/main/resources/domains/StockMove.xml
```diff
@@ -99,6 +99,24 @@
...
+      <field name="estimatedDate" />
+      <field name="realDate" if="statusSelect == 3" />
+      <field name="originTypeSelect" if="originTypeSelect != null" />
+      <field name="exTaxTotal" />
+      <message if="true" on="CREATE">Move created</message>
+      <message if="statusSelect == 1" tag="important">Draft move</message>
+      <message if="statusSelect == 2" tag="info">Planned move</message>
+      <message if="statusSelect == 3" tag="success">Move realized</message>
+      <message if="statusSelect == 4" tag="success">Move canceled</message>
+    </track>
```
sos-axelor: Remove 'exTaxTotal', 'originTypeSelect', 'also' separate messages are there for 'statusSelect', so no need to put it as field. Add message for cancel status.

PR #2432, opened 2018-09-28
RM#13888 Invoice: Fix fiscal position account for partner account

aga-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderInvoiceServiceImpl.java
```diff
@@ -251,6 +252,11 @@ public Invoice createInvoiceAndLines(
     invoice.setOperationSubTypeSelect(operationSubTypeSelect);
 
     if (partnerAccount != null) {
+      if (invoice.getPartner() != null) {
+        partnerAccount =
+            Beans.get(FiscalPositionServiceAccountImpl.class)
+                .getAccount(invoice.getPartner().getFiscalPosition(), partnerAccount);
```
sos-axelor: Extend the FiscalPositionService interface in account module and add the method 'getAccount'  on that interface

PR #2434, opened 2018-09-28
Anomaly #13935 SALE ORDER : Partial invoicing on pack lines

rch-axelor:


sos-axelor:
No need of this , as subline qty should be never edited.

PR #2445, opened 2018-10-03
Anomaly RM#13788 ACCOUNT MOVE : report

rch-axelor:


axelor-account/src/main/resources/i18n/messages_fr.csv
```diff
@@ -1645,6 +1645,7 @@
 "moveLine.origin","Origine",,
 "moveLine.partner","Tiers",,
 "moveLine.reconcile.list","Liste des réconciliations",,
+"moveLine.reconcileGroup",,,
```
sos-axelor: Translation is already there on title of the field. You can put same

PR #2447, opened 2018-10-03
RM#13995 POL : add dummy boolean 'Filter on supplier' to filter the p…

aga-axelor:
…roduct on supplier catelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@
 - Add a button on studio process tracking to display related workflow dashboard.
 - Stock Location Line : Add Button to display detailed stock location lines with tracking number.
 - Stock move : enable the changing tracking on stock move form.
+- Purchase Order Line : add dummy boolean field 'Filter on supplier' to filter the product on supplier catelog.
 
 ## Improvements
 - Move project resource menu entry to project configuration menu.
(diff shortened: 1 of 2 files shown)


axelor-purchase/src/main/java/com/axelor/apps/purchase/web/PurchaseOrderLineController.java
```diff
@@ -339,4 +339,38 @@ public void supplierPartnerDomain(ActionRequest request, ActionResponse response
...
+                    .collect(Collectors.toList())
+                    .toString()
+                    .replace('[', '(')
+                    .replace(']', ')');
+      } else {
+        domain = "self.id = 0";
+      }
+    } else {
+      domain =
+          "self.isModel = false and self.expense = false and (self.endDate = null or self.endDate > :__date__) and self.purchasable = true";
```
gdu-axelor: the domain "self.isModel = false and self.expense = false and (self.endDate = null or self.endDate > :__date__) and self.purchasable = true" should be also applied when usFilterOnSupplier is true

axelor-purchase/src/main/java/com/axelor/apps/purchase/web/PurchaseOrderLineController.java
```diff
@@ -339,4 +339,36 @@ public void supplierPartnerDomain(ActionRequest request, ActionResponse response
...
+            " and self.id IN "
+                + purchaseOrder
+                    .getSupplierPartner()
+                    .getSupplierCatalogList()
+                    .stream()
+                    .map(it -> it.getProduct().getId())
+                    .collect(Collectors.toList())
+                    .toString()
+                    .replace('[', '(')
+                    .replace(']', ')');
```
sos-axelor: Should not get product and ids..etc. Instead use sub query like  " self.id in (select product.id from SupplierCatalog where supplierPartner.id = :supplierPartnerId)"

PR #2452, opened 2018-10-04
Anomaly RM#13992 MANUF ORDER : note

rch-axelor:


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -182,6 +182,9 @@
                 <field name="costPrice" readonly="true"/>
                 <field name="costSheetList" colSpan="12" readonly="true" form-view="cost-sheet-bill-of-material-form" grid-view="cost-sheet-bill-of-material-grid"/>
 			</panel>
+			<panel name="note" title="Note">
+				<field name="note" widget="html" colSpan="12" showTitle="false"/>
```
sos-axelor: html not asked.

PR #2459, opened 2018-10-05
RM#13986 Accounting Export : Fix move lines filter

aga-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -195,6 +195,8 @@
 		<view type="grid" name="move-line-grid" />
 		<view type="form" name="move-line-form" />
 		<view-param name="search-filters" value="move-line-filters"/>
+		<domain>self.move.accountingReport.id = :accountingReport_id</domain>
+		<context name="accountingReport_id" expr="eval: _accountingReportId"/>
```
sos-axelor: This will create issue when click from normal Move grid (from menu), as that time no 'accountingReportId' will be there.

PR #2460, opened 2018-10-08
Feature RM#14090 PRODUCT: Standard and Approval

aga-axelor:


axelor-base/src/main/resources/views/Product.xml
```diff
@@ -382,6 +386,11 @@
 				   name="value" for="inAti" expr="eval: __config__.app.getApp('base').getProductInAtiSelect() == 2" />
     </action-attrs>
     
+    <action-attrs name="action-product-attrs-control-on-receipt">
+    	<attribute name="value" for="controlOnReceipt" expr="eval: productApproval?.controlOnReceipt"/>
+    	<attribute name="readonly" for="controlOnReceipt" expr="eval: productApproval?.controlOnReceipt"/>
```
sos-axelor: This readonly will work only on onchange and not on onLoad. Use readonlyIf instead.

PR #2463, opened 2018-10-09
RM#13778 Business Project: Fix discount compute when invoicing timesh…

aga-axelor:
…eet lines

### DIFF
--- a/axelor-human-resource/src/main/java/com/axelor/apps/hr/service/timesheet/TimesheetServiceImpl.java
+++ b/axelor-human-resource/src/main/java/com/axelor/apps/hr/service/timesheet/TimesheetServiceImpl.java
@@ -561,7 +561,8 @@ public class TimesheetServiceImpl implements TimesheetService {
       PriceList priceList)
       throws AxelorException {
 
-    int discountTypeSelect = 1;
+    int discountMethodTypeSelect = PriceListLineRepository.TYPE_DISCOUNT;
+    int discountTypeSelect = PriceListLineRepository.AMOUNT_TYPE_FIXED;
     if (product == null) {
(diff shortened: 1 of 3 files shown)


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/timesheet/TimesheetServiceImpl.java
```diff
@@ -583,28 +584,23 @@ public Timesheet createTimesheet(User user, LocalDate fromDate, LocalDate toDate
       if (priceListLine != null) {
         discountTypeSelect = priceListLine.getTypeSelect();
       }
+
+      Map<String, Object> discounts =
+          priceListService.getDiscounts(priceList, priceListLine, price);
+      if (discounts != null) {
+        discountAmount = (BigDecimal) discounts.get("discountAmount");
+        priceDiscounted =
+            priceListService.computeDiscount(
```
sos-axelor: discountType should be also get from discounts.

axelor-human-resource/src/main/java/com/axelor/apps/hr/service/timesheet/TimesheetServiceImpl.java
```diff
@@ -561,14 +561,16 @@ public Timesheet createTimesheet(User user, LocalDate fromDate, LocalDate toDate
       PriceList priceList)
       throws AxelorException {
 
-    int discountTypeSelect = 1;
+    int discountMethodTypeSelect = 1;
+    int discountTypeSelect = 2;
```
sos-axelor: Use constants from  PriceListLineRepository instead of literals.

PR #2464, opened 2018-10-09
Anomaly RM#14068 Data backup: Export Apps

rch-axelor:


axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -175,31 +179,87 @@ public File create(Integer fetchLimit) {
...
+        }
+      } catch (Exception e) {
+      }
+      if (subClasses != null && subClasses.size() > 0) {
+        subClassMap.put(metaModel.getFullName(), subClasses);
+      }
+    }
+    return subClassMap;
+  }
+
```
sos-axelor: optimisation is required on this method, it can be done with single loop no need of inner loop.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -175,31 +179,87 @@ public File create(Integer fetchLimit) {
...
+      MetaModel metaModel, int start, Integer fetchLimit, List<String> subClasses)
+      throws ClassNotFoundException {
+    String query = "";
+    if (subClasses != null && subClasses.size() > 0) {
+      String whereStr = "";
+      for (String subClassName : subClasses) {
+        whereStr += "id NOT IN (select id from " + subClassName + ")";
+        if (whereStr.length() > 0 && !subClassName.equals(subClasses.get(subClasses.size() - 1))) {
+          whereStr += " AND ";
+        }
```
sos-axelor: Instead of checking index of last subClass , you can just add "AND" before condition everytime except first time where its empty.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -175,31 +179,87 @@ public File create(Integer fetchLimit) {
...
+      }
+      query = "SELECT count(*) FROM " + metaModel.getName() + " WHERE " + whereStr;
+    } else {
+      query = "SELECT count(*) FROM " + metaModel.getName();
+    }
+    return (long) JPA.em().createQuery(query).getSingleResult();
```
sos-axelor: Method for count and getting model should be only one. There is no need of count(*) funciton, but you can use count() method of query.  Use JpaRepository.of(modelClass) instead of JPA.em()

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -245,6 +305,9 @@ private CSVInput writeCSVData(
       if (AutoImportModelMap.containsKey(csvInput.getTypeName())) {
         csvInput.setSearch(AutoImportModelMap.get(csvInput.getTypeName()).toString());
       }
+      if (Class.forName(metaModel.getFullName()).getSuperclass() == App.class) {
+        csvInput.setSearch("self.code = :code");
```
sos-axelor: it should be else If  here .

PR #2467, opened 2018-10-09
Feature RM#14122 PARTNER : simpleFullname

rch-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/PartnerService.java
```diff
@@ -103,6 +105,7 @@ public Partner createPartner(
   public void setPartnerFullName(Partner partner) {
 
     partner.setFullName(this.computeFullName(partner));
+    partner.setSimpleFullName(this.computeSimpleFullName(partner));
   }
 
   public String computeFullName(Partner partner) {
```
sos-axelor: Code of this method should be factorised as mentioned on ticket. Use computeSimpleFullName and add sequence to it in this method.

axelor-base/src/main/resources/domains/Partner.xml
```diff
@@ -67,6 +67,7 @@
 	<string name="taxNbr" title="Tax N°"/>
 	
  	<string name="fullName" title="Name" namecolumn="true"/>
+ 	<string name="simpleFullName" title="Name" namecolumn="true"/>
```
gdu-axelor: namecolumn should be defined for fullName only.

PR #2484, opened 2018-10-11
Feature RM#14161 MRP : user and company

rch-axelor:


axelor-supplychain/src/main/resources/views/Mrp.xml
```diff
@@ -7,7 +7,9 @@
         <hilite if="statusSelect == 1" strong="true" color="warning"/>
         <hilite if="statusSelect == 2" strong="true" color="success"/>
         <field name="createdOn"/>
+        <field name="createdBy"/>
         <field name="stockLocation" form-view="stock-location-form" grid-view="stock-location-grid" colSpan="6"/>
+        <field name="stockLocation.company" title="company"/>
         <field name="endDate" colSpan="6"/>
```
sos-axelor: Why you have provided title for the company and also in small case ?

PR #2486, opened 2018-10-12
Anomaly RM#14098 Fix binary field restore & required field model import sequence

rch-axelor:
…port sequence


axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupServiceImpl.java
```diff
@@ -29,6 +29,8 @@
 
   static String configFileName = "config.xml";
 
+  static String byteArrFnameColName = "byteArrFileName";
```
sos-axelor: make all string constants as 'final'

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -307,6 +328,8 @@ private String getMetaModelHeader(Object value, Property property, CSVInput csvI
     switch (propertyTypeStr) {
       case "LONG":
         return propertyName.equalsIgnoreCase("id") ? "importId" : propertyName;
+      case "BINARY":
+        return DataBackupServiceImpl.byteArrFnameColName;
```
sos-axelor: It will not work when you have more than one byte[] field. Also better to have just prefix to identify it as a byte[] field. Like 'byte_', so if field name is 'image' then column name will be 'byte_image'.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -363,7 +386,17 @@ private String getMetaModelData(Property property, Object value) {
             ? ((ZonedDateTime) value).toLocalDateTime().toString()
             : value.toString();
       case "BINARY":
-        return Arrays.toString((byte[]) value);
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmSS");
+        String fileName =
+            LocalDateTime.now().format(formatter) + "_" + random.nextInt(10000) + ".png";
+        try {
```
sos-axelor: No needs to use random, better user ModelName_fieldName_id.png

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupCreateService.java
```diff
@@ -266,22 +271,38 @@ private CSVInput writeCSVData(
...
+                        property, metaModelMapper.get(dataObject, property.getName()), dirPath));
               }
             }
+            if (headerArr.contains("methodName")) {
+              dataArr.add(0, methodValue);
+            }
 
             if (headerFlag) {
+              if (headerArr.contains(DataBackupServiceImpl.byteArrFnameColName)) {
+                csvInput.setCallable(
```
sos-axelor: add one more flag for byteArray method, so no needs to check contains.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupRestoreService.java
```diff
@@ -131,4 +135,38 @@ public boolean SeuencesExist() {
...
+    int numRead = 0;
+    InputStream is = new FileInputStream(image);
+
+    try {
+      while (offset < bytes.length
+          && (numRead = is.read(bytes, offset, bytes.length - offset)) >= 0) {
+        offset += numRead;
+      }
+      java.lang.reflect.Method method = bean.getClass().getMethod(methodName, byte[].class);
+      method.invoke(bean, bytes);
```
sos-axelor: No needs to use java reflection, use Mapper instead to get values.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupRestoreService.java
```diff
@@ -131,4 +138,46 @@ public boolean SeuencesExist() {
...
+      final File image = path.resolve(fileName).toFile();
+      byte[] bytes = new byte[(int) image.length()];
+      int offset = 0;
+      int numRead = 0;
+      InputStream is = new FileInputStream(image);
+
+      try {
+        while (offset < bytes.length
+            && (numRead = is.read(bytes, offset, bytes.length - offset)) >= 0) {
+          offset += numRead;
```
sos-axelor: What is the use of while loop ??

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupRestoreService.java
```diff
@@ -131,4 +138,46 @@ public boolean SeuencesExist() {
...
+        }
+        Mapper.of(bean.getClass()).set(bean, fieldName.substring(5), bytes);
+      } catch (Exception e) {
```
sos-axelor: Create mapper one time out of loop.

axelor-admin/src/main/java/com/axelor/apps/base/service/app/DataBackupRestoreService.java
```diff
@@ -131,4 +138,46 @@ public boolean SeuencesExist() {
     long total1 = (long) JPA.em().createQuery("SELECT count(*) FROM MrpLineType").getSingleResult();
     return total > 0 || total1 > 0 ? true : false;
   }
+
+  public Object importObjectWithByteArray(Object bean, Map<String, Object> values)
+      throws IOException {
+    assert bean instanceof Model;
+    final Path path = (Path) values.get("__path__");
+
+    for (String fieldName : getByteArrFields(values)) {
```
sos-axelor: No need of getByteArrFields method as its not used anywhere else.  Same loop can be used here to avoid double looping.

PR #2496, opened 2018-10-15
account: allow both substitution and reverseCharge on TaxEquiv

beuss:
Standard accounting rules for intra-EU VAT is as follow in France (for a standard rate VAT):

- French VAT account is replaced by intra-EU VAT account;
- Reverse charging is done on the opposite intra-EU VAT account.

For a purchase invoice it gives:

 - 4456xx is replaced by 445662
 - reverse charge is done on 4452xx

So we've to be able to both replace and reverse charge at the same time.


beuss:
Just rebased on dev, please @gdu-axelor or @axeloradmin review & merge or reject.

PR #2505, opened 2018-10-16
Feature RM#14250 STOCK MOVE : passed for payment

rch-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/db/repo/StockMoveSupplyChainRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.axelor.apps.account.db.Invoice;
+import com.axelor.apps.stock.db.StockMove;
+import com.axelor.apps.stock.db.repo.StockMoveManagementRepository;
+
+public class StockMoveSupplyChainRepository extends StockMoveManagementRepository {
+
+  @Override
+  public StockMove save(StockMove entity) {
+    StockMove stockMove = super.find(entity.getId());
```
sos-axelor: Using find method is totally wrong, no needs to use it.

PR #2507, opened 2018-10-16
account: add an option to allow editing of validated move

beuss:
Add a new option to allow editing validated account moves as long as
they haven't been exported. This allows to be more flexible when an
external accouting software is used.
This works around the fact that the accountingDayBook option is almost
never used (and the code flow in the invoicing system is such a maze
that this can not be fixed on a short term).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## Improvements
 - Show full name for products in Mrp report.
 - Sale and Purchase order form: Remove edit from stockLocation field.
+- Account: add option to allow modification of validated but not-yet-exported moves.
 - PERIOD : allow to reopen a period if the fiscal year is not closed
 
 ## Bug Fixes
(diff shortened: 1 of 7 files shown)


gdu-axelor:
Hi,

We can't allow to edit a validated move. We work on the way to get the accounting label NF203, so the only way to edit a move is to generate it in daybook mode.

The daybook mode feature is done really badly.

So, I've completely reviewed the accountingDayBook mode in order to have exactly the same behavior that for directly validated move, except that we can update it until there are validated.
https://github.com/axelor/axelor-business-suite/commit/55b1f20a0d256ab06ddcd4cf269091ddfb84a521

Also, the account move will be automatically validated when we close the period or year.

I have to manage a last point : update partner balances for a partner that we detach manually of the account move.

I will push the work soon.

Please ,tell me if it's enough and if this PR is still relevant.

Regards

beuss:
Hi,
The proposed commit is far more convenient that this one (as said in the original comment, this was more a workaround).
Can't wayt for your PR
Regards

gdu-axelor:
Hi,

FYI, I have pushed the end of the work in order to have a daybook mode fully usable.

Regards

PR #2514, opened 2018-10-17
RM#14228 Account : add viewer to display balance on account form

aga-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/AccountController.java
```diff
@@ -0,0 +1,58 @@
...
+      Account account = request.getContext().asType(Account.class);
+      account = accountRepository.find(account.getId());
+
+      BigDecimal balance =
+          accountService.computeBalance(account, AccountService.BALANCE_TYPE_DEBIT_BALANCE);
+
+      if (balance.compareTo(BigDecimal.ZERO) >= 0) {
+        response.setAttr("$balance", "title", I18n.get("Debit balance"));
+      } else {
+        response.setAttr("$balance", "title", I18n.get("Credit balance"));
```
sos-axelor: Put static text in ITranslation interface.

axelor-account/src/main/java/com/axelor/apps/account/web/AccountController.java
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+      response.setValue("$balance", balance);
```
sos-axelor: Balance must be always positive as mentioned on ticket. So if balance is negative multiply it by '-1'

PR #2517, opened 2018-10-18
Feature RM#13998 StockLocationLine : Add dashlet to know quantities o…

rch-axelor:
…f product


axelor-stock/src/main/resources/views/StockLocationLine.xml
```diff
@@ -24,6 +24,7 @@
 			<field name="lastInventoryRealQty" title="Real Qty"/>
 			<field name="lastInventoryDateT" title="Date"/>
 		</panel>
+		<panel-dashlet action="action-stock-location-line-view-product-qty" title="Product quantity in other stock locations"/>
 		<panel>
```
sos-axelor: Put dashlet at the end of form view.

PR #2524, opened 2018-10-19
RM 11913 : fixed wrong behaviours on discounts in sales/purchases/inv…

wde-axelor:
…oices


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/InvoiceLineSupplychainService.java
```diff
@@ -82,48 +87,91 @@ public Unit getUnit(Product product, boolean isPurchase) {
...
+
+    Map<String, Object> info = null;
+
+    List<SupplierCatalog> supplierCatalogList = invoiceLine.getProduct().getSupplierCatalogList();
+    if (supplierCatalogList != null && !supplierCatalogList.isEmpty()) {
+      SupplierCatalog supplierCatalog =
+          Beans.get(SupplierCatalogRepository.class)
+              .all()
+              .filter(
+                  "self.product = ?1 AND self.minQty <= ?2 AND self.supplierPartner = ?3 ORDER BY self.minQty DESC",
```
gdu-axelor: These process are already done in Purchase module, so better to use it in Supplychain too instead of duplicate the code.

PR #2537, opened 2018-10-23
Anomaly RM#14402 STOCKMOVELINE : sequence

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -311,6 +311,8 @@ public StockMoveLine createStockMoveLine(
       stockMoveLine.setProductTypeSelect(product.getProductTypeSelect());
     }
 
+    stockMoveLine.setSequence(stockMove.getStockMoveLineList().size() * 10);
+
```
sos-axelor: don't multiply by 10 because it will be changed when user move lines manually from o2m.

PR #2538, opened 2018-10-23
Feature RM#14367 Inventory : filter on Racks range

aga-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/InventoryService.java
```diff
@@ -517,6 +518,16 @@ public Boolean fillInventoryLineList(Inventory inventory) throws AxelorException
...
+      query += " and self.rack >= ?";
+      params.add(inventory.getFromRack());
+    }
+
+    if (Strings.isNullOrEmpty(inventory.getToRack())) {
+      query += " and self.rack >= ?";
+      params.add(inventory.getToRack());
+    }
+
     return Beans.get(StockLocationLineRepository.class)
```
sos-axelor: Both the conditions are wrong. You haven't tested ?

axelor-stock/src/main/resources/views/Inventory.xml
```diff
@@ -47,7 +47,11 @@
             <field name="productCategory" onSelect="action-inventory-on-select-product-category" onChange="action-inventory-on-change-product-category" grid-view="product-category-grid" form-view="product-category-form"/>
             <field name="excludeOutOfStock" colSpan="3"/>
             <field name="includeObsolete" colSpan="3"/>
-            <spacer/>
+           	<panel title="Racks range" colSpan="12">
+		      <field name="fromRack" colSpan="3"/>
+		      <field name="toRack" colSpan="3"/>
+		    </panel>
+		    <spacer/>
             <button name="fillInventory" title="Fill inventory" showIf="stockLocation != null" onClick="action-group-stock-inventory-fillinventory-click" colSpan="3"/>
```
sos-axelor: replace original <spacer> with this new subpanel and no needs to provide any colspan.

sos-axelor: My mistake, I typed 'original' instead of 'spacer' , keep main panel but replace spacer with sub panel for 'Rack range'

PR #2542, opened 2018-10-24
RM 14423 : Fixed bug where a fixed discount could sometimes be applie…

wde-axelor:
…d several times to the same unit price


wde-axelor:
No changelog as it is part of the discounts rework of PR#2524

PR #2544, opened 2018-10-24
Feature RM#14394 STOCKLOCATION : aggregate

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/web/StockLocationController.java
```diff
@@ -70,15 +76,24 @@ public void print(ActionRequest request, ActionResponse response) throws AxelorE
     List<Integer> lstSelectedLocations = (List<Integer>) request.getContext().get("_ids");
     if (lstSelectedLocations != null) {
       for (Integer it : lstSelectedLocations) {
-        locationIds += it.toString() + ",";
+        Set<Long> idSet =
+            stockLocationService.getContentStockLocationIds(stockLocationRepo.find(new Long(it)));
+        if (!idSet.isEmpty()) {
+          locationIds += org.apache.commons.lang3.StringUtils.join(idSet, ',') + ",";
```
sos-axelor: Use 'Joiner' instead of org.apache.commons.lang3.StringUtils.join, also just use outside the loop, so better create one 
List of ids and than use joiner. 
Also improve method to find subLocations , for reference 
https://github.com/axelor/axelor-business-suite/blob/master/axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/MrpServiceImpl.java#L869 (You can move this method to stockLocationService and use it for both but with parameter to consider virtual location or not).

PR #2556, opened 2018-10-25
RM#14443: PRODUCT : translation of shippingCoef fields

rme-axelor:


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Enable editable grid on AccountEquiv and added sort on AccountEquiv and TaxEquiv
 - Added possibility to hide lines with currentQuantity and futureQuantity equal to 0 in stock locations.
 - BankOrder : Specify limit for BankOrderLines.
+- Base: Changes in translation field.
```
sos-axelor: Its should not be  base but 'Product' and not changes in translation field but 'Fix wrong translation of fields'.

PR #2562, opened 2018-10-25
Anomaly #14461 PARTNER : balance viewer

sve-axelor:


axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -741,7 +741,8 @@
 	</action-attrs>
 
 	<action-record name="action-partner-record-set-balance" model="com.axelor.apps.base.db.Partner">
-		<field name="$balance" expr="eval: __repo__(AccountingSituation).all().filter('self.partner.id = ?', id).fetch().sum{ it.balanceDueCustAccount }"/>
+		<field name="$balance" expr="eval: 0" if="__repo__(AccountingSituation).all().filter('self.partner.id = ?', id).fetch().sum{ it.balanceDueCustAccount } == null"/>
+		<field name="$balance" expr="eval:__repo__(AccountingSituation).all().filter('self.partner.id = ?', id).fetch().sum{ it.balanceDueCustAccount }.abs()" if="__repo__(AccountingSituation).all().filter('self.partner.id = ?', id).fetch().sum{ it.balanceDueCustAccount } != null"/>
```
sos-axelor: its not good to calculate $balance everytime for if and expression, better create another dummy field  $positiveBalance (title will remain same as $balance) display it instead of $balance and its value will be similar to $balance but just multiply $balance by -1 if its negative.

PR #2575, opened 2018-10-26
RM#14506:AccountingReport Fix NPE

cma-axelor:


sos-axelor:
Its a adk issue, so changes are required on abs side.

PR #2579, opened 2018-10-29
RM#14512 Account Move : Removed readonlyIf from MoveLine-form and made editable even after validated status

jpo-axelor:


beuss:
I have to admit it's starting to become tiring to (try to) contribute to ABS… the PR #2507 has been there for 13 days, why starting another one with less configuration capabilities instead of commenting on the existing one?
@gdu-axelor can you have a word on this? If foreign PR aren't going to be merged, just say it explicitely.
Regards

gdu-axelor:
Initial request (RM#14512) of this PR is to put the analytic move line in readonly, absolutely not all move fields. It's a misunderstanding with the developper.

For the PR #2507, I have direclty add comment on it

PR #2586, opened 2018-10-29
RM#14512 Account Move : Removed readonlyIf from MoveLine-form and made editable even after validated status

jpo-axelor:


axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -24,7 +24,7 @@
         <field name="description"/>
     </grid>
     
-    <grid name="move-line-move-grid" model="com.axelor.apps.account.db.MoveLine" title="Move lines" editable="true" onNew="action-account-move-line-onnew-group">
+    <grid name="move-line-move-grid" model="com.axelor.apps.account.db.MoveLine" title="Move lines"  onNew="action-account-move-line-onnew-group">
```
sos-axelor: why you have removed editable="true" ?

PR #2587, opened 2018-10-29
RM#14536 Sequence : Fix fill company and put it readonly

aga-axelor:


axelor-base/src/main/resources/views/Sequence.xml
```diff
@@ -66,8 +66,9 @@ xsi:schemaLocation="http://axelor.com/xml/ns/object-views http://axelor.com/xml/
     <action-record name="action-sequence-default" model="com.axelor.apps.base.db.Sequence">
     	<field name="padding" expr="3"/>
     	<field name="toBeAdded" expr="1"/>
-    	<field name="company"  expr="eval:__user__.activeCompany" if="__user__.activeCompany != null"/>
-    	<field name="company"  expr="eval:__repo__(Company).all().fetchOne()" if="__user__.activeCompany == null &amp;&amp; __repo__(Company).all().fetch().size == 1"/>
+    	<field name="company" expr="eval: _parent?.company"/>
+    	<field name="company"  expr="eval:__user__.activeCompany" if="(_parent == null || _parent.company == null) &amp;&amp; __user__.activeCompany != null"/>
+    	<field name="company"  expr="eval:__repo__(Company).all().fetchOne()" if="(_parent == null || _parent.company == null) &amp;&amp; __user__.activeCompany == null &amp;&amp; __repo__(Company).all().fetch().size == 1"/>
     </action-record>
```
sos-axelor: Use count instead of fetch for company query.

sos-axelor: Not > 0 but equals 1 , means if there is only 1 company select it directly.

PR #2588, opened 2018-10-30
RM#14540 Fix issue when we click on Cancel Receipt button and changed…

vbh-axelor:
… fr translation for 'Cancel receipt'


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/PurchaseOrderServiceSupplychainImpl.java
```diff
@@ -348,7 +348,8 @@ public void cancelReceipt(PurchaseOrder purchaseOrder) throws AxelorException {
     List<StockMove> stockMoveList =
         Beans.get(StockMoveRepository.class)
             .all()
-            .filter("self.purchaseOrder = ?1 AND self.statusSelect = 2", purchaseOrder)
+            .filter("self.originTypeSelect LIKE ? AND self.originId = ? AND self.statusSelect = 2",
```
sos-axelor: It must be '=' not LIKE

PR #2595, opened 2018-10-30
RM#14512 ACCOUNT MOVE : Removed readonlyIf from MoveLine-form and made editable form even after validated status

jpo-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -71,7 +71,7 @@
             </panel>
 		</panel>
 		
-		<panel-related field="moveLineList" showIf="journal &amp;&amp; company &amp;&amp; date &amp;&amp; period" readonlyIf="statusSelect == 3" grid-view="move-line-move-grid" form-view="move-line-move-form"/>
+		<panel-related field="moveLineList" showIf="journal &amp;&amp; company &amp;&amp; date &amp;&amp; period" canNew="statusSelect &lt; 3" canRemove="statusSelect &lt; 3" grid-view="move-line-move-grid" form-view="move-line-move-form"/>
```
sos-axelor: add it on o2m and make it editable="false" when status = validated.

PR #2602, opened 2018-10-31
[Feature #14597] LOGISTICAL FORM : alert

psh-axelor:


axelor-stock/src/main/resources/views/LogisticalForm.xml
```diff
@@ -21,7 +21,7 @@
     </grid>
 
     <form model="com.axelor.apps.stock.db.LogisticalForm" title="Logistical form" name="logistical-form-form"
-        width="large" onNew="action-logistical-form-new" onSave="action-logistical-form-save">
+        width="large" onNew="action-logistical-form-new" onSave="action-logistical-form-save,action-logistical-form-validate-mass-on-save">
         <toolbar>
```
sos-axelor: action-logistical-form-save is already a action-group, add this new action in that action-group

PR #2603, opened 2018-10-31
RM#14595:PRODUCT:Add field Country Of Origin

cma-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -297,6 +297,10 @@ public StockMoveLine createStockMoveLine(
     stockMoveLine.setUnit(unit);
     stockMoveLine.setTrackingNumber(trackingNumber);
 
+    if (product.getCountryOfOrigin() != null) {
```
sos-axelor: No need of null check

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -863,6 +867,10 @@ public void setProductInfo(StockMove stockMove, StockMoveLine stockMoveLine, Com
       stockMoveLine.setProductModel(product.getParentProduct());
     }
 
+    if (product.getCountryOfOrigin() != null) {
+      stockMoveLine.setCountryOfOrigin(product.getCountryOfOrigin());
```
sos-axelor: No need of null check

PR #2604, opened 2018-10-31
Anomaly RM#14387 PRODUCT : Stock managed

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationLineServiceImpl.java
```diff
@@ -98,37 +98,40 @@ public void updateLocation(
...
-        product.getCode(),
-        qty,
-        current,
-        future,
-        isIncrement,
-        lastFutureStockMoveDate);
+    if (!product.getStockManaged()) {
+      StockLocationLine stockLocationLine =
+          this.getOrCreateStockLocationLine(stockLocation, product);
+
```
sos-axelor: move product.getStockManaged() int to getOrCreateStockLocationLine method, and put null check everywhere where that method is called.

sos-axelor:
Managed it by other commit

PR #2605, opened 2018-10-31
[Feature #14591] STOCK MOVE : specificPackage

psh-axelor:


axelor-stock/src/main/resources/domains/StockMove.xml
```diff
@@ -76,6 +76,8 @@
     <string name="origin" readonly="true" title="Origin"/>
 
     <enum name="modeOfTransport" ref="com.axelor.apps.stock.db.ModeOfTransport"/>
+    
+    <string name="saleOrderSpecificPackage" large="true"/>
```
sos-axelor: add field on supplychain module only and display only if supplychain app is  installed

axelor-supplychain/src/main/resources/domains/StockMove.xml
```diff
@@ -9,6 +9,7 @@
   
     <!-- Onglet Configuration -->
     <many-to-one name="invoice" ref="com.axelor.apps.account.db.Invoice" title="Invoice"/>
+    <string name="saleOrderSpecificPackage" large="true"/>
```
sos-axelor: add title same as on SaleOrder domain

PR #2606, opened 2018-10-31
RM#14592 Product : Add a decimal field 'economicManufOrderQty' and product.economicManufOrderQty in ManufOrder form view and its translation.

jpo-axelor:


axelor-base/src/main/resources/views/Product.xml
```diff
@@ -191,6 +191,7 @@
                     if-module="axelor-production"
                     if="__config__.app.isApp('production')">
                     <field name="costSheetGroup" showIf="productSubTypeSelect == 3"/>
+                    <field name="product.economicManufOrderQty"/>
```
sos-axelor: what is this ? , why you have added dotted field in product?

axelor-base/src/main/resources/i18n/messages_en.csv
```diff
@@ -1,1430 +1,1430 @@
-"key","message","comment","context"
```
sos-axelor: Correct formatting of translation files , see here " is removed.

PR #2607, opened 2018-10-31
Anomaly #14584 PARTNER : grid and card view

sve-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/PartnerService.java
```diff
@@ -498,4 +499,20 @@ public String getPhoneNumberFieldName(String actionName) {
...
+
+  public String computeCompanyStr(Partner partner) {
+    List<String> companyCodeList = new ArrayList<String>();
+    if (partner.getCompanySet() != null) {
+      for (Company company : partner.getCompanySet()) {
+        companyCodeList.add(company.getCode());
+      }
+    }
+    String companyStr =
+        companyCodeList != null ? companyStr = String.join(",", companyCodeList) : null;
```
sos-axelor: Don't create list of code and than join list, instead directly join in loop.

axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -526,13 +533,16 @@
...
 			  	<address>
 			  	{{address.fullName}}
 			  	</address>
 			    <span>
 			    {{fixedPhone}}<br/>
 			    {{emailAddress.address}}<br/>
-			    {{partnerCategory.name}}</span>
+			    {{partnerCategory.name}}<br/>
+			    {{companyStr}}<br/>
+			    Fiscal Position : {{fiscalPosition.code}}</span>
```
sos-axelor: Use x-translate for translation, check sale order cards view for reference.

PR #2609, opened 2018-11-01
RM#14626 Sale Order Line : Fix calculate 'Available stock' on onload

aga-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/web/SaleOrderLineController.java
```diff
@@ -114,10 +114,15 @@ public void checkStocks(ActionRequest request, ActionResponse response) {
...
-    if (saleOrderLine.getSaleOrder() == null) {
-      return;
+    Context context = request.getContext();
+    SaleOrderLine saleOrderLine = context.asType(SaleOrderLine.class);
+    SaleOrder saleOrder = saleOrderLine.getSaleOrder();
+
+    if (saleOrder == null && context.getParent() != null) {
+      saleOrder = context.getParent().asType(SaleOrder.class);
+      saleOrderLine.setSaleOrder(saleOrder);
     }
```
sos-axelor: Use method saleOrderLineService.getSaleOrder(context) to get saleOrder, don't set it on line but send it as another parameter to method getAvailableStock, also fix getAvailableStock method in service, it should use stocklocation service to get  stockLocationLine

PR #2613, opened 2018-11-01
RM#14593 ManufOrder : Add new many-to-one field 'unit' and set its value from billOfMaterial unit and also added to report.

jpo-axelor:


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -252,6 +253,10 @@
 	  <attribute if="clientPartner == null" name="domain" expr="self.statusSelect = 3" for="saleOrder"/>
 	</action-attrs>
 	
+	<action-attrs name="action-manuf-order-attrs-set-unit">
+		<attribute name="value" for="unit" expr="eval: billOfMaterial.unit"/>
```
sos-axelor: It will generate error when billOfMaterial is null

axelor-production/src/main/resources/reports/ManufOrder.rptdesign
```diff
@@ -148,6 +148,8 @@
             </list-property>
             <property name="odaDriverClass">org.postgresql.Driver</property>
             <property name="odaURL">jdbc:postgresql://localhost:5432/</property>
+            <property name="odaUser">axelor</property>
+            <encrypted-property name="odaPassword" encryptionID="base64">YXhlbG9y</encrypted-property>
```
sos-axelor: always remove user/password from report design

axelor-production/src/main/java/com/axelor/apps/production/service/ManufOrderWorkflowService.java
```diff
@@ -94,6 +95,9 @@ public ManufOrder plan(ManufOrder manufOrder) throws AxelorException {
     }
 
     manufOrder.setPlannedEndDateT(this.computePlannedEndDateT(manufOrder));
+    if(!ObjectUtils.isEmpty(manufOrder.getBillOfMaterial()) && !ObjectUtils.isEmpty(manufOrder.getBillOfMaterial().getUnit())) {
```
sos-axelor: no needs to check isEmpty , just check != null and only one time (means don't check for unit).

PR #2632, opened 2018-11-02
RM#14579:Move Origin Date

cma-axelor:


axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -351,6 +354,10 @@
 		<field name="credit" expr="0" if="credit &lt; 0"/>
 	</action-record>
 	
+	<action-record name="action-record-account-moveline-date-onchange" model="com.axelor.apps.account.db.MoveLine">
```
sos-axelor: Change 'action-record-account-moveline-' to 'action-move-line-record-'

PR #2642, opened 2018-11-03
Anomaly RM#14668 Databackup : Manage metafile documents

rch-axelor:


sos-axelor:
Attachment backup is not required.

PR #2647, opened 2018-11-05
RM#14669 - Sale Order: Add translation for prompt message

jpo-axelor:


axelor-sale/src/main/resources/i18n/messages_en.csv
```diff
@@ -457,7 +457,7 @@
 "Value",,,
 "Version Number",,,
 "WT",,,
-"Warning, a completed order can't be changed anymore, do you want to continue?",,,
+"Warning, a completed order can't be changed anymore, do you want to continue?","Warning, a completed order can't be changed anymore, do you want to continue?",,
```
sos-axelor: no needs to add translation for english as string is already in english

PR #2660, opened 2018-11-06
RM#14722 Review SaleOrderLine: get saleOrder.

rme-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
```diff
@@ -386,24 +386,19 @@ public SaleOrder getSaleOrder(Context context) {
 
     Context parentContext = context.getParent();
 
-    if (!parentContext.getContextClass().toString().equals(SaleOrder.class.toString())) {
-
-      parentContext = parentContext.getParent();
-    }
-
     if (parentContext == null) {
       return null;
```
sos-axelor: It will not work when saleOrderLine open directly in form view. Because at that time parentContext will be null.

axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
```diff
@@ -386,24 +386,19 @@ public SaleOrder getSaleOrder(Context context) {
...
 
-    SaleOrder saleOrder = parentContext.asType(SaleOrder.class);
+    SaleOrder saleOrder = null;
 
-    if (!parentContext.getContextClass().toString().equals(SaleOrder.class.toString())) {
-
-      SaleOrderLine saleOrderLine = context.asType(SaleOrderLine.class);
-
-      saleOrder = saleOrderLine.getSaleOrder();
+    if (parentContext.getContextClass().toString().equals(SaleOrder.class.toString())) {
```
sos-axelor: no needs to use  toString() to compare class.

PR #2661, opened 2018-11-06
RM#14710:Add columns in inventoryLineList and change default real quantity

cma-axelor:


axelor-stock/src/main/resources/views/InventoryLine.xml
```diff
@@ -34,4 +36,7 @@
 		<call class="com.axelor.apps.stock.web.InventoryLineController" method="updateInventoryLine"/>
 	</action-method>
     
+    <action-method name="action-inventory-line-method-compute-gap">
```
sos-axelor: Change the action name from 'action-inventory-line-method-compute-gap' to 'action-inventory-line-method-compute', also change the method name on controller. It seems you haven't tested this action, because method name in action-method and method on controller is different.

PR #2676, opened 2018-11-15
RM#14673 PRODUCT: advanced search

vbh-axelor:


axelor-base/src/main/resources/i18n/messages_en.csv
```diff
@@ -805,7 +809,7 @@
 "No duplicate finder field configured.",,,
 "No duplicate records found",,,
 "No help",,,
-"No period found or it has been closed for the company %s",,,
+"No period found or it has been closed for the company %s for the date %s","Aucune période trouvée ou celle-ci clôturée pour la société: %s et la date: %s",,
```
sos-axelor: why you have added this translation ?

PR #2682, opened 2018-11-19
RM#14936 : Generate new barcodes on copying MO

rme-axelor:


axelor-production/src/main/java/com/axelor/apps/production/db/repo/ManufOrderManagementRepository.java
```diff
@@ -52,8 +56,19 @@ public ManufOrder copy(ManufOrder entity, boolean deep) {
...
       }
     }
     return super.copy(entity, deep);
   }
+
+  @Override
+  public ManufOrder save(ManufOrder entity) {
+    for (OperationOrder operationOrder : entity.getOperationOrderList()) {
+      if (operationOrder.getBarCode() == null) {
+        operationOrderRepo.save(operationOrder);
```
sos-axelor: Create separate method in operationOrderService to generate barcode. Use that method in operation order save and from this save.

PR #2708, opened 2018-11-23
Feature RM#15025 Supplier Invoice : Alert if already created

rch-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -871,6 +871,14 @@ InvoiceSet.length > 0))"
 		<error message="Error - You are not authorized to perform this action" if="__user__.group.isClient"/>
     </action-validate>
 
+    <!-- ACTION CONDITION -->
+    
+    <action-condition name="action-account-invoice-condition-supplier-invoice-number">
```
sos-axelor: put this condition on 'onSave' too, because on duplicate it will not check this condition.

PR #2714, opened 2018-11-26
Anomaly RM#15070 ACCOUNTING EXPORT : NPE

rch-axelor:


sos-axelor:
No need as done on dev

PR #2715, opened 2018-11-26
Feature RM#15054 AppStock : Add new field 'economicArea'

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/DeclarationOfExchangesServiceImpl.java
```diff
@@ -65,9 +63,24 @@ public DeclarationOfExchangesServiceImpl(StockConfigService stockConfigService)
...
+      exportServiceClassMap =
+          ImmutableMap.of(
+              economicArea.getCode(),
+              ImmutableMap.of(
+                  ProductRepository.PRODUCT_TYPE_STORABLE,
+                  DeclarationOfExchangesExporterGoods.class,
+                  ProductRepository.PRODUCT_TYPE_SERVICE,
+                  DeclarationOfExchangesExporterServices.class));
 
+      if (exportServiceClassMap != null) {
```
sos-axelor: You are assigning a value to map in previous statement , how it can be null ?

PR #2737, opened 2018-11-28
[Feature #14633] MOVE : automatic tax line generation

psh-axelor:


psh-axelor:
Move to dev-14633 branch.

PR #2743, opened 2018-11-29
Feature RM#15133 StockMove : filter on available products

rch-axelor:


axelor-stock/src/main/resources/views/StockMove.xml
```diff
@@ -210,7 +210,9 @@
                     <viewer>{{record.toAddressStr}}</viewer>
                 </field>
             </panel>
-
+            
+            <field name="$filterOnAvailableProducts" title="Filter on Available Products" type="boolean" onChange="save,action-stock-move-method-update-stock-move-line"/>
```
sos-axelor: Add a real field on stockMove instead of dummy.  Don't save on onChange, but just update field on all lines.

axelor-stock/src/main/java/com/axelor/apps/stock/web/StockMoveController.java
```diff
@@ -459,4 +459,17 @@ public void setAvailableStatus(ActionRequest request, ActionResponse response) {
...
     response.setValue("stockMoveLineList", stockMove.getStockMoveLineList());
   }
+
+  public void updateMoveLineFilterOnAvailableproduct(
+      ActionRequest request, ActionResponse response) {
+    Context context = request.getContext();
+    if (context.containsKey("filterOnAvailableProducts")) {
+      boolean filterOnAvailableProducts = (boolean) context.get("filterOnAvailableProducts");
+      StockMove stockMove = context.asType(StockMove.class);
+      for (StockMoveLine stockMoveLine : stockMove.getStockMoveLineList()) {
```
sos-axelor: Put null check on getStockMoveLineList(), it will give error when method called with new StockMove and without lines.

axelor-stock/src/main/resources/views/StockMove.xml
```diff
@@ -210,7 +210,10 @@
                     <viewer>{{record.toAddressStr}}</viewer>
                 </field>
             </panel>
-
+            
+            <field name="$filterOnAvailableProducts" type="boolean" title="Filter on available products" onChange="action-stock-move-method-update-stock-move-line"/>
```
sos-axelor: Put this field also as transient instead of dummy and with default true same as on line.

PR #2749, opened 2018-11-30
Anomaly RM#15172 Factor : new organization for debt recovery submenu

rch-axelor:


axelor-account/src/main/resources/views/Menu.xml
```diff
@@ -178,8 +178,10 @@
 		<view type="form" name="debt-recovery-history-form" />
 	</action-view>
 
+    <menuitem title="Factoring" name="debt-recovery-root-factoring" parent="account-root-debt-recovery"/>
```
sos-axelor: add menu order

PR #2752, opened 2018-12-03
RM#15185: APPBASE : unable to save the record.

rme-axelor:


axelor-base/src/main/resources/views/AppBase.xml
```diff
@@ -34,7 +34,7 @@
 		    	<panel name="productBarCodePanel" title="Product BarCode">
 		    		<field name="activateBarCodeGeneration" widget="Boolean-switch"/>
 		    		<field name="editProductBarcodeType" showIf="activateBarCodeGeneration==true" widget="boolean-switch"/>
-     			 	<field name="barcodeTypeConfig" showIf="activateBarCodeGeneration==true" requiredIf="editProductBarcodeType==false"/>
+     			 	<field name="barcodeTypeConfig" showIf="activateBarCodeGeneration==true" requiredIf="activateBarCodeGeneration==true &amp;&amp; editProductBarcodeType==false"/>
```
sos-axelor: remove == true or == false statements, if its boolean field we don't needs compare == true or false but just field

PR #2770, opened 2018-12-11
RM#15295 Partner : address on grid view

jpo-axelor:


axelor-base/src/main/java/com/axelor/apps/base/db/repo/PartnerBaseRepository.java
```diff
@@ -69,6 +69,14 @@ public Partner save(Partner partner) {
               I18n.get(IExceptionMessage.PARTNER_EMAIL_EXIST));
         }
       }
+
+      for (PartnerAddress partnerAddress : partner.getPartnerAddressList()) {
```
sos-axelor: If not contactPartner first set null to mainContact and then set mainAddress in loop

PR #2775, opened 2018-12-11
RM#15295 Partner : address on grid view

jpo-axelor:


axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -18,7 +18,7 @@
 		<field name="partnerCategory" grid-view="partner-category-grid" form-view="partner-category-form"/>
 		<field name="fiscalPosition.code" title="Fiscal position"/>
 		<field name="registrationCode" />
-		<field name="address.fullName" title="Address"/>
+		<field name="mainAddress.fullName" title="Address"/>
```
sos-axelor: Just "mainAddress" and not "mainAddress.fullName". The 'fullName' is a nameColumn of Address object so it will display fullName. Also no need of title here as its already given on domain.

PR #2788, opened 2018-12-13
RM#15320 SUBROGATION RELEASE : add change tracking and new fields

aga-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/SubrogationReleaseServiceImpl.java
```diff
@@ -111,6 +111,7 @@ public void transmitRelease(SubrogationRelease subrogationRelease) throws Axelor
 
     subrogationRelease.setSequenceNumber(sequenceNumber);
     subrogationRelease.setStatusSelect(SubrogationReleaseRepository.STATUS_TRANSMITTED);
+    subrogationRelease.setTransmissionDate(LocalDate.now());
```
sos-axelor: Use appBaseService to get datetime.

PR #2792, opened 2018-12-14
RM#15320 SUBROGATION RELEASE : add change tracking and new fields

aga-axelor:


axelor-account/src/main/resources/domains/SubrogationRelease.xml
```diff
@@ -26,6 +32,15 @@
 			public static final int STATUS_CLEARED = 4;
 			public static final int STATUS_CANCELED = 5;
 		]]></extra-code>
+		
+		<track>
+			<field name="statusSelect"/>
+			<message if="true" on="CREATE">Release created</message>
```
sos-axelor: Keep message same as status due to translation issue, as there is no translation for 'Release created'

PR #2797, opened 2018-12-14
RM#15250 StockMove : Added two boolean 'isWithBackorder' & 'isWithReturnSurplus' in StockConfig and copy their values in StockMove

jpo-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveServiceImpl.java
```diff
@@ -157,7 +157,8 @@ public StockMove createStockMove(
     stockMove.setForwarderPartner(forwarderPartner);
     stockMove.setIncoterm(incoterm);
     stockMove.setIsIspmRequired(stockMoveToolService.getDefaultISPM(clientPartner, toAddress));
-
+    stockMove.setIsWithBackorder(company.getStockConfig().getIsWithBackorder());
+    stockMove.setIsWithReturnSurplus(company.getStockConfig().getIsWithReturnSurplus());
     return stockMove;
```
sos-axelor: Not set in this method but method called by this method to create stockMove.

PR #2804, opened 2018-12-14
RM#13329 : Import FEC file exported from accounting export

aga-axelor:


axelor-account/src/main/java/com/axelor/csv/script/ImportMove.java
```diff
@@ -0,0 +1,123 @@
...
+                  .filter("self.partnerSeq = ?", values.get("partnerSeq").toString())
+                  .fetchOne();
+          move.setPartner(partner);
+        }
+        moveRepository.save(move);
+      }
+      moveLine.setMove(move);
+
+    } catch (Exception e) {
+      e.printStackTrace();
```
sos-axelor: Use traceback service

axelor-account/src/main/java/com/axelor/csv/script/ImportMove.java
```diff
@@ -0,0 +1,123 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Map;
+import javax.transaction.Transactional;
+
+public class ImportMove {
+
+  @Inject private MoveRepository moveRepository;
+
+  @Transactional
+  public Object importMove(Object bean, Map<String, Object> values) throws AxelorException {
```
sos-axelor: Method name should be 'importFECMove'

PR #2841, opened 2018-12-20
Anomaly RM#15483 SALE ORDER : AnalyticDistributionTemplate

rch-axelor:


axelor-sale/src/main/resources/views/SaleOrder.xml
```diff
@@ -274,7 +274,7 @@
 	  </panel>
 	  <panel-tabs name="mainPanelTab" hideIf="clientPartner == null">
 	    <panel name="contentPanel" title="Content">
-	      <panel-related name="detailLinesPanel" title="Details Lines" readonlyIf="clientPartner == null || statusSelect == 2 || (statusSelect == 3 &amp;&amp; !orderBeingEdited) || statusSelect == 4" colSpan="12" field="saleOrderLineList" form-view="sale-order-line-form" grid-view="sale-order-line-grid" orderBy="sequence" onChange="action-sale-order-line-change-group" canMove="true"/>
+	      <panel-related name="detailLinesPanel" title="Details Lines" colSpan="12" field="saleOrderLineList" form-view="sale-order-line-form" grid-view="sale-order-line-grid" orderBy="sequence" onChange="action-sale-order-line-change-group" canMove="true"/>
```
sos-axelor: It will allows to remove lines. the attribute 'canRemove' should be disable on all previous conditions, also it must be readonly  when clientPartner is null

PR #2846, opened 2018-12-21
Feature RM#15485 Project module : New object 'announcement'

rch-axelor:


axelor-project/src/main/resources/views/Announcement.xml
```diff
@@ -0,0 +1,31 @@
...
+    <field name="title"/>
+    <field name="project"/>
+    <field name="content"/>
+    <field name="date"/>
+  </grid>
+  
+  <form name="announcement-form" title="Announcement"
+    model="com.axelor.apps.project.db.Announcement" width="large"
+    onNew="action-project-record-announcement-onnew">
+    <panel>
```
sos-axelor: Add name for panel. Follow naming convention for panel , means all panel names should be ends with 'Panel'. 
This is single main panel, so its name should be 'mainPane'. Take care of this in future development.

PR #2859, opened 2018-12-27
RM#15562 : Business Support module : assignment field in TeamTask

rme-axelor:


axelor-business-support/src/main/resources/i18n/messages_fr.csv
```diff
@@ -0,0 +1,4 @@
+"key","message","comment","context"
+"Assignment","Affectation infrench",,
```
sos-axelor: "infrench" is not a translation

PR #2878, opened 2018-12-28
RM#15589 TeamTask : Added action onClick of 'acceptOrderBtn' to set v…

jpo-axelor:
…alue for assignment


axelor-project/src/main/resources/views/TeamTask.xml
```diff
@@ -344,8 +344,9 @@
 		<attribute name="value" for="taskEndDate" expr="eval:taskDate.plus((taskDuration/86400).longValue(), java.time.temporal.ChronoUnit.DAYS)" if="taskDate != null"/>
 	</action-attrs>
 	
-	<action-attrs name="action-task-attrs-set-isOrderAccepted-onclick">
+	<action-attrs name="action-task-attrs-set-values-onclick">
```
sos-axelor: why you have changed action name and made it more general ? Previous one was ok, instead it should be like
action-task-attrs-on-click-order-accepted

PR #2883, opened 2018-12-31
Feature RM#15566 Timesheet : Assign toInvoice while generating lines

rch-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/TimesheetProjectServiceImpl.java
```diff
@@ -144,4 +147,21 @@ public TimesheetProjectServiceImpl(
 
     return invoiceLineList;
   }
+
+  @Override
+  public TimesheetLine generateTimesheetLine(
+      Timesheet timesheet, ProjectPlanningTime projectPlanningTime) throws AxelorException {
+    TimesheetLine timesheetLine = super.generateTimesheetLine(timesheet, projectPlanningTime);
+    if (timesheetLine != null) {
+      if (timesheetLine.getTeamTask().getTimeInvoicing()
```
sos-axelor: The 'teamTask' field is not a required field, why there is no null check on it ?

PR #2885, opened 2018-12-31
RM#15591 Business support model : ProjectCategory/TeamTask objects -A…

jpo-axelor:
…dding defaultInvoicing field -Onchange TeamTask.projectCategory


axelor-project/src/main/resources/views/TeamTask.xml
```diff
@@ -412,4 +417,8 @@
 		<call class="com.axelor.apps.project.web.TeamTaskController" method="cancelTimer"/>
 	</action-method>
 	
+	<action-method name="action-team-task-method-set-default-values">
+		<call class="com.axelor.apps.businesssupport.web.TeamTaskController" method="setDefaultValues"/>
```
sos-axelor: method should be onChangeCategory and not setDafaultValues

axelor-business-support/src/main/java/com/axelor/apps/businesssupport/web/TeamTaskController.java
```diff
@@ -0,0 +1,49 @@
...
+import com.axelor.team.db.TeamTask;
+
+public class TeamTaskController {
+
+  public void setDefaultValues(ActionRequest request, ActionResponse response) {
+
+    TeamTask task = request.getContext().asType(TeamTask.class);
+
+    if (task.getProjectCategory() == null) {
+      return;
```
sos-axelor: don't return but add this condition to second if, means if projectCategory is null it will behave like projectCategory.defaultInvoicing =  'No'. Also remove null check of projectCategory from action

axelor-business-support/src/main/resources/domains/ProjectCategory.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<domain-models xmlns="http://axelor.com/xml/ns/domain-models"
+               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+               xsi:schemaLocation="http://axelor.com/xml/ns/domain-models http://axelor.com/xml/ns/domain-models/domain-models_5.0.xsd">
+
+	<module package="com.axelor.apps.project.db" name="project"/>
+
+	<entity name="ProjectCategory">
+		<integer name="defaultInvoicing" title="Default Invoicing" selection="support.default.invoicing.select"/>
```
sos-axelor: set default=1 for defaultInvoicing, else it will create issue. Also provide sql script to set default value.

axelor-project/src/main/resources/views/TeamTask.xml
```diff
@@ -124,7 +124,7 @@
 				</panel-related>
 			</panel>
 			<panel name="financialDataPanel" title="Financial Data" colSpan="12">
-				<field name="timeInvoicing" colSpan="4" onChange="action-team-task-record-set-invoicing-type" if="__config__.app.isApp('business-project')" if-module="axelor-business-project"/>
+				<field name="timeInvoicing" colSpan="4" onChange="action-team-task-record-set-invoicing-type" if="__config__.app.isApp('business-project')" if-module="axelor-business-project" readonlyIf="__config__.app.isApp('business-support') &amp;&amp; projectCategory == null"/>
```
sos-axelor: Why you have added the readonly condition ?

PR #2895, opened 2019-01-03
[Feature #15605] TeamTask object : change team-task-form

psh-axelor:


axelor-business-project/src/main/resources/views/TeamTask.xml
```diff
@@ -35,6 +35,8 @@
 	<action-record name="action-team-task-record-set-invoicing-type" model="com.axelor.team.db.TeamTask">
 		<field name="invoicingType" expr="eval: null"/>
 		<field name="toInvoice" expr="eval: false"/>
+		<field name="isOrderProposed" expr="eval: true" if="timeInvoicing"/>
```
sos-axelor: no need of 'if' condition and two fields should be done with one field

PR #2897, opened 2019-01-03
RM#15607 : TeamTask object : Update action on button name='addMultipl…

rme-axelor:
…eLinesBtn'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 ## Bug Fixes
 - TeamTask : Resolve NPE on save of new team task.
 - Studio: Fix import app without image. 
+- TeamTask : Resolve NPE by updating action of HR module.
 
 ## [Unreleased 5.1.1]
 ## Features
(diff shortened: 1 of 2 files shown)


axelor-human-resource/src/main/resources/views/ProjectPlanningTime.xml
```diff
@@ -114,6 +114,7 @@
 		<context name="_task" expr="eval:__self__" if="_model == 'com.axelor.team.db.TeamTask'"/>
 		<context name="_project" expr="eval:__self__.project" if="_model == 'com.axelor.team.db.TeamTask'"/>
 		<context name="_project" expr="eval:__self__" if="_model == 'com.axelor.apps.project.db.Project'"/>
+		<context name="_timeSpent" expr="eval:false"/>
```
sos-axelor: Same error is there when we create multiple planning lines from project. So don't change xml but change java code.

PR #2899, opened 2019-01-03
[Feature #15606] TeamTask object : Change label 'Task assigned to the provider'

psh-axelor:


axelor-project/src/main/resources/views/TeamTask.xml
```diff
@@ -347,6 +350,10 @@
     	<field name="project" expr="eval:_project" />
 	</action-record>
 	
+	<action-record name="action-team-task-business-support-provider" model="com.axelor.team.db.TeamTask">
+		<field name="$provider" expr="eval: __config__.app.getApp('business-support').provider.name"/>
```
sos-axelor: put null check for provider

PR #2934, opened 2019-01-11
Anomaly RM#15725 Project : Error InvoicingProject

rch-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/TimesheetProjectServiceImpl.java
```diff
@@ -133,7 +133,9 @@ public TimesheetProjectServiceImpl(
       Project project = (Project) timesheetInformations[5];
       PriceList priceList = project.getPriceList();
       if (consolidate) {
-        strDate = startDate.format(ddmmFormat) + " - " + endDate.format(ddmmFormat);
+        if (startDate != null && endDate != null) {
+          strDate = startDate.format(ddmmFormat) + " - " + endDate.format(ddmmFormat);
+        }
       } else {
         strDate = startDate.format(ddmmFormat);
```
sos-axelor: No date check in else part ? . Also check uses of that date, and handle the case when date is null.

PR #2940, opened 2019-01-11
RM#15703 MOVE : tax management

jpo-axelor:


sos-axelor:
Fixed by https://github.com/axelor/axelor-business-suite/pull/2939

PR #2944, opened 2019-01-14
dev : #14845 added check of move balance before saving

dda-axelor:


axelor-account/src/main/java/com/axelor/apps/account/db/repo/MoveManagementRepository.java
```diff
@@ -66,6 +67,7 @@ public Move copy(Move entity, boolean deep) {
   @Override
   public Move save(Move move) {
     try {
+      Beans.get(MoveValidateService.class).checkPreconditions(move);
```
gdu-axelor: Should be checked only if the status is daybook

gdu-axelor:
Changelog ?

PR #2949, opened 2019-01-16
[Feature #14633] MOVE : automatic tax line generation

psh-axelor:


axelor-account/src/main/java/com/axelor/apps/account/exception/IExceptionMessage.java
```diff
@@ -136,6 +136,8 @@
       "Account missing on the tax line : %s (company : %s)" /*)*/;
   static final String ANALYTIC_DISTRIBUTION_MISSING = /*$$(*/
       "Analytic distribution is missing on configuration for line : %s (company : %s)" /*)*/;
+  static final String TAX_LINE_ACCOUNTING_CONFIGURATION = /*$$(*/
```
sos-axelor: Use existing one MOVE_LINE_6

PR #2960, opened 2019-01-16
RM 15668 : possibility to add products en masse to bills of materials

wde-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -300,4 +303,59 @@ public void removeInvalidTree(List<Long> validBomIds, BillOfMaterial bom) {
...
+      priority += 10;
+      BillOfMaterial newComponent =
+          createBomFromRawMaterial(Long.valueOf((int) rawMaterial.get("id")), priority);
+      components.add(newComponent);
+    }
+
+    return components;
+  }
+
+  private BillOfMaterial createBomFromRawMaterial(long productId, int priority) {
```
gdu-axelor: Please don't use private method. Else we can't overrride it in another specific module.

axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -300,4 +303,59 @@ public void removeInvalidTree(List<Long> validBomIds, BillOfMaterial bom) {
...
+    BillOfMaterial newBom = new BillOfMaterial();
+    JPA.runInTransaction(
```
gdu-axelor: Transaction not required here because you return the components to the view. Moreover, BillOfmaterial change are not saved before run process

PR #2971, opened 2019-01-18
RM#15151 Fixed asset : invoice generated from purchase order

vbh-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/AccountManagementAccountService.java
```diff
@@ -57,4 +58,14 @@ public Account getProductAccount(
...
+
+  /**
+   * Get the product fixed asset category
+   *
+   * @param product
+   * @param company
+   * @return
+   * @throws AxelorException
+   */
+  public FixedAssetCategory getProductFixedAssetCateGory(Product product, Company company);
```
sos-axelor: It should 'getProductFixedAssetCategory', 'g' not 'G'.

PR #2974, opened 2019-01-18
Anomaly RM#12877 - EMAIL - Print option

rme-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/message/TemplateMessageServiceBaseImpl.java
```diff
@@ -130,6 +130,43 @@ public File generateBirtTemplate(
     }
   }
 
+  public String generateBirtFileLink(
+      TemplateMaker maker,
```
sos-axelor: Create a method for common code between generateBirtFileLink and generateBirtTemplate.

axelor-base/src/main/java/com/axelor/apps/base/service/message/TemplateMessageServiceBaseImpl.java
```diff
@@ -101,6 +101,60 @@ public File generateBirtTemplate(
...
+      throws AxelorException {
+
+    String birtTemplateFileLink = null;
+
+    ReportSettings reportSettings =
+        generateTemplate(maker, fileName, modelPath, format, birtTemplateParameterList);
+
+    if (reportSettings != null) {
+      try {
+        birtTemplateFileLink = reportSettings.generate().getFileLink();
```
sos-axelor: .generate() should be in common method

PR #2977, opened 2019-01-18
[Feature #15275] FINANCIAL ACCOUNT : new boolean analyticDistributionRequiredOnMoveLines

psh-axelor:


axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -421,6 +425,11 @@
 		<domain>self.account.id = :_id</domain>
 	</action-view>
 	
+	<action-attrs name="action-move-line-attrs-analytic-distribution-template">
+		<attribute name="readonly" for="analyticDistributionTemplate" expr="!account.analyticDistributionAuthorized"/>
```
sos-axelor: No need to have action for readonly, for required its ok as it depends on o2m . Use reaonlyIf directly

PR #2978, opened 2019-01-18
FACTOR : new menu in accounting

jpo-axelor:


axelor-account/src/main/resources/views/Menu.xml
```diff
@@ -249,7 +249,18 @@
...
 		<view type="form" name="debt-recovery-level-form"/>
 	</action-view>
+	
+	<menuitem name="debt-recovery-root-conf-factor"
+		parent="debt-recovery-root-conf" title="Factor"
+		action="debt.recovery.root.conf.factor" />
 
+	<action-view name="debt.recovery.root.conf.factor"
+		title="Factor" model="com.axelor.apps.base.db.Partner">
+		<view type="grid" name="partner-grid" />
```
sos-axelor: add cards  view for partner

PR #3021, opened 2019-01-25
Feature #15953 MANUF. ORDER : cancel reasons

sve-axelor:


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -429,12 +452,21 @@
...
+		<attribute for="cancelReasonStr" name="hidden" expr="cancelReason.freeText == false"/>
+	</action-attrs>
 
 	<action-validate name="action-manuf-order-validate-planned-end-date"
 	   model="com.axelor.apps.production.db.ManufOrder">
 	   <alert message="The planned end date is inferior to the actual date"
 	       if="plannedEndDateT.isBefore(__config__.app.getTodayDateTime().toLocalDateTime())"/>
 	</action-validate>
+	
+	<action-condition name="action-manuf-order-cancel-reason-check">
```
sos-axelor: No need to add this action, as its same as the one for sale order, we can use it directly.

axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -429,12 +452,21 @@
 	<action-attrs name="action-manuf-order-attrs-price-digits">
 		<attribute name="scale" for="costPrice" expr="eval: __config__.app.getNbDecimalDigitForUnitPrice()"/>
 	</action-attrs>
+	
+	<action-attrs name="action-attrs-manuf-order-other-reason">
```
sos-axelor: What is the use of this action ?

axelor-production/src/main/resources/i18n/messages_de.csv
```diff
@@ -1,6 +1,7 @@
 "key","message","comment","context"
 "105",,,
 "5",,,
+"A cancel reason must be selected","Ein Abbruchgrund muss ausgewählt werden",,
```
sos-axelor: No need of this translation as its already added in sale module.

PR #3040, opened 2019-01-28
Anomaly RM#15994:User View:changes on some fields

cma-axelor:


axelor-base/src/main/resources/views/User.xml
```diff
@@ -61,7 +62,10 @@
         </panel>
         <panel-tabs name="PermisionsPanelTab">
           <panel-related name="rolesPanel" field="roles" form-view="role-form" grid-view="role-grid"/>
-          <panel-related name="permissionsPanel" field="permissions" grid-view="simple-permission-grid" form-view="permission-form"/>
+          <panel name="permissionsPanel" title="Permissions">
```
sos-axelor: Change panel name, two panel must not have same name.

PR #3043, opened 2019-01-28
[Feature #15996] MANUFACTURING : new menu Workshop

psh-axelor:


axelor-stock/src/main/resources/views/StockLocation.xml
```diff
@@ -90,6 +90,8 @@
 	    <field name="typeSelect" expr="1"/>
 	    <field name="company"  expr="eval:__user__.activeCompany" if="__user__.activeCompany != null"/>
     	<field name="company"  expr="eval:__repo__(Company).all().fetchOne()" if="__user__.activeCompany == null &amp;&amp; __repo__(Company).all().fetch().size == 1"/>
+    	<field name="usableOnProduction" expr="eval:true" if="_productionWorkshop"/>
+    	<field name="isWorkshop" expr="eval:true" if="_productionWorkshop"/>
```
sos-axelor: No need of if condition because context value is boolean. Directly assign it.

PR #3059, opened 2019-01-29
Anomaly RM#16011 - STOCKMOVE : Map - NPE

rme-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/MapService.java
```diff
@@ -302,7 +302,8 @@ public String getDirectionUrl(
     LOG.debug("arrivalString = {}", aString);
     Map<String, Object> result = new HashMap<>();
     try {
-      if (BigDecimal.ZERO.compareTo(dLat) == 0 || BigDecimal.ZERO.compareTo(dLon) == 0) {
+      if ((dLat != null && BigDecimal.ZERO.compareTo(dLat) == 0)
```
sos-axelor: Do similar for aLat and aLon variables, also verify null check everywhere for bigdecimal in this method.

PR #3066, opened 2019-01-29
Dev 12587

fri-axelor:


axelor-marketing/src/main/resources/views/Campaign.xml
```diff
@@ -166,6 +167,12 @@
 		<domain>self.campaign.id = :_id AND self.statusSelect IN (2,3)</domain>
 	</action-view>
 	
+	 <action-view name="dashlet.generated.emails" title="Generated emails" model="com.axelor.apps.message.db.Message">
+    	<view type="grid" name="message-grid"/>
+    	<view type="form" name="message-form"/>
+    	<domain>self.relatedTo1SelectId = :id</domain>
```
gdu-axelor: Origin type should be checked here. Else we can see the message related to another object that has the same id.

PR #3071, opened 2019-01-31
[Anomaly #15106] Studio: Workflow roles management

psh-axelor:


axelor-studio/src/main/java/com/axelor/studio/service/wkf/WkfTransitionService.java
```diff
@@ -202,6 +202,10 @@ private void addButton(WkfTransition transition, String condition, Integer seque
     button.setWidgetAttrs("{\"colSpan\": \"3\"}");
     button.setOnClick(addButtonActions(transition, name));
 
+    Set<Role> buttonRoles = button.getRoles();
```
sos-axelor: always set new roles, not get existing one because in this way we can't remove old ones.

PR #3078, opened 2019-02-01
[Anomaly #16173] PICKING ORDER : add externalReference on printing

psh-axelor:


axelor-stock/src/main/resources/i18n/messages_de.csv
```diff
@@ -401,9 +400,11 @@
 "Partner product quality ratings","Bewertungen der Produktqualität durch Partner",,
 "Partner stock settings","Partnerbestandseinstellungen",,
 "Perishable expiration date","Verfallsdatum der verderblichen Ware",,
+"Picking printed","Bon de prépa édité",,
```
gdu-axelor: Why french translation appears here ?

psh-axelor: It's an ADK issue.
It was automatically generated on I18n build.

PR #3080, opened 2019-02-01
[Feature #16186] STOCK CONFIG : new boolean for picking order printing

psh-axelor:


axelor-stock/src/main/resources/domains/StockConfig.xml
```diff
@@ -40,6 +40,8 @@
 	<boolean name="displayLineDetailsOnPrinting" title="Display line details on printing"/>
 	<boolean name="isWithReturnSurplus" title="Return Surplus"/>
 	<boolean name="isWithBackorder" title="Manage backorder"/>
+	
+	<boolean name="pickingOrderPrinting" title="Picking order printing detailed"/>
```
gdu-axelor: boolean name is not specific enough, we can understand that the goal is to enable the picking order printing.

PR #3086, opened 2019-02-04
Feature RM#16191 - Template Task of Product : level 0 tree

rme-axelor:


axelor-business-project/src/main/resources/views/ProductTaskTemplate.xml
```diff
@@ -70,6 +70,10 @@
 
   </tree>
 
+  <action-attrs name="action-product-task-template-attrs-set-required">
+	<attribute name="required" for="parentTaskTemplate" expr="eval: __repo__(TaskTemplate).all().filter('self.product = ?',product).fetch().size()>0?true:false"/>
```
sos-axelor: Use count() instead of fetch()

PR #3115, opened 2019-02-06
RM#16307 MOVE : improve reversion process

vbh-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveService.java
```diff
@@ -459,7 +466,20 @@ public Move generateReverse(Move move) throws AxelorException {
...
 
-      if (validatedMove) {
+      if (isUnreconcileOriginalMove) {
+        List<Reconcile> reconcileList =
+            Beans.get(ReconcileRepository.class)
+                .all()
+                .filter("self.debitMoveLine = ?1 OR self.creditMoveLine = ?1", moveLine)
+                .fetch();
+        for (Reconcile reconcile : reconcileList) {
+          if (reconcile.getStatusSelect() != ReconcileRepository.STATUS_CANCELED) {
```
sos-axelor: Instead of checking with if, directly check on filter.

PR #3133, opened 2019-02-08
Feature RM#16300 - Project : New adress field

rme-axelor:


axelor-project/src/main/resources/views/Project.xml
```diff
@@ -538,7 +541,17 @@
...
+		<action name="action-project-attrs-customer-address-domain"/>
+	</action-group>
+
+	<action-group name="action-project-client-partner-on-change-group">
+		<action name="action-project-method-get-partner-data"/>
+		<action name="action-project-attrs-customer-address-domain"/>
 	</action-group>
+
+	<action-attrs name="action-project-attrs-customer-address-domain">
+		<attribute name="domain" for="customerAddress" expr="eval: (clientPartner?.partnerAddressList?.collect{it.id}?.size() &gt; 0) ? &quot;self.id IN (${clientPartner?.partnerAddressList?.collect{it.address.id}?.join(',')})&quot; : &quot;self.id IN (null)&quot;"/>
```
sos-axelor: Use sub query condition on address from PartnerAddress object to set domain. 
So no need of action-attrs and can be directly added on address field.

PR #3142, opened 2019-02-08
RM16426 : debt recovery no longer considering cancelled moves when se…

wde-axelor:
…arching for concerned invoices


axelor-account/src/main/java/com/axelor/apps/account/service/debtrecovery/DebtRecoveryService.java
```diff
@@ -228,7 +228,8 @@ public LocalDate getReferenceDate(DebtRecovery debtRecovery) {
         Move move = moveLine.getMove();
         // facture exigibles non bloquée en relance et dont la date de facture + délai
         // d'acheminement < date du jour
-        if (move.getInvoice() != null
+        if (move.getStatusSelect() != 4
```
gdu-axelor: Please, don't use hard coded value.

wde-axelor: oops :s

PR #3143, opened 2019-02-08
RM16426 : debt recovery no longer considering cancelled moves when se…

wde-axelor:
…arching for concerned invoices


wde-axelor:
PR #3142

PR #3145, opened 2019-02-11
Feature #16375 MANUF. ORDER : notify by automatic email when MO finis…

sve-axelor:
…hed or partially finished


axelor-supplychain/src/main/resources/views/AppSupplychain.xml
```diff
@@ -28,6 +28,18 @@
 			<field name="generateInvoiceFromStockMove" widget="boolean-switch" colSpan="4"/>
 			<field name="manageStockReservation" widget="boolean-switch" colSpan="4"/>
 		</panel>
+		<panel name="ManufOrderPanel" title="Manufacturing order">
```
sos-axelor: Naming convention issue 'manufOrderPanel' not 'ManufOrderPanel'.

PR #3149, opened 2019-02-11
RM#16452 PRODUCT : ProductTaskTemplate Remove Button

vbh-axelor:


axelor-business-project/src/main/resources/views/ProductTaskTemplate.xml
```diff
@@ -131,4 +122,8 @@
       method="removeTask"/>
   </action-method>
 
+  <action-validate name="action-product-task-template-validate-remove-task">
+  	<alert message="Do you really want to delete the selected record(s)?"/>
```
sos-axelor: Use prompt instead of action-validate

PR #3154, opened 2019-02-12
Feature RM#16172:SUPPL INVOICES:new buttons to validate

cma-axelor:


axelor-account/src/main/resources/views/AccountingSituation.xml
```diff
@@ -25,9 +25,10 @@
 			<field name="customerAccount" form-view="customer-account-form" grid-view="account-grid" canNew="true"/>
 			<field name="supplierAccount" form-view="supplier-account-form" grid-view="account-grid" canNew="true"/>
 			<field name="employeeAccount" form-view="employee-account-form" grid-view="account-grid" canNew="true" />
-			<field name="pfpValidatorUser" domain="self.activeCompany = :company and self.isPfpValidator = true" canEdit="false" showIf="company.accountConfig.isManagePassedForPayment"/>
+			<field name="pfpValidatorUser" domain=":company member of self.companySet and self.isPfpValidator = true" canEdit="false" showIf="company.accountConfig.isManagePassedForPayment &amp;&amp; partner.isSupplier"/>
```
sos-axelor: It will display the field if partner is not set yet, means when we create from o2m inside partner form.

PR #3183, opened 2019-02-15
Anomaly RM#16643 - CRM CONFIG : opportunity lost reasons with freeText

rme-axelor:


axelor-crm/src/main/resources/views/Opportunity.xml
```diff
@@ -231,7 +232,11 @@
     	<action name="action-opportunity-set-closed-lost"/>
     	<action name="save"/>
     </action-group>
-    
+
+  <action-attrs name="action-crm-opportunity-attrs-lost-reason-onchange">
+    <attribute name="hidden" for="lostReasonStr" expr="eval: !lostReason?.freeText"/>
```
sos-axelor: Use hideIf with dotted hidden field, instead of creating action and calling it on onChange.

PR #3186, opened 2019-02-18
Template customization

beuss:
This PR adds support for JasperReports in ReportSettings and provide new datastructure to be able to override default templates on a per user or per company basis. This will ease improvements of existing templates and, why not, allow community to provide customised templates.


beuss:
Note that since I'm not a gradle expert, this PR depends on https://github.com/axelor/abs-webapp/pull/27 to properly configure repositories.

PR #3187, opened 2019-02-18
Anomaly RM#16684 : MARKETING : Issues with target list filtering

rme-axelor:


rme-axelor:
Task Moved to dev branch

PR #3189, opened 2019-02-18
RM#16686 DEMO DATA : prod mode

vbh-axelor:


axelor-base/src/main/resources/views/AppBase.xml
```diff
@@ -81,4 +82,10 @@
 		<call class="com.axelor.apps.base.web.MailBatchController" method="remindTimesheetGeneral"/>
 	</action-method>
 	
+	<!-- ATCION ATTRS-->
+
+	<action-attrs name="action-app-base-attrs">
+		<attribute name="readonly" for="today" expr="eval: true" if="com.axelor.app.AppSettings.get().get('application.mode') == 'prod'"/>
```
sos-axelor: Better just hide the field ,put if condition directly on field.

PR #3196, opened 2019-02-19
RM#16622 CUSTOM MODEL : selection

jpo-axelor:


axelor-studio/src/main/resources/views/Menu.xml
```diff
@@ -80,4 +80,10 @@
         <domain>self.menuAction = false</domain>
     </action-view>
 
+	<menuitem title="Selections" name="app-builder-models-view-selections" parent="app-builder-config" action="app.builder.models.views.selections.all" order="46"/>
+	  <action-view name="app.builder.models.views.selections.all" title="Selections" model="com.axelor.meta.db.MetaSelect">
```
sos-axelor: No needs to create separate action , use the one used in 'Administration' menu.

PR #3206, opened 2019-02-21
[Anomaly #16696] MENU BUILDER

psh-axelor:


axelor-studio/src/main/resources/views/Selection.xml
```diff
@@ -88,4 +88,807 @@
       <option value="2">End node</option>
     </selection>
     
+    <selection name="studio.menu.builder.icon.select">
```
sos-axelor: Set 'icon' property on all options.

psh-axelor:
Updated on wip-5.3

PR #3216, opened 2019-02-25
RM#16705 MARKETING : Add a boolean in targetModelSet

vbh-axelor:


axelor-marketing/src/main/resources/views/Campaign.xml
```diff
@@ -134,6 +134,10 @@
   		<attribute  name="value" if="eval: team != null &amp;&amp; eventUser != null &amp;&amp; team != eventUser.activeTeam" expr="eval: null" for="eventUser"/>
 	</action-attrs>
 
+	<action-attrs name="action-campaign-attrs-is-allow-editing-of-targets-on-change">
+		<attribute name="value" for="manageAttendees" expr="eval: false" if="!isAllowEditingOfTargets"/>
```
sos-axelor: Use value of 'isAllowEditingOfTargets' in expr instead of using 'if'

PR #3229, opened 2019-02-26
RM#16515 SALE ORDER : notify when stock move is generated automatically from sale order

jpo-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/web/SaleOrderController.java
```diff
@@ -619,4 +619,24 @@ public void fillSaleOrderLinesEstimatedDate(ActionRequest request, ActionRespons
...
+    StockMoveRepository stockMoveRepo = Beans.get(StockMoveRepository.class);
+    StockMove stockMove =
+        stockMoveRepo
+            .all()
+            .filter(
+                "self.originTypeSelect = ?1 AND self.originId = ?2 AND self.statusSelect != ?3",
+                "com.axelor.apps.sale.db.SaleOrder",
+                saleOrder.getId(),
+                StockMoveRepository.STATUS_CANCELED)
+            .fetchOne();
```
sos-axelor: Use count instead of fetchOne

PR #3239, opened 2019-02-27
RM16700 : Fix NPE in the move line balance calculation

fri-axelor:


pdo-axelor:
Please use trackback service when id or moveLine is null. Those are abnormal situations we need to log so that we can invistigate further.

axelor-account/src/main/resources/i18n/messages_ru.csv
```diff
@@ -1768,9 +1792,15 @@
 "With Payment Rejected","С отклоненным платежом",,
 "Year","Год",,
 "Year or period or date is null, you must set a period.","Год, период или дата недействительны, вы должны установить период.",,
-"You can't remove this record, because the move has already been validated","Вы не можете удалить эту запись, потому что перемещение уже было подтверждено.",,
+"You can't remove this record, because the move %s has already been validated",,,
```
gdu-axelor: Please, don't commit these translation changes. See RM16308

PR #3243, opened 2019-02-28
RM#16860 STOCKMOVELINE : NPE

vbh-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/LogisticalFormLineServiceImpl.java
```diff
@@ -44,9 +44,11 @@
   @Override
   public BigDecimal getUnspreadQty(LogisticalFormLine logisticalFormLine) {
     StockMoveLine stockMoveLine = logisticalFormLine.getStockMoveLine();
-    return Beans.get(StockMoveLineService.class)
-        .computeSpreadableQtyOverLogisticalFormLines(
-            stockMoveLine, logisticalFormLine.getLogisticalForm());
+    return stockMoveLine != null
+        ? Beans.get(StockMoveLineService.class)
```
sos-axelor: Better do null check on method where its actually used that is computeSpreadableQtyOverLogisticalFormLines

PR #3244, opened 2019-02-28
RM#15196-Stock Location: fill company on creating inventory

ade-axelor:


axelor-stock/src/main/resources/views/Inventory.xml
```diff
@@ -140,6 +140,7 @@
 	    <field name="dateT" expr="eval:__config__.app.getTodayDateTime()"/>
 	    <field name="formatSelect" expr="pdf"/>
 	    <field name="stockLocation" expr="eval:_location"/>
+	    <field name="company" expr="eval:__repo__(StockLocation).find(_location.id).getCompany()"/>
```
sos-axelor: What if _location is null ?

PR #3264, opened 2019-03-06
Anomaly #16939 MANUF ORDER : Cancel reason shouldn't be required for …

sve-axelor:
…draft or planned manuf order


axelor-production/src/main/resources/views/ManufOrder.xml
```diff
@@ -252,6 +252,7 @@
   	  	<view-param name="popup-save" value="false"/>
 	  	<view-param name="forceEdit" value="true" />
 	  	<context name="_showRecord" expr="eval: __this__.id" />
+	  	<context name="_statusSelect" expr="eval: __this__.statusSelect" />
```
sos-axelor: what is the need to set '_statusSelect' context variable , its not working directly with 'statusSelect' ?

PR #3265, opened 2019-03-06
RM#16945 PO: Fix error display when fields in supplierCatalogList are…

aga-axelor:
… empty


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseOrderLineServiceImpl.java
```diff
@@ -261,8 +261,8 @@ public PurchaseOrderLine fill(PurchaseOrderLine line, PurchaseOrder purchaseOrde
 
     String productName = getProductSupplierInfos(purchaseOrder, line)[0];
     String productCode = getProductSupplierInfos(purchaseOrder, line)[1];
```
sos-axelor: Do changes on getProductSupplierInfos and no needs to call method two times.

PR #3267, opened 2019-03-06
RM#16142 CRM : partner stock settings

vbh-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/PartnerStockSettingsServiceImpl.java
```diff
@@ -63,4 +64,17 @@ public PartnerStockSettings createMailSettings(Partner partner, Company company)
     partner.addPartnerStockSettingsListItem(mailSettings);
     return Beans.get(PartnerStockSettingsRepository.class).save(mailSettings);
   }
+
+  @Override
+  public StockLocation getDefaultStockLocation(Partner partner, Company company) {
+
+    List<PartnerStockSettings> partnerStockSettingsList = partner.getPartnerStockSettingsList();
+
+    for (PartnerStockSettings partnerStockSettings : partnerStockSettingsList) {
```
sos-axelor: Use query instead of loop, and put null check on partner and company.

PR #3308, opened 2019-03-12
RM#17054 Configurator : Fix demo data

aga-axelor:


axelor-production/src/main/resources/demo/production-configurator-config.xml
```diff
@@ -126,8 +126,8 @@
       </bind>
 
       <bind node="configuratorProductFormulaList/configuratorProductFormula" to="configuratorProductFormulaList">
-        <bind node="metaField" to="metaField" search="self.id = :metaField" if="metaField">
-          <bind node="text()" to="id" alias="metaField"/>
+        <bind node="metaField" to="metaField" search="self.name = :metaField" if="metaField">
+          <bind node="text()" to="name" alias="metaField"/>
```
sos-axelor: Can't just search metaField with name, model search must be there to find correct one.

PR #3313, opened 2019-03-13
RM#16375 MANUF. ORDER : display message when no default email account…

sve-axelor:
… found to send email


axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderWorkflowService.java
```diff
@@ -219,8 +220,13 @@ public void finish(ManufOrder manufOrder) throws AxelorException {
     manufOrderRepo.save(manufOrder);
     AppProduction appProduction = Beans.get(AppProductionService.class).getAppProduction();
     if (appProduction != null && appProduction.getFinishMoAutomaticEmail()) {
+      if (Beans.get(EmailAccountRepository.class).all().filter("self.isDefault = true").fetchOne()
+          == null) {
+        return false;
+      }
       this.sendMail(manufOrder, appProduction.getFinishMoMessageTemplate());
```
sos-axelor: Check email account on this.sendMail method after template check in that method. So it will be common for partial finish and finish.

PR #3326, opened 2019-03-15
Anomaly #17153 SERVICES : stockManaged should be FALSE

rch-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/db/repo/ProductStockRepository.java
```diff
@@ -36,6 +36,15 @@
 
   @Inject private StockLocationLineService stockLocationLineService;
 
+  @Override
+  public Product save(Product product) {
+    product = super.save(product);
```
sos-axelor: Create onChange instead overriding save method

PR #3359, opened 2019-03-20
Anomaly #17261 PRODUCT : automatic sequence with generateProductSeque…

sve-axelor:
…nce = true

### DIFF
--- a/axelor-base/src/main/java/com/axelor/apps/base/db/repo/ProductBaseRepository.java
+++ b/axelor-base/src/main/java/com/axelor/apps/base/db/repo/ProductBaseRepository.java
@@ -47,6 +47,16 @@ public class ProductBaseRepository extends ProductRepository {
 
   @Override
   public Product save(Product product) {
+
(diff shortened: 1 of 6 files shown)


axelor-base/src/main/java/com/axelor/apps/base/db/repo/ProductBaseRepository.java
```diff
@@ -88,9 +87,9 @@ public Product save(Product product) {
     }
 
     try {
-      if (Strings.isNullOrEmpty(product.getSequence())
-          && appBaseService.getAppBase().getGenerateProductSequence()) {
-        product.setSequence(Beans.get(ProductService.class).getSequence());
+      if (appBaseService.getAppBase().getGenerateProductSequence()) {
+        product.setCode(Beans.get(ProductService.class).getSequence());
```
sos-axelor: It will set sequence each time whenever method 'save' is called.

PR #3372, opened 2019-03-22
[Feature #15275] Alert user about an analytic distribution settings

psh-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -1099,6 +1099,18 @@ InvoiceSet.length > 0))"
 		<call class="com.axelor.apps.account.web.InvoiceController" method="checkPartnerBankDetailsList"/>
 	</action-method>
 
+	<action-group name="action-invoice-group-invoice-ventilate-onclick">
+		<action name="save"/>
+		<action name="action-invoice-validate-analytical-distribution-template-prompt" 
+			if="invoiceLineList.find{it->!it.account.analyticDistributionAuthorized &amp;&amp; it.analyticDistributionTemplate != null} != null"/>
```
sos-axelor: Don't create action-validate but throw AxelorException on 'ventilate' method of InvoiceServiceImpl

PR #3388, opened 2019-03-27
RM#17389 SO line : Fix delivered qty not increase if stock move is in…

aga-axelor:
…ternal


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/StockMoveServiceSupplychainImpl.java
```diff
@@ -205,7 +205,9 @@ protected void updateSaleOrderLinesDeliveryState(StockMove stockMove, boolean qt
                 saleOrderLine.getProduct());
 
         if (qtyWasDelivered) {
-          saleOrderLine.setDeliveredQty(saleOrderLine.getDeliveredQty().add(realQty));
+          if (stockMove.getTypeSelect() != StockMoveRepository.TYPE_INTERNAL) {
+            saleOrderLine.setDeliveredQty(saleOrderLine.getDeliveredQty().add(realQty));
+          }
         } else {
           saleOrderLine.setDeliveredQty(saleOrderLine.getDeliveredQty().subtract(realQty));
```
sos-axelor: Check for subtract too.

PR #3419, opened 2019-04-02
Proposal #12903 Menu entry for MetaPermission object

sve-axelor:


axelor-admin/src/main/resources/views/Menu.xml
```diff
@@ -21,6 +21,15 @@
     <view type="form" name="meta-group-menu-assistant-form" />
   </action-view>
   
+  <menuitem name="menu-meta-permission" parent="menu-auth" title="Meta Permissions"
+    action="action-meta-permission" />
+
+  <action-view name="action-meta-permission" title="Meta Permissions"
```
sos-axelor: Don't create new but, use same action as the one used in administration menu.

PR #3449, opened 2019-04-05
Anomaly #17638 INVENTORY : formatSelect selection should be unique

sve-axelor:


axelor-stock/src/main/resources/views/Selects.xml
```diff
@@ -126,5 +126,11 @@
 		<option value="com.axelor.apps.production.db.ManufOrder" data-grid="manuf-order-grid" data-form="manuf-order-form">Manufacturing order</option>
 		<option value="com.axelor.apps.production.db.OperationOrder" data-grid="operation-order-grid" data-form="operation-order-form">Operation order</option>
 	</selection>
+	
+	<selection name='inventory.format.select'>
```
sos-axelor: name should be 'inventory.export.format.select'

PR #3495, opened 2019-04-12
Feature #17828 STOCKMOVE LINE : add tracking

sve-axelor:


axelor-stock/src/main/resources/views/StockMoveLine.xml
```diff
@@ -327,6 +330,9 @@
             <field name="natureOfTransaction"/>
             <field name="countryOfOrigin"/>
         </panel>
+        <panel-mail name="mailPanel">
+        	<mail-messages />
```
sos-axelor: add default limit="4", also add followers

PR #3503, opened 2019-04-15
RM#17220 Pack / SaleOrder

vbh-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/db/repo/SaleOrderManagementRepository.java
```diff
@@ -64,6 +65,7 @@ public SaleOrder copy(SaleOrder entity, boolean deep) {
   @Override
   public SaleOrder save(SaleOrder saleOrder) {
     try {
+      Beans.get(SaleOrderComputeService.class).computePackTotal(saleOrder);
```
sos-axelor: This method should be called only if feature is enabled.

PR #3529, opened 2019-04-17
RM17879: in SaleOrder, display error message when total Amount = 0 an…

jch-axelor:
…d saleOrderLineList is not empty.


axelor-sale/src/main/resources/views/SaleOrder.xml
```diff
@@ -1261,7 +1261,7 @@
 	</action-method>
 
 	<action-validate name="action-sale-order-validate-check-on-confirm">
-		<alert message="The total amount is zero." if="inTaxTotal &lt;= 0"/>
+		<alert message="The total amount of quotation is zero. Do you want to proceed ?" if="inTaxTotal &lt;= 0 &amp;&amp; !saleOrderLineList.isEmpty()"/>
 	</action-validate>
```
gdu-axelor: better to check on exTaxTotal field. Also, if we have a negative amount, we shouldn't display this message.

PR #3532, opened 2019-04-18
RM #17639 DEBT RECOVERY : Hide 'Customer recovery' button if there is…

sve-axelor:
… no debt recovery for Partner

### DIFF
--- a/axelor-base/src/main/resources/views/Partner.xml
+++ b/axelor-base/src/main/resources/views/Partner.xml
@@ -102,7 +102,7 @@
 
 	<form name="partner-customer-form"  title="Customer" model="com.axelor.apps.base.db.Partner"
 		  onNew="action-group-base-partner-onnew"
-		  onLoad="action-group-partner-onload"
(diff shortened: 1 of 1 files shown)


axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -778,6 +778,11 @@
 	<action-attrs name="action-partner-attrs-set-domain">
 		<attribute name="domain" for="partner" expr="eval: &quot; (self.isCustomer = true OR self.isSupplier = true OR self.isProspect = true) AND self.isEmployee = false &quot;"/>
 	</action-attrs>
+	
+	<action-attrs name="action-partner-attrs-hide-customer-recovery-btn">
+		<attribute name="hidden" for="customerRecoveryBtn" expr="eval: true" if="accountingSituationList.every{accountingSituation -> accountingSituation.debtRecovery == null}"/>
```
sos-axelor: can be done with single line, just move If condition to 'expr'

PR #3538, opened 2019-04-19
Anomaly #17378 PURCHASE REQUEST : Generate PO from Purchase Request

sve-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseRequestServiceImpl.java
```diff
@@ -53,4 +68,104 @@ public void acceptRequest(List<PurchaseRequest> purchaseRequests) {
       purchaseRequestRepo.save(purchaseRequest);
     }
   }
+
+  @Transactional
+  @Override
+  public ActionViewBuilder generatePo(
```
sos-axelor: Better return PO list and do ActionViewBuilder on controller, so that this method can be used by other service/method.

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/PurchaseRequestServiceSupplychainImpl.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  protected PurchaseOrder getPoBySupplierAndDeliveryAddress(
+      PurchaseRequest purchaseRequest, List<PurchaseOrder> purchaseOrderList) {
+
+    PurchaseOrder purchaseOrder =
+        super.getPoBySupplierAndDeliveryAddress(purchaseRequest, purchaseOrderList);
+    List<StockLocation> stockLocations =
+        stockLocationRepo
+            .all()
+            .filter("self.address = ?1", purchaseRequest.getDeliveryAddress())
```
sos-axelor: It will not give any error if purchaseRequest.getDeliveryAddress() is null ?

PR #3545, opened 2019-04-22
[Anomaly #17927] ICalendar : error message

psh-axelor:


axelor-base/src/main/java/com/axelor/apps/base/ical/ICalendarService.java
```diff
@@ -688,6 +688,13 @@ protected ICalendar doSync(
           ICalendarStore.getModifiedEventsInRange(
               collection, lastSynchro, allRemoteUids, startDate, endDate);
     }
+
+    if (events.isEmpty() || events == null) {
```
sos-axelor: it should be "events == null || events.isEmpty()" (always put proper order in expression)

PR #3558, opened 2019-04-25
[Anomaly #17957] EMPLOYEE : Newly created employee already has timesheets

psh-axelor:


axelor-human-resource/src/main/resources/views/Employee.xml
```diff
@@ -408,7 +408,7 @@
 				 model="com.axelor.apps.hr.db.Timesheet">
 		<view type="grid" name="my-timesheets-grid"/>
 		<view type="form" name="timesheet-form"/>
-        <domain>self.user = :__user__</domain>
+        <domain>self.user = :user AND self.user.employee = :id</domain>
```
sos-axelor: only 'self.user.employee = :id' is enough

psh-axelor:
No longer needed as updated on dev.

PR #3567, opened 2019-04-26
RM#17260 PARTNER : Link to employee form

jpo-axelor:


axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -787,6 +799,10 @@
 	<action-attrs name="action-partner-attrs-hide-customer-recovery-btn">
 		<attribute name="hidden" for="customerRecoveryBtn" expr="eval: accountingSituationList.every{accountingSituation -> accountingSituation.debtRecovery == null}"/>
 	</action-attrs>
+	
+	<action-attrs name="action-partner-attrs-hide-employee-panel">
+		<attribute name="hidden" for="employeePanel" expr="eval:!( __this__.employee != null &amp;&amp; __user__.employee != null  &amp;&amp; (__user__.employee.contactPartner == __this__) )"/>
```
sos-axelor: Not just current user but employee manager should be able to see the button.

PR #3583, opened 2019-04-30
RM#18049 EMPLOYEE : form view

ade-axelor:


axelor-human-resource/src/main/resources/views/Employee.xml
```diff
@@ -266,7 +265,11 @@
     <action-method name="action-employee-method-normalize-emergencyNumber">
         <call class="com.axelor.apps.base.web.PartnerController" method="normalizePhoneNumber"/>
     </action-method>
-
+    
+	<action-validate name="action-employee-validate-dates">
```
sos-axelor: Use action-condition instead of action-validate , check existing action-condition for reference.

PR #3587, opened 2019-05-01
RM#17260 PARTNER : Link to employee form

jpo-axelor:
Previous pull req. was closed by mistake
so your previous PR comment was :
"Not just current user but employee manager should be able to see the button."


jpo-axelor:
Previous pull req. was closed by mistake
so previous PR comment was :
"Not just current user but employee manager should be able to see the button."

PR #3598, opened 2019-05-03
Anomaly RM#17860 - BONUS MGT : Error when computing amounts

rme-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/EmployeeBonusService.java
```diff
@@ -174,6 +178,14 @@ public String replaceExpressionInFormula(
                   employeeService.getDaysWorksInPeriod(
                       employee, period.getFromDate(), period.getToDate())));
     }
+
+    // For checking that formula is replaced by variables or not.
+    if (tempForumla.equals(formula)) {
```
sos-axelor: throw error only  if formula contains any variable, check pattern for '$*$'.

PR #3627, opened 2019-05-13
RM18205: Generate Invoice from sale order : take in consideration cla…

jch-axelor:
…ssic invoice only


gdu-axelor:
Changelog file change is missing

PR #3635, opened 2019-05-15
Anomaly #14931 MRP: Add the start date time and end date time

dna-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/MrpServiceImpl.java
```diff
@@ -206,6 +207,7 @@ protected void finish(Mrp mrp) {
     log.debug("Finish MRP");
 
     mrp.setStatusSelect(MrpRepository.STATUS_CALCULATION_ENDED);
+    mrp.setEndDateTime(Beans.get(AppBaseService.class).getTodayDateTime().toLocalDateTime());
```
sos-axelor: Dont use Beans.get if service is already injected on constructor

PR #3639, opened 2019-05-15
RM#18226 SaleOrderLine : Allow to change quantity even if sale order …

vbh-axelor:
…is invoiced or delivered when Allow Pending Order Modification is true.

### DIFF
--- a/axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderLineServiceSupplyChainImpl.java
+++ b/axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderLineServiceSupplyChainImpl.java
@@ -20,6 +20,7 @@ package com.axelor.apps.supplychain.service;
 import com.axelor.apps.account.db.AnalyticDistributionTemplate;
 import com.axelor.apps.account.db.AnalyticMoveLine;
 import com.axelor.apps.account.db.repo.AnalyticMoveLineRepository;
+import com.axelor.apps.account.db.repo.InvoiceRepository;
 import com.axelor.apps.account.service.AnalyticMoveLineService;
 import com.axelor.apps.account.service.app.AppAccountService;
 import com.axelor.apps.base.db.Partner;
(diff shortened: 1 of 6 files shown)


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderLineServiceSupplyChainImpl.java
```diff
@@ -188,4 +191,37 @@ public void updateDeliveryState(SaleOrderLine saleOrderLine) {
...
+  public BigDecimal checkInvoicedOrDeliveredOrderQty(SaleOrderLine saleOrderLine) {
+    BigDecimal qty = saleOrderLine.getQty();
+    BigDecimal deliveredQty = saleOrderLine.getDeliveredQty();
+    BigDecimal invoicedQty = BigDecimal.ZERO;
+
+    List<InvoiceLine> invoiceLineList =
+        Beans.get(InvoiceLineRepository.class)
+            .all()
+            .filter(
+                "self.invoice.statusSelect = ? AND self.saleOrderLine.id = ?",
```
sos-axelor: Better use  sum function directly with query.

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderLineServiceSupplyChainImpl.java
```diff
@@ -188,4 +191,37 @@ public void updateDeliveryState(SaleOrderLine saleOrderLine) {
...
+    if (!invoiceLineList.isEmpty()) {
+      invoicedQty =
+          invoiceLineList
+              .stream()
+              .map(invoiceLine -> invoiceLine.getQty())
+              .reduce(BigDecimal::add)
+              .get();
+    }
+
+    return invoicedQty.compareTo(BigDecimal.ZERO) > 0
```
sos-axelor: don't make return statement complex.

PR #3651, opened 2019-05-17
RM #17378 PURCHASE REQUEST : Fix generating PO for null stock location

sve-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseRequestServiceImpl.java
```diff
@@ -161,4 +150,8 @@ protected PurchaseOrder createPurchaseOrder(PurchaseRequest purchaseRequest)
             purchaseRequest.getSupplierUser(),
             null));
   }
+
+  protected String getKey(PurchaseRequest purchaseRequest) {
```
sos-axelor: put more specific name for method like 'getPurchaseOrderGroupBySupplierKey'

PR #3657, opened 2019-05-21
RM #17952 AVAILABITY REQUEST : error message if availableStock < realQty

sve-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveServiceImpl.java
```diff
@@ -1253,4 +1253,10 @@ public void setPickingStockMovesEditDate(List<Long> ids, String userType) {
       }
     }
   }
+
+  @Override
+  public BigDecimal getAvailableStock(StockMove stockMove, StockMoveLine stockMoveLine) {
+    // defined in supplychain
+    return BigDecimal.ZERO;
```
sos-axelor: will not work when supplychain module is not installed, so instead of '0' return 'stockMoveLine.realqty'

PR #3658, opened 2019-05-21
[Anomaly #18303] MARKETING : Precise domain for model in message template

psh-axelor:


axelor-base/src/main/resources/views/Template.xml
```diff
@@ -70,5 +70,9 @@
 	 <action-method name="action-template-check-receptor">
     	<call class="com.axelor.apps.base.web.TemplateController" method="checkTargetReceptor"/>
     </action-method>
-    
+
+	<action-attrs name="action-template-attrs-meta-model-domain">
+		<attribute name="domain" for="metaModel" expr="eval: &quot; self.name = 'Partner' OR self.name = 'Lead' &quot;" if="_mediaType == 4"/>
```
sos-axelor: action must be defined inside marketing module only

PR #3659, opened 2019-05-21
RM18348: Create an object to store streets.

dtr-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -271,4 +277,42 @@ public String computeAddressStr(Address address) {
...
+            .bind("country", country)
+            .fetch();
+    City city = cities.size() == 1 ? cities.get(0) : null;
+    address.setCity(city);
+    address.setAddressL6(city != null ? zip + " " + city.getName() : null);
+
+    List<Street> streets =
+        streetRepository
+            .all()
+            .filter("self.city = :city AND self.country = :country")
```
gdu-axelor: I don't think that we have a field "country" in the Street object

PR #3680, opened 2019-05-27
Anomaly RM#18457 CRM : lead description in opportunity description

pia-axelor:


axelor-crm/src/main/resources/views/Opportunity.xml
```diff
@@ -122,6 +122,9 @@
             <panel name="descriptionPanel" title="Description">
                 <field name="description" showTitle="false" readonlyIf="salesStageSelect == 5 || salesStageSelect == 6" colSpan="12" widget="html"/>
             </panel>
+            	<panel name="customerDescriptionPanel" title="Customer Description">
+            		<field name="customerDescription"  colSpan="12" widget="html"/>
```
sos-axelor: put showTitle="false"

axelor-crm/src/main/resources/views/Opportunity.xml
```diff
@@ -261,6 +265,14 @@
     	<field name="description" expr="eval: _partner.description"/>
     </action-record>
     
+    <action-attrs name="action-opportunity-attrs-description-panel-show">
```
sos-axelor: use 'If' condition instead of 'action-attrs' like used for all other apps setting check.

axelor-crm/src/main/resources/views/Opportunity.xml
```diff
@@ -261,6 +265,14 @@
...
+    	<attribute name="hidden" for="customerDescriptionPanel" expr="eval:!(__config__.app.getApp('crm').displayCustomerDescriptionInOpportunity)"/>
+    </action-attrs>
+    
+    <action-attrs name="action-opportunity-attrs-from-customer">
```
sos-axelor: don't add onChange action for customer but change action when opportunity is created from customer

PR #3683, opened 2019-05-27
Anomaly RM#18460 DELIVERY CONDITION : hide mention in BIRT if deliver…

pia-axelor:
…yCondition = null


axelor-sale/src/main/resources/reports/SaleOrder.rptdesign
```diff
@@ -3565,7 +3565,7 @@ row["ClientPartFirstName"]
                         <list-property name="visibility">
                             <structure>
                                 <property name="format">all</property>
-                                <expression name="valueExpr" type="javascript">row._outer["DisplayDeliveryCondition"] == 0</expression>
+                                <expression name="valueExpr" type="javascript">row._outer["deliveryCondition"] == null</expression>
```
sos-axelor: condition must be on both displayDelCondOnPrinting and deliveryCondition != null

PR #3691, opened 2019-05-28
Anomaly RM#16643 : Display freetext if it is not null in Opportunity

rme-axelor:


axelor-crm/src/main/resources/views/Opportunity.xml
```diff
@@ -136,7 +136,8 @@
                 <field if="__config__.app.getApp('base').getTeamManagement()" name="team" canEdit="false" form-view="team-form" grid-view="team-grid"/>
                 <field if="__config__.app.getApp('base').enableTradingNamesManagement" name="tradingName" canEdit="false" domain=":company MEMBER OF self.companySet" form-view="trading-name-form" grid-view="trading-name-grid"/>
 				<field if="__config__.app.getApp('base').getManageMultiBanks()" if-module="axelor-cash-management" name="bankDetails" widget="SuggestBox" onSelect="action-accounting-batch-attrs-bank-details-domain" form-view="bank-details-form" grid-view="bank-details-grid"/>
-				<field name="lostReason" readonly="true" showIf="lostReason != null"/>
+				<field name="lostReason" readonly="true" showIf="lostReason != null &amp;&amp; !lostReasonStr"/>
+				<field name="lostReasonStr" readonly="true" showIf="lostReason != null &amp;&amp; lostReasonStr"/>
```
sos-axelor: set full colspan as its a text field, also display both at same place right now place get changed when lostReasonStr displays

PR #3700, opened 2019-05-29
RM18495: TIMESHEET confirm now update total

tga-axelor:


axelor-human-resource/src/main/resources/views/Timesheet.xml
```diff
@@ -189,12 +189,14 @@
 	</action-attrs>
 
 	<action-group name="action-timesheet-group-confirm">
+		<action name="action-timesheet-method-timesheet-compute-period-total"/>
```
gdu-axelor: Better to call this method directly from the confirm method in order to avoid to have some WS for one operation.

axelor-human-resource/src/main/java/com/axelor/apps/hr/web/timesheet/TimesheetController.java
```diff
@@ -383,6 +383,10 @@ public void confirm(ActionRequest request, ActionResponse response) {
       timesheet = timesheetRepositoryProvider.get().find(timesheet.getId());
       TimesheetService timesheetService = timesheetServiceProvider.get();
 
+      BigDecimal periodTotal = timesheetServiceProvider.get().computePeriodTotal(timesheet);
```
gdu-axelor: You can use in same WS method the response.setReload and the response.setValue

PR #3714, opened 2019-05-31
5.o dev 18526

tga-axelor:


axelor-account/src/main/resources/views/InvoiceLineTax.xml
```diff
@@ -11,7 +11,7 @@
     
     <form name="invoice-line-tax-form" title="Tax Line" model="com.axelor.apps.account.db.InvoiceLineTax"  >
     	<panel name="main" >
-    		<field name="taxLine" grid-view="tax-line-grid" form-view="tax-line-form"/>
+    		<field name="taxLine" canEdit="false"grid-view="tax-line-grid" form-view="tax-line-form"/>
```
ale-axelor: There should be a space here, or the xml won't be valid.

PR #3742, opened 2019-06-05
RM18336 : REPORTING new indicator for qualified leads on Lead Dashboa…

tga-axelor:
…rd One.


axelor-crm/src/main/resources/views/Dashboard.xml
```diff
@@ -239,6 +239,7 @@
 		<dashlet action="dashlet.team.leads" height="350"/>		
 		<dashlet action="dashlet.created.leads.per.month" height="350" />
 		<dashlet action="dashlet.created.leads.by.industry.sector" height="350" />
+		<dashlet action="dashlet.lead.new.opportunities" height="350"/>
```
gdu-axelor: no need to use action-view here. You can just call the chart like that  : 		
`<dashlet action="chart:chart.lead.new.ooportunity" height="350"/>
`

tga-axelor: Got it.

PR #3752, opened 2019-06-07
Anomaly #18591 PACKING LIST : display productName instead of product …

dna-axelor:
…on printing


axelor-supplychain/src/main/resources/reports/PackingList.rptdesign
```diff
@@ -3069,7 +3093,7 @@ default:
                                             <structure>
                                                 <property name="paramName">param_1</property>
                                                 <simple-property-list name="expression">
-                                                    <value type="javascript">'value:'+row["product_description"]</value>
```
sos-axelor: remove 'product_description' from query too, if not used anywhere

PR #3758, opened 2019-06-11
RM#10992 PRODUCTION ORDER : plan

jpo-axelor:


axelor-business-production/src/main/java/com/axelor/apps/businessproduction/service/ProductionOrderWizardServiceBusinessImpl.java
```diff
@@ -89,6 +89,13 @@ public Long validate(Context context) throws AxelorException {
       startDateT = appProductionService.getTodayDateTime();
     }
 
+    if (context.containsKey("_endDate") && context.get("_endDate") != null) {
+      endDateT =
+          ZonedDateTime.parse(
```
sos-axelor: why used ZonedDateTime.parse ?, LocalDateTime.parse is not working ?

PR #3764, opened 2019-06-11
RM18629 : EMPLOYEE CARD : add tag for rh, company and department field

tga-axelor:


axelor-human-resource/src/main/resources/views/Employee.xml
```diff
@@ -180,9 +183,14 @@
 			  	<span>
 				<strong>{{contactPartner.fullName}}</strong><br/>
 			    FP : {{contactPartner.fixedPhone}}<br/>
-			    MP : {{contactPartner.mobilePhone}}
+			    MP : {{contactPartner.mobilePhone}}<br/>
```
gdu-axelor: PF, MP, Co., Dept. should be translated

PR #3774, opened 2019-06-12
RM#18617 EMPLOYEE : chart on training

jdo-axelor:
…mployee


axelor-talent/src/main/resources/views/Employee.xml
```diff
@@ -29,5 +29,9 @@
 		<domain>self.employee = :__self__</domain>
 	</action-view>
 	
+	<action-view name="action-chart-employee-training-per-category" title="Training per category" model="com.axelor.apps.hr.db.Employee">
```
gdu-axelor: I think no need the action-view here

jdo-axelor: I used an action view in order to inject the employee id to the chart by the context, i don't know how to doing it directly in the action-dashlet or in the chart. If you know a solution you are welcome if not I'll search tomorrow morning for an another solution.

jdo-axelor: So yes, it's as you wrote the action view is not needed.

PR #3817, opened 2019-06-19
RM11964 : make use of helper

tga-axelor:


ale-axelor:
For this case I think we do not need to update the changelog because it has no incidence on the end user.

PR #3837, opened 2019-06-25
Anomaly #17954 SALE ORDER : line not editable if availability request…

dna-axelor:
… = true


axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -196,8 +196,8 @@
 				<field name="$companyCurrency" readonly="true" title="Currency" type="many-to-one" target="com.axelor.apps.base.db.Currency"/>
 			</panel>
 			<panel name="deliveryPanel" title="Delivery">
-				<field name="desiredDelivDate"/>
-				<field name="estimatedDelivDate"/>
+				<field name="desiredDelivDate" readonlyIf="$availabiltyRequest == true"/>
```
sve-axelor: No need to write == true, just write readonlyIf="$availabiltyRequest"

axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -196,8 +196,8 @@
...
+				<field name="estimatedDelivDate" readonlyIf="$availabiltyRequest == true"/>
```
sve-axelor: No need to write == true, just write readonlyIf="$availabiltyRequest"

PR #3858, opened 2019-06-28
Feature #18903 ACCOUNTINGREPORT : Downloads accounting export attache…

dna-axelor:
…ment for TypeSelect 'Export journal entry -> Administration' on Export.


axelor-account/src/main/java/com/axelor/apps/account/web/AccountingReportController.java
```diff
@@ -173,11 +175,24 @@ public void printExportMoveLine(ActionRequest request, ActionResponse response)
...
+                  .add(
+                      "html",
+                      "ws/rest/com.axelor.meta.db.MetaFile/"
+                          + accesssFile.getId()
+                          + "/content/download?v="
+                          + accesssFile.getVersion())
+                  .param("download", "true")
+                  .map());
+        } else {
+          moveLineExportService.exportMoveLine(accountingReport);
```
sve-axelor: No need to add same lines 2 times. Call this method outside if.

PR #3915, opened 2019-07-10
RM#19134 CONFORMITY CERTIFICATE : error on printing

jdo-axelor:


jdo-axelor:
Update CC and Picking stock move as in CITEL with in addition tracking number

PR #3927, opened 2019-07-11
Feature RM#18822 LOGISTICAL FORM: Update net mass

pia-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveServiceImpl.java
```diff
@@ -84,6 +84,9 @@
   private StockMoveToolService stockMoveToolService;
   private StockMoveLineRepository stockMoveLineRepo;
 
+  @Inject
+  private ProductRepository productRepository;
+
```
rme-axelor: Add ProductRepository in the constructor

PR #3961, opened 2019-07-16
RM#19365 Studio : Fix error when click on wkf buttons

aga-axelor:


axelor-studio/src/main/java/com/axelor/studio/web/WkfController.java
```diff
@@ -255,4 +258,20 @@ public void setDefaultNodes(ActionRequest request, ActionResponse response) {
 
     return Collections.emptyList();
   }
+
+  public void track(ActionRequest request, ActionResponse response) {
+    try {
+      Long id = (Long) request.getContext().get("id");
+
+      MetaJsonRecord record = Beans.get(MetaJsonRecordRepository.class).find(id);
```
sos-axelor: Check null for id  not for record, it will give error if no id is there.

PR #4002, opened 2019-07-23
Anomaly RM#19555: USER: login with a user account that is not valided.

msh-axelor:


axelor-base/src/main/resources/data-init/input-config.xml
```diff
@@ -57,7 +57,10 @@
     
     <input file="meta_metaMenu.csv" separator=";" type="com.axelor.meta.db.MetaMenu" search="self.name = :name and self.module = 'axelor-core'" update="true" />
     
-    <input file="auth_user.csv" separator=";" type="com.axelor.auth.db.User"  search="self.code = :code" />
+    <input file="auth_user.csv" separator=";" type="com.axelor.auth.db.User"  search="self.code = :code">
+        <bind to="blocked" column=":blocked" />
```
rch-axelor: No need to specify this binding in config as column name is same as in domain. 
And If specified no need to suffix it with colon(:)

PR #4005, opened 2019-07-23
RM#19579:status behavior of bank order now is like the behavior of a …

jdo-axelor:
…payment and when the bank order status also change all the associate payment status


axelor-bank-payment/src/main/resources/views/BankOrder.xml
```diff
@@ -286,8 +286,8 @@
     		<button name="confirmBtn" title="Confirm" showIf="statusSelect == 1" onClick="action-bank-order-group-confirm"/>
     		<button name="signBtn" title="Sign" showIf="statusSelect == 2" onClick="save,action-bank-order-method-sign"/>
     		<button name="carryOutBtn" title="Carry out" showIf="statusSelect == 3" onClick="save,action-bank-order-method-realize,save"/>
-    		<button name="cancelBtn" title="Cancel" showIf="statusSelect == 1" onClick="save,action-bank-order-record-cancel,save"/>
-			<button name="correctBtn" title="Correct" showIf="statusSelect == 2 || statusSelect == 3 || statusSelect == 6" onClick="save,action-bank-order-record-correct,save"/>
+    		<button name="cancelBtn" title="Cancel" showIf="statusSelect == 1 || statusSelect == 2 || statusSelect == 3" onClick="save,action-bank-order-method-cancel"/>
```
ale-axelor: This button is updated but in contrary with the cancel button above it has no prompt attribute. Is this normal ?

PR #4006, opened 2019-07-24
Anomaly #19535 STOCK CONFIG : customsPanel

dna-axelor:


axelor-stock/src/main/resources/views/AppStock.xml
```diff
@@ -11,13 +11,12 @@
 			<field name="isRequiredShipmentSupplierDetails" widget="boolean-switch" colSpan="4"/>
 		</panel> 
 		<panel name="actionPanel">
+			<panel name="customsPanel" title="Customs">
```
rme-axelor: Keep outside of actionPanel

PR #4023, opened 2019-07-26
RM#19453 Advanced Import : trim data before import

aga-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/advanced/imports/DataImportServiceImpl.java
```diff
@@ -291,16 +291,16 @@ private void initializeVariables() {
...
+      if (Strings.isNullOrEmpty(dataRow[cell])) {
         dataList.add(
             !Strings.isNullOrEmpty(fileField.getDefaultIfNotFound())
-                ? fileField.getDefaultIfNotFound()
+                ? fileField.getDefaultIfNotFound().trim()
                 : "");
         continue;
       }
 
+      String dataCell = dataRow[cell].trim();
```
sos-axelor: It will be never null ?

PR #4038, opened 2019-07-30
RM#19698 ACCOUNTING : general ledger new columns on BIRT

ade-axelor:


axelor-account/src/main/resources/reports/AccountingReportType1.rptdesign
```diff
@@ -131,8 +131,9 @@
                 </ex-property>
             </list-property>
             <property name="odaDriverClass">org.postgresql.Driver</property>
-            <property name="odaURL">jdbc:postgresql://localhost:5432/axelor-business-suite</property>
+            <property name="odaURL">jdbc:postgresql://localhost:5432/axelor-business-suite-dev</property>
```
cma-axelor: never put credentials like this

PR #4114, opened 2019-08-27
RM20249 : ACCOUNTING : period status control on move validation

srab-axelor:


axelor-account/src/main/java/com/axelor/apps/account/exception/IExceptionMessage.java
```diff
@@ -171,6 +171,8 @@
   static final String MOVE_9 = /*$$(*/ "Tax is mandatory for the account %s" /*)*/;
   static final String MOVE_10 = /*$$(*/
       "Analytic distribution template is mandatory for the account %s." /*)*/;
+  static final String MOVE_11 = /*$$(*/
```
ale-axelor: A lot of constant in this interface are named following the pattern "{classname}_{number}" but there is no real reason to name it that way. It's an error from old legacy code. So for new constants, it's better to use a meaningful name instead of MOVE_11.

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateService.java
```diff
@@ -207,6 +207,11 @@ public void validate(Move move, boolean updateCustomerAccount) throws AxelorExce
     log.debug("Validation de l'écriture comptable {}", move.getReference());
 
     this.checkPreconditions(move);
+
```
ale-axelor: Please apply spotless

PR #4130, opened 2019-08-30
RM20348: Refactor message module.

dtr-axelor:


axelor-message/src/main/java/com/axelor/apps/message/service/TemplateMessageServiceImpl.java
```diff
@@ -182,7 +182,7 @@ public Message generateMessage(Long objectId, String model, String tag, Template
     Message message =
         messageService.createMessage(
             model,
-            Long.valueOf(objectId).intValue(),
+            objectId.intValue(),
```
ale-axelor: Below you replaced the same expression with `Math.toIntExact(objectId)`, which expression is the best ?

dtr-axelor: In this case, `objectId` is a `Long` whereas in `GenerateMessageController` it is a `long`. Since `Math.toIntExact(objectId)` can be used in both case and is safer, we should use it in both cases.

https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#toIntExact-long-

ale-axelor: Can you do it please ? :)

dtr-axelor: Yes I can :)

PR #4140, opened 2019-09-03
Anomaly #20347 MANUF ORDER : copy MO but not operations orders

dna-axelor:


axelor-business-production/src/main/java/com/axelor/apps/businessproduction/db/repo/ManufOrderBusinessProductionManagementRepository.java
```diff
@@ -0,0 +1,15 @@
+package com.axelor.apps.businessproduction.db.repo;
```
rme-axelor: add header in new java files

PR #4151, opened 2019-09-05
Feature #19544 STOCK RULE : message template

dna-axelor:


axelor-message/src/main/resources/views/GenerateMessageWizard.xml
```diff
@@ -34,4 +35,8 @@
 		<attribute for="_xTemplate" name="value" expr="eval: null"/>
 	</action-attrs>
 
+	<action-attrs name="action-attrs-template-value">
```
rme-axelor: action name is different

PR #4153, opened 2019-09-05
RM#20449 Expense : Move.period / Move.date

jpo-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/expense/ExpenseServiceImpl.java
```diff
@@ -349,12 +349,11 @@ public Move ventilate(Expense expense) throws AxelorException {
   }
 
   protected Move createAndSetMove(Expense expense) throws AxelorException {
-    LocalDate moveDate = expense.getMoveDate();
+    LocalDate moveDate = expense.getPeriod().getToDate();
```
jpo-axelor: In expense, period is required to fill on validate btn, so no null check added.

PR #4164, opened 2019-09-05
RM#19814: in expense  add the payment move and its generation and cor…

jdo-axelor:
…rected expense bank order, now if a expense bank order is cancel the expense is also cancel, warning add move service interface and change move service name to move service implementation


axelor-human-resource/src/main/java/com/axelor/apps/hr/web/expense/ExpenseController.java
```diff
@@ -341,7 +341,7 @@ public void addPayment(ActionRequest request, ActionResponse response) {
       expenseServiceProvider.get().addPayment(expense);
       response.setReload(true);
     } catch (Exception e) {
-      TraceBackService.trace(e);
+      TraceBackService.trace(response, e);
       response.setException(e);
```
ale-axelor: Please remove setException here.

PR #4171, opened 2019-09-09
RM#18336 REPORTING : number of qualified leads

srab-axelor:


ale-axelor:
Please update changelog.

PR #4172, opened 2019-09-09
RM#18417 REPORTING : number of new customers

srab-axelor:


ale-axelor:
Please update changelog.

PR #4201, opened 2019-09-14
Feature #20509 EBICSUSER : replace the O2M ebicsRequestLogList per a …

dna-axelor:
…dashlet


axelor-bank-payment/src/main/resources/views/EbicsUser.xml
```diff
@@ -162,6 +162,11 @@
     	<view-param name="show-toolbar" value="false"/>
     </action-view>
     
+    <action-view name="action-ebics-user-view-ebics-request-log" title="Request log" model="com.axelor.apps.bankpayment.db.EbicsRequestLog">
+        <view type="grid" name="ebics-request-log-grid"/>
+        <view type="form" name="ebics-request-log-form"/>
+    </action-view>
```
rch-axelor: Specify domain to display related ebics user's Request Log only.

PR #4218, opened 2019-09-17
RM#20787: corrected bank order, now manage multiple expense

jdo-axelor:


jdo-axelor:
Correction for ticket 19814 and closed PR 4164

PR #4261, opened 2019-09-24
RM20977: remove empty mrp in progress due to stock details by product

jch-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/ProjectedStockServiceImpl.java
```diff
@@ -68,6 +69,13 @@
             .order("id")
             .fetch();
 
+    if (mrpLineList.isEmpty()) {
+      List<MrpLine> mrpLineListToDelete =
+          Beans.get(MrpLineRepository.class).all().filter("self.mrp = ?1", mrp, product).fetch();
```
ale-axelor: I think you forgot to remove "product" in the filter argument list, as the filter only needs 1 argument.

PR #4274, opened 2019-09-26
RM#10692 Deleting deprecated interfaces

srab-axelor:


axelor-talent/src/main/resources/domains/JobPosition.xml
```diff
@@ -31,6 +31,8 @@
 	   			public static final Integer STATUS_ON_HOLD = 2;
 	   			public static final Integer STATUS_CLOSED = 3;
 	   			public static final Integer STATUS_CANCELED = 3;
+	   			
+	   			public static final String SEQUENCE_CODE_JOB_POSITION = "jobPosition";
```
gdu-axelor: Better to move it into Sequence.xml file

srab-axelor: Done

PR #4280, opened 2019-09-26
cost sheet: properly convert purchase units

beuss:


beuss:
@gdu-axelor @axeloradmin Candidate for 5.1.14?

PR #4281, opened 2019-09-27
Anomaly RM#20821: Advance data import : Add action apply support

cma-axelor:


axelor-base/src/main/resources/domains/FileTab.xml
```diff
@@ -21,6 +21,9 @@
     <many-to-one name="validationLog" ref="com.axelor.meta.db.MetaFile" title="Validation log"/>
     
     <many-to-one name="advancedImport" ref="com.axelor.apps.base.db.AdvancedImport" title="Advanced import"/>
+    
+    <string name="actions" title="Actions"/>
```
sos-axelor: Use large="true"

PR #4304, opened 2019-10-01
RM20851 : Fix computed fields NPE from all domains

nsw-axelor:


axelor-base/src/main/resources/domains/TaxLine.xml
```diff
@@ -17,7 +17,7 @@
 	    	String name = "";
 	    	if(tax != null)  { name += tax.getName();  }
 	    	name += " : " + value;
-			name += " : " + startDate;
+			if(startDate !=null) {name += " : " + startDate;}
```
vbh-axelor: startDate is already required field

axelor-account/src/main/resources/domains/IrrecoverableInvoiceLine.xml
```diff
@@ -14,7 +14,7 @@
  	
  	<string name="name" title="Name">
  		<![CDATA[
-	    	if (invoice != null) return invoice.getInvoiceId() + " - " + invoiceLineSeq;
+	    	if (invoice != null && invoice.getInvoiceId() != null) return invoice.getInvoiceId() + " - " + invoiceLineSeq;
 	    	else return invoiceLineSeq.toString();
```
vbh-axelor: Add null check for invoiceLineSeq

PR #4327, opened 2019-10-07
Feature RM#21323 EMPLOYEE : Add dates to files management

pda-axelor:


axelor-base/src/main/resources/domains/FileType.xml
```diff
@@ -8,6 +8,7 @@
 	<entity name="FileType">
 		<string name="code" title="Code" />
 		<string name="name" title="Name" />
+		<boolean name="hasExpirationDate" title="Display an expiration date" default="false" />
```
cma-axelor: remove default false

axelor-base/src/main/resources/i18n/messages_fr.csv
```diff
@@ -1,1650 +1,1653 @@
-"key","message","comment","context"
-"${fullName} variants",,,
-"%D : Day","%D : Jour",,
+key,message,comment,context
```
cma-axelor: double quotes  should not be removed in translation file please take care of this things

PR #4395, opened 2019-10-16
Feature #20359 META-MODEL-FORM : add tracking on json fields

dna-axelor:


axelor-studio/src/main/java/com/axelor/studio/web/MetaJsonModelController.java
```diff
@@ -52,19 +57,29 @@ public void openWorkflow(ActionRequest request, ActionResponse response) {
...
 
     } catch (Exception e) {
       TraceBackService.trace(response, e);
     }
   }
+
+  public void fieldTrackingOnMetaModel(ActionRequest request, ActionResponse response) {
+    MetaJsonField metaJsonField = request.getContext().asType(MetaJsonField.class);
+    MetaJsonField metaJsonFieldName =
+        Beans.get(MetaJsonFieldRepo.class).findByName(metaJsonField.getName());
```
aga-axelor: If create field with same name but different model, it will not add in tracking

PR #4417, opened 2019-10-21
Anomaly #21712 Distance travelled calculation of kilometricExpenseLin…

pda-axelor:
…e with Open street map


axelor-human-resource/src/main/java/com/axelor/apps/hr/web/expense/ExpenseController.java
```diff
@@ -632,51 +632,57 @@ public void computeDistanceAndKilometricExpense(ActionRequest request, ActionRes
       throws AxelorException {
 
     // Compute distance.
+    try {
```
cma-axelor: Remove unnecessary changes due to white spaces.

PR #4433, opened 2019-10-23
RM#21843 CIVIL YEAR : Allowing periods generation using wizard on the…

srab-axelor:
… Civil Year


axelor-base/src/main/java/com/axelor/apps/base/web/YearController.java
```diff
@@ -0,0 +1,25 @@
...
+import com.axelor.apps.base.service.YearService;
+import com.axelor.exception.AxelorException;
+import com.axelor.exception.service.TraceBackService;
+import com.axelor.rpc.ActionRequest;
+import com.axelor.rpc.ActionResponse;
+import com.google.inject.Inject;
+
+public class YearController {
+
+  @Inject private YearService yearService;
```
ale-axelor: Please inject using Beans.get in controller

axelor-account/src/main/resources/views/Year.xml
```diff
@@ -88,7 +76,7 @@
     	<call class="com.axelor.apps.account.web.YearController" method="adjust"/>
     </action-method>
     
-    <action-validate name="action-year-validate-generate-period">
+    <action-validate id="account-action-year-validate-generate-period" name="action-year-validate-generate-period">
 	    <error message="To generate periods, please fill in the limits of the fiscal year." if="fromDate == null || toDate == null" />
```
ale-axelor: I think we can delete this and keep the action validate in base, to avoid code duplication.

PR #4438, opened 2019-10-24
Fix NullPointerException in FixedAssetServiceSupplyChainImpl#createFixedAssets

beuss:
Looks like some logic was inverted


ale-axelor:
Thanks for the fix.
All our pull requests need to update the changelog, except if it fixes a feature that appeared in the current version or if it is only code refactoring so I added an entry.

beuss:
> Thanks for the fix.
> All our pull requests need to update the changelog, except if it fixes a feature that appeared in the current version or if it is only code refactoring so I added an entry.

Yes, I was lazy and used the github edit file feature this time :-)

PR #4485, opened 2019-11-06
RM#21554: MOVE LINE : in reverse process, analytic is now copy from o…

jdo-axelor:
…rigin move line to reverse move line.


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveServiceImpl.java
```diff
@@ -527,6 +532,19 @@ protected MoveLine generateReverseMoveLine(
...
+    reverseMoveLine = this.copyAnalytic(orgineMoveLine, reverseMoveLine);
+    return reverseMoveLine;
+  }
+
+  protected MoveLine copyAnalytic(MoveLine origineMoveLine, MoveLine reverseMoveLine) {
+    reverseMoveLine.setAnalyticDistributionTemplate(
+        origineMoveLine.getAnalyticDistributionTemplate());
+    for (AnalyticMoveLine origineAnalyticMoveLine : origineMoveLine.getAnalyticMoveLineList()) {
+      AnalyticMoveLine copyAnalyticMoveLine =
+          analyticMoveLineRepository.copy(origineAnalyticMoveLine, true);
```
ale-axelor: I think the parameter "deep" here needs to be false instead to avoid unnecessary copies

PR #4490, opened 2019-11-07
Feature #22768: ADDRESS : when one address on partner please treat it…

msh-axelor:
… as a default address


axelor-base/src/main/java/com/axelor/apps/base/service/PartnerServiceImpl.java
```diff
@@ -425,6 +425,9 @@ public void addContactToPartner(Partner contact) {
   protected Address getAddress(Partner partner, String querySpecific, String queryComman) {
 
     if (partner != null) {
+      if (partner.getPartnerAddressList().size() == 1) {
```
sos-axelor: instead of using o2m , use query as like done in following steps

PR #4494, opened 2019-11-07
RM#22684: MOVE LINE : reconciled or lettered move lines are now read …

jdo-axelor:
…only and MOVE : if a move line is reconciled or lettered display message.


axelor-account/src/main/java/com/axelor/apps/account/web/MoveController.java
```diff
@@ -254,4 +255,23 @@ public void filterPartner(ActionRequest request, ActionResponse response) {
       response.setAttr("partner", "domain", domain);
     }
   }
+
+  public void isHiddenMoveLineListViewer(ActionRequest request, ActionResponse response) {
+
+    Move move = request.getContext().asType(Move.class);
+    boolean isHidden = true;
+    try {
+      if (move.getMoveLineList() != null && move.getStatusSelect() < 3) {
```
ale-axelor: Please use constant for status select value.

PR #4497, opened 2019-11-08
RM#21375: ACCOUNT CONFIG/ACCOUNTING REPORT: add new field line minimu…

jdo-axelor:
…m in account configuration used in accouting report for new message that appear before generation if the minimum of line is found. and ACCOUNTING REPORT : corrected result of search button in function of the report.


axelor-account/src/main/java/com/axelor/apps/account/service/AccountingReportService.java
```diff
@@ -67,4 +67,6 @@
   public BigDecimal getCreditBalance(AccountingReport accountingReport, String queryFilter);
 
   public BigDecimal getCreditBalanceType4();
+
+  public boolean isToManyLine(AccountingReport accountingReport) throws AxelorException;
```
ale-axelor: isToManyLine => isThereTooManyLines

PR #4535, opened 2019-11-18
purchases: fix computation of reverse charged taxes

beuss:
TaxEquiv has to be fetched using original tax line not the mapped one,
as for invoices

Changelog entry (not in commit to avoid stupid merges back & forth)
purchase orders: Correctly compute reverse charged tax lines


beuss:
@gdu-axelor @axeloradmin  Candidate for 5.1.14?

PR #4536, opened 2019-11-18
Accounting: correctly use invoice date for currency conversion

beuss:
Invoices can be created long time after they've been emitted. General accounting rules states that we should use the invoice date to compute values in company currency, not today's date.

As usual; no changelog in commit to ease merge…
Accounting: use invoice date (if filled) to compute company in/ex tax total


beuss:
@gdu-axelor  @axeloradmin Candidate for 5.1.14?

PR #4553, opened 2019-11-20
Feature #23109 : Phone widget

nsw-axelor:


axelor-studio/src/main/resources/views/MenuBuilder.xml
```diff
@@ -49,7 +49,7 @@
 	      	<field name="moduleToCheck" />
 	      	<field name="top" />
 	      	<field name="left" />
-	      	<field name="mobile" />
+	      	<field name="mobile" widget="phone"/>
```
vbh-axelor: Keep mobile field as it is. Remove phone widget.

axelor-account/src/main/resources/views/DebtRecovery.xml
```diff
@@ -63,8 +63,8 @@
         		<panel-related name="partnerAddressListPanel" field="accountingSituation.partner.partnerAddressList" colSpan="12" form-view="partner-address-form" >
         			<field name="address" form-view="address-form" grid-view="address-grid"/>
         		</panel-related>
-        		<field name="accountingSituation.partner.fixedPhone" colSpan="6" />
-        		<field name="accountingSituation.partner.fax" colSpan="6" />
+        		<field name="accountingSituation.partner.fixedPhone" widget="phone" colSpan="6" />
+        		<field name="accountingSituation.partner.fax" widget="phone" colSpan="6" />
```
vbh-axelor: I think we don't need phone widget for this dotted as it will be readonly.

PR #4557, opened 2019-11-20
Anomaly #23084 : STOCK RULE : manuf. order

sma-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/StockRulePurchaseOrder.java
```diff
@@ -0,0 +1,12 @@
+package com.axelor.apps.production.service;
```
rme-axelor: add header in java files.

PR #4562, opened 2019-11-21
Feature RM#23080 : ADVANCEDEXPORT : Sort ascending or descending

sma-axelor:


axelor-base/src/main/resources/i18n/messages_en.csv
```diff
@@ -98,18 +98,18 @@
 "ABC classes",,,
 "ARTMIN",,,
 "AZTEC",,,
-"AbcAnalysis.analysisResult","Analysis result",,
```
rme-axelor: Why removed this translations?

PR #4627, opened 2019-12-05
RM#23390: INVOICE : corrected the possibility to add a payment when the amount remaining is inferior or equal to 0.

jdo-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentValidateServiceImpl.java
```diff
@@ -121,6 +126,16 @@ public void validate(InvoicePayment invoicePayment, boolean force)
     invoicePaymentRepository.save(invoicePayment);
   }
 
+  protected void checkConditionBeforeValidate(InvoicePayment invoicePayment)
+      throws AxelorException {
+    if (invoicePayment.getInvoice().getAmountRemaining().compareTo(BigDecimal.ZERO) <= 0) {
```
ale-axelor: Please check if it is possible to have invoice payment without invoice, because if not it will throw a NPE.

ale-axelor:
see internal issue because we need to find another solution.

PR #4686, opened 2019-12-16
RM 23048: Added base configuration to set number of digits on quantit…

wde-axelor:
…ies fields throughout the supplychain modules


wde-axelor:
Lacking Changelog + please add significant number of digits to birt reports for quantitie based on configuration.

PR #4729, opened 2019-12-23
RM#23981 MetaSchedule :ContractBatch

ssa-axelor:


axelor-contract/src/main/java/com/axelor/apps/contract/batch/service/BatchContractService.java
```diff
@@ -0,0 +1,23 @@
+package com.axelor.apps.contract.batch.service;
```
vbh-axelor: always add license header in new java file

PR #4769, opened 2020-01-08
Anomaly #RM24445 : MRP forecast : Reset status on copy

ple-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/db/repo/MrpForecastManagementRepository.java
```diff
@@ -0,0 +1,13 @@
+package com.axelor.apps.supplychain.db.repo;
```
vbh-axelor: Always add licence header in new java file

PR #4811, opened 2020-01-16
RM#23390: INVOICE : corrected the possibility to add a payment when t…

jdo-axelor:
…he amount remaining is inferior or equal to 0.


jdo-axelor:
It's in the saveBtn in InvoicePayment just before save.

PR #4855, opened 2020-01-24
RM16428: display advise message when batch has been already launch wi…

jch-axelor:
…th the same move date


axelor-supplychain/src/main/resources/views/SupplychainBatch.xml
```diff
@@ -81,12 +81,13 @@
 
 	<!-- ACTION ATTRS -->
 	<action-attrs name="action-supplychain-batch-attrs-show-label">
-		<attribute name="hidden" expr="eval: !(moveDate &amp;&amp; batchList &amp;&amp; batchList?.find{it.startDate.month == __date__.month &amp;&amp; it.startDate.year == __date__.year} != null)" for="adviseUser"/>
+		<attribute name="hidden" expr="eval: !(moveDate &amp;&amp; batchList &amp;&amp; batchList?.find{it.moveDate?.day == moveDate.day &amp;&amp; it.moveDate?.month == moveDate.month &amp;&amp; it.moveDate?.year == moveDate.year &amp;&amp; it.accountingCutOffTypeSelect == accountingCutOffTypeSelect} != null)" for="adviseUser"/>
```
ale-axelor: Simply use `==` (in groovy, `==` is the same as  `equals()` in java) to check if two localdates are equals.

PR #4872, opened 2020-01-29
Anomaly RM#24922 : INVOICE : Issue when we ventilate a refund invoice…

ple-axelor:
… linked to a stock move


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/workflow/WorkflowVentilationServiceSupplychainImpl.java
```diff
@@ -254,14 +254,18 @@ private void stockMoveProcess(Invoice invoice) throws AxelorException {
     // update qty invoiced in stock move line
     for (InvoiceLine invoiceLine : invoice.getInvoiceLineList()) {
       StockMoveLine stockMoveLine = invoiceLine.getStockMoveLine();
+      StockMove stockMove = stockMoveLine.getStockMove();
```
vbh-axelor: It will be better if we add this line inside if block after null check of stockMoveLine

PR #4949, opened 2020-02-17
RM 25522,25512,25517

wde-axelor:


wde-axelor:
Do not merge yet.
First draft of RM 25522 (without agencyList), 2551 and 25517

PR #4983, opened 2020-02-25
RM24716 Removing the persistable field on lead-form view

erg-axelor:
Also realigned a bit by removing a spacer


ale-axelor:
Please add changelog

PR #4986, opened 2020-02-26
RM#21666 NPE fix & chart domain linked to user

mga-axelor:
RM#21666

* NPE fixed with double security.
 -> fields stay hidden if user is not linked with any partner
 -> onClick on fields now show a prepared error message and no more NPE

* Translations added in french (& english) for the prepared error message

* Chart now shows only invoices related to the user's partner


axelor-client-portal/src/main/resources/i18n/messages_fr.csv
```diff
@@ -57,5 +57,6 @@
 "Tasks in progress","Tâches en cours",,
 "Total projects","Total des projets",,
 "Total remaining","Total restant",,
+"User is not linked with any partner. Please disconnect.","L'utilisateur n'est relié à aucun tier. Veuillez vous déconnecter.",,
```
ale-axelor: typo: "tier" => "tiers"

PR #5020, opened 2020-03-03
RM#26227: add account filter to summary and gross report

jdo-axelor:


axelor-account/src/main/resources/reports/AccountingReportType17.rptdesign
```diff
@@ -511,6 +512,7 @@ WHERE
    AND (AccountingReport.date_to IS NULL OR Asset.acquisition_date <= AccountingReport.date_to)
    AND Asset.company = AccountingReport.company
    AND (AccountingReport.analytic_distribution_template IS NULL OR Asset.analytic_distribution_template = AccountingReport.analytic_distribution_template)
+   AND(AccountSet.account_set IS NULL OR AccountSet.account_set = Account.id)
```
gdu-axelor: Please, apply the same query like done on other report. We should be able to select the account 20100, or 20, or 2 and filter on children accounts

ale-axelor: will be done in next PR

PR #5032, opened 2020-03-03
RM#25664 : Fix hidden invoicesCopySelect when partner is supplier

mga-axelor:
The field invoicesCopySelect now corectly appear when the partner is a supplier

RM#25664


axelor-account/src/main/resources/views/Partner.xml
```diff
@@ -6,7 +6,7 @@
 	<form name="incl-invoice-payment-partner-form" title="Partner" model="com.axelor.apps.base.db.Partner" onLoad="action-partner-attrs-show-specific-tax-note">
 	    <panel name="mainPanel" colSpan="12">
 	    	<field name="invoiceSendingFormatSelect" hideIf="isEmployee"/>
-	        <field name="invoicesCopySelect" hideIf="!isCustomer || isEmployee"/>
+	        <field name="invoicesCopySelect" hideIf="(!isCustomer || isEmployee) &amp;&amp; !isSupplier"/>
```
ale-axelor: if isEmployee is true we have to hide the field regardless of other booleans.

PR #5048, opened 2020-03-04
RM#25898 : Manuf. order bugfix. No longer diplay unavaible tracking numbers

mga-axelor:
On manuf-order-form > consumedProductsPanel > consumedStockMoveLineList > "split by tracking numbers" button
No longer displays tracknig numbers when avaibleQty equals 0

RM#25898 (cherry-pick from 5.2, pull request #5047)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,11 @@
 - Fix Timesheet Reminder Batch sendReminder method
 - Stock Move Line reservation: correctly set qty requested flag when generated from a sale order line.
 - Stock Move: Delete empty date field in form view.
+<<<<<<< HEAD
 - Advance data import: Fix search issue on main object to import.
 - LEAD : removed the persistable field on the form view
 - LEAVEREQUEST : Fix the NPE when no leaveRequest is selected to be edited
+- MANUFACTURING ORDER : On consumed product, no longer diplay tracking numbers if avaible quantity equals 0.
 
 ## [5.3.0] - 2020-02-25
 ## Features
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,9 +11,11 @@
 - Fix Timesheet Reminder Batch sendReminder method
 - Stock Move Line reservation: correctly set qty requested flag when generated from a sale order line.
 - Stock Move: Delete empty date field in form view.
+<<<<<<< HEAD
```
ale-axelor: please don't forget to check the result after a merge is done.

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -1200,7 +1200,7 @@ public void setAvailableStatus(StockMoveLine stockMoveLine) {
             + stockMoveLine.getProduct().getId()
             + " AND self.id in (select stockLocationLine.trackingNumber.id from StockLocationLine stockLocationLine join StockLocation sl on sl.id = stockLocationLine.detailsStockLocation.id WHERE sl.id = "
```
ale-axelor: Since you changed this method, please also improve its readability:
- The first line of the method is not necessary, we can declare `trackingNumbers` variable as we set it.
- Split this line so we do not have a line longer than 100 characters (to follow our coding convention)

axelor-stock/src/main/java/com/axelor/apps/stock/web/StockMoveLineController.java
```diff
@@ -262,10 +262,7 @@ public void displayAvailableTrackingNumber(ActionRequest request, ActionResponse
       StockLocationLine detailStockLocationLine =
           stockLocationLineService.getDetailLocationLine(
               stockMove.getFromStockLocation(), stockMoveLine.getProduct(), trackingNumber);
-      BigDecimal availableQty =
-          detailStockLocationLine != null
```
ale-axelor: Please keep this check. I agree that we should always find a detailStockLocationLine here but if there is an issue in the future it's better to avoid a NPE.

PR #5051, opened 2020-03-05
RM#26253 : Mass invoicing : sort generated lines by stock move

mga-axelor:
Because the invoices lines are generated sequentialy, orderBy field is now on ID on invoiceLineList

RM#26253


wde-axelor:
Must keep using sequence and not id. The sequence of generated lines must be set in the java code on generation as discussed together.

PR #5075, opened 2020-03-09
RM#25946 : Add leave lines configuration menu in leave management

mga-axelor:
New menu : leaveManagement -> configuration -> all employees's leave lines
Bugfix : leaveLine's fields employee & leaveReason are now required
Add new views corresponding to the new menu
Add French & English translation for the menu title

RM#25946


axelor-human-resource/src/main/resources/views/LeaveLine.xml
```diff
@@ -20,8 +28,19 @@
...
+		</panel>
+		<panel-include view="leave-line-form"/>
+	</form>
+	
 	<action-method name="action-leave-line-method-compute-quantity-available">
 		<call class="com.axelor.apps.hr.web.leave.management.LeaveManagementController" method="computeQuantityAvailable"/>
 	</action-method>
 	
+	<action-record name="action-leave-line-record-set-name" model="com.axelor.apps.hr.db.LeaveLine">
+		<field name="name" expr="eval: leaveReason.leaveReason"/>
```
ale-axelor: onChange is also triggered when you empty a field, so here I get a NPE if I empty the leaveReason field.
You can simply use `?.` to empty the name when leaveReason is emptied.

PR #5079, opened 2020-03-10
RM26403 Renamed DateOfHire field as hireDate

erg-axelor:


ale-axelor:
Please update changelog

PR #5091, opened 2020-03-11
RM#26313 : Signature on user & template

mga-axelor:
User's new field: emailSignature.
On templates (from message) add a new formula field: the email signature.
This formula field may point on a user's emailSignature directly or from an emailAccount's signature.

When an email is created from template, the signature is automaticly add at the end of the email's content.
If the formula field was empty or incorrectly filled, the default emailAccount's signature is automaticly add at the end of the mail instead.

RM#26313 (main features commit, missing the autofill of the formula field)


axelor-base/src/main/java/com/axelor/apps/base/service/message/MessageServiceBaseImpl.java
```diff
@@ -79,23 +82,37 @@ public Message createMessage(
       Set<MetaFile> metaFiles,
       String addressBlock,
       int mediaTypeSelect,
-      EmailAccount emailAccount) {
+      EmailAccount emailAccount,
+      String signature) {
 
+    emailAccount =
+        emailAccount != null
+            ? Beans.get(EmailAccountRepository.class).find(emailAccount.getId())
```
ale-axelor: Replace this expression by
```
if (emailAccount != null) {
  emailAccount = Beans.get(EmailAccountRepository.class).find(emailAccount.getId());
}
```
Which is easier to understand

PR #5097, opened 2020-03-11
RM24763 Fixed NPE when company has no linked partner on saleOrder form

erg-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/web/SaleOrderController.java
```diff
@@ -714,7 +714,7 @@ public void updateSaleOrderLineTax(ActionRequest request, ActionResponse respons
   public void getSaleOrderPartnerDomain(ActionRequest request, ActionResponse response) {
     SaleOrder saleOrder = request.getContext().asType(SaleOrder.class);
     Company company = saleOrder.getCompany();
-    long companyId = company.getPartner().getId();
+    Long companyId = company.getPartner() == null ? null : company.getPartner().getId();
```
ale-axelor: This is ok but just to be safe please rename companyId => companyPartnerId, keep it a long and set it to 0 instead of null

PR #5115, opened 2020-03-12
RM20115 default language is based on application.ppties for users created automatically

erg-axelor:
Default language is based on application.ppties even if the user created automatically from java process


axelor-base/src/main/java/com/axelor/apps/base/db/repo/UserBaseRepository.java
```diff
@@ -31,6 +32,10 @@
   @Override
   public User save(User user) {
     try {
+      if (user.getId() == null || (user.getId() != null && user.getVersion() == 0)) {
```
ale-axelor: Please check if language is not already set here. Else if the user creates a new user from the form view and sets manually a language, this language will be overwritten on save.

PR #5137, opened 2020-03-16
RM26521 : Improve naming & translation of Target & Target Configuration

aol-axelor:


axelor-crm/src/main/resources/i18n/messages_fr.csv
```diff
@@ -449,6 +448,7 @@
 "Source description","Description provenance",,
 "Start","Démarrer",,
 "Start date","Date de début",,
+"Starting date",,,
```
ale-axelor: Please translate this new key

PR #5218, opened 2020-03-26
RM26675 : Add totals for an amount type move template

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/MoveTemplateController.java
```diff
@@ -134,4 +137,10 @@ public void setIsValid(ActionRequest request, ActionResponse response) {
       }
     }
   }
+
+  public void computeTotals(ActionRequest request, ActionResponse response) {
+    MoveTemplate moveTemplate = request.getContext().asType(MoveTemplate.class);
```
ale-axelor: In all java controller, every process must be inside a : try { // Something } catch(Exception e) { TraceBackService.trace(response, e); }.

PR #5260, opened 2020-04-02
RM27058 : Fix email not being duplicated when creating a partner from…

aol-axelor:
… a lead in an opportunity


axelor-crm/src/main/java/com/axelor/apps/crm/service/OpportunityServiceImpl.java
```diff
@@ -67,6 +69,8 @@ public Partner createClientFromLead(Opportunity opportunity) throws AxelorExcept
           addressService.createAddress(
               null, null, lead.getPrimaryAddress(), null, addressL6, lead.getPrimaryCountry());
       address.setFullName(addressService.computeFullName(address));
+
+      email = new EmailAddress(lead.getEmailAddress().getAddress());
```
ale-axelor, line 73: The if condition is checking primaryAddress, but not email address, currently this could throw a NPE.

PR #5270, opened 2020-04-03
Anomaly #27080 PURCHASE ORDER : viewer makes an empty untitled field …

shve-axelor:
…appear on edit mode.


wde-axelor:
Wasn't this viewer introduced on wip-6.0 ? There is no need to include a changelog entry if you are fixing an issue that was included in the same release.

PR #5299, opened 2020-04-07
RM#27246: BANK ORDER REPORT : fix the problem of empty report if bank…

jdo-axelor:
… order line sequences are too big.


ale-axelor:
I merged it manually

PR #5351, opened 2020-04-22
RM27523 : Add inventory valuation config

aol-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/InventoryLineService.java
```diff
@@ -101,10 +102,21 @@ public InventoryLine compute(InventoryLine inventoryLine, Inventory inventory) {
               : BigDecimal.ZERO;
       inventoryLine.setGap(gap);
 
-      if (stockLocationLine != null) {
-        inventoryLine.setGapValue(
-            stockLocationLine.getAvgPrice().multiply(gap).setScale(2, RoundingMode.HALF_EVEN));
+      BigDecimal price;
+      int value = stockLocation.getCompany().getStockConfig().getInventoryValuationTypeSelect();
+      if (value == StockConfigRepository.VALUATION_TYPE_ACCOUNTING_VALUE) {
```
ale-axelor, line 107: I think sale price should be chosen if we value using sale.

PR #5375, opened 2020-04-29
RM #27754 : Invoice, ventilation process unit check

mga-axelor:
- When ventilating an invoice, when checking if the stockMoveQties are right, check if the unit is not the same and convert units to compare if possile
- Add translation

RM #27754


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/workflow/WorkflowVentilationServiceSupplychainImpl.java
```diff
@@ -263,7 +266,28 @@ private void stockMoveProcess(Invoice invoice) throws AxelorException {
         } else {
           qty = qty.add(invoiceLine.getQty());
         }
-        if (stockMoveLine.getRealQty().compareTo(qty) >= 0) {
+
+        BigDecimal qtyToCompare = qty;
+
+        Unit movUnit = stockMoveLine.getUnit(), invUnit = invoiceLine.getUnit();
+        if (movUnit != invUnit) {
```
ale-axelor, line 273: this case is already managed in UnitConversionService.convert

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/exception/IExceptionMessage.java
```diff
@@ -242,6 +242,9 @@
   static final String STOCK_MOVE_INVOICE_ERROR = /*$$(*/
       "Stock move has already been totally invoiced." /*)*/;
 
+  static final String STOCK_MOVE_INVOICE_QTY_INVONVERTIBLE_UNIT = /*$$(*/
+      "The invoice's unit is different and invonvertible into the stock move's unit." /*)*/;
```
ale-axelor, line 246: Fix typo invonvertible => inconvertible

PR #5389, opened 2020-05-04
Invoice, ventilation process unit check

mga-axelor:
- When ventilating an invoice, when checking if the stockMoveQties are right, check if the unit is not the same and convert units to compare if possile
- Add translation

RM #27754 (cherry-pick from 5.3)


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/workflow/WorkflowVentilationServiceSupplychainImpl.java
```diff
@@ -263,7 +266,28 @@ private void stockMoveProcess(Invoice invoice) throws AxelorException {
...
           qty = qty.add(invoiceLine.getQty());
         }
-        if (stockMoveLine.getRealQty().compareTo(qty) >= 0) {
+
+        BigDecimal qtyToCompare = qty;
+
+        Unit movUnit = stockMoveLine.getUnit(), invUnit = invoiceLine.getUnit();
+        try {
+          qtyToCompare =
+              Beans.get(UnitConversionService.class)
```
gdu-axelor, line 275: Better to avoid to use conditional injection in a loop.

PR #5424, opened 2020-05-08
Feature RM#28009 OPPORTUNITY:sequence

sma-axelor:


axelor-crm/src/main/java/com/axelor/apps/crm/service/OpportunityServiceImpl.java
```diff
@@ -92,4 +94,19 @@ public Partner createClientFromLead(Opportunity opportunity) throws AxelorExcept
 
     return partner;
   }
+
+  @Override
+  public void onSave(Opportunity opportunity) throws AxelorException {
```
rme-axelor, line 99: Give the proper name of this method, it should be setSequence

axelor-crm/src/main/java/com/axelor/apps/crm/service/OpportunityServiceImpl.java
```diff
@@ -92,4 +94,19 @@ public Partner createClientFromLead(Opportunity opportunity) throws AxelorExcept
...
+    if (opportunity.getOpportunitySeq() == null) {
```
rme-axelor, line 100: and keep this check in repo method only [https://github.com/axelor/axelor-open-suite/pull/5424/files#diff-0a5df4699240475325794087f81bcb5bR37](url)

PR #5435, opened 2020-05-13
RM #22940 : MoveLine & BankSatement, add 2 new dashlet in form views

mga-axelor:
In move-line-form add a dashlet to display associated bank reconciliation lines
In bank-statement-form add a dashlet to display associated move lines

RM #22940


axelor-bank-payment/src/main/resources/views/BankStatement.xml
```diff
@@ -78,6 +80,18 @@
  		<context name="_id" expr="eval: id"/>
  	</action-view>
 
+	<action-view name="action-bankStatement-view-moveLines" title="Move lines" model="com.axelor.apps.account.db.MoveLine">
+		<view type="grid" name="move-line-move-grid"/>
```
ale-axelor, line 84: isn't it better to use move-line-grid and move-line-form here ? move-line-move-grid is made to display information about move line inside a move form.

PR #5438, opened 2020-05-13
#27882 added 'desired delivery date' in axelor-purchase PurchaseOrder

ThomasSynk:


ThomasSynk:
#27882 added 'desired delivery date' in axelor-purchase PurchaseOrder  + french translation

PR #5449, opened 2020-05-15
Anomaly #28200: FORECAST RECAP : many anomalies

msh-axelor:


axelor-cash-management/src/main/resources/views/ForecastRecapLine.xml
```diff
@@ -5,7 +5,10 @@
     
     <grid name="forecast-recap-line-grid" title="Lines" model="com.axelor.apps.cash.management.db.ForecastRecapLine" orderBy="estimatedDate,typeSelect,id">
        	<field name="estimatedDate"/>
+       	<field name="forecastRecapLineType.name" title="forecastRecapLineTypeName" />
```
rch-axelor, line 8: Title should be in Capital case
And title should be 'Forecast recap line type'

PR #5573, opened 2020-06-26
Wip 28254

fpr-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderWorkflowService.java
```diff
@@ -716,7 +716,14 @@ public void merge(List<Long> ids) throws AxelorException {
       }
     }
 
+    LocalDateTime minDate =
+        manufOrderList.stream()
+            .map(mo -> mo.getPlannedStartDateT())
+            .min(LocalDateTime::compareTo)
+            .get();
```
ale-axelor, line 723: wrong use of optional, this could raise an exception.
Instead use the method ifPresent to set the found value if there is one.

PR #5597, opened 2020-07-03
RM27184 added description field on lead reports

erg-axelor:


axelor-crm/src/main/resources/reports/Lead.rptdesign
```diff
@@ -70,6 +76,9 @@
             <property name="valueType">static</property>
             <property name="dataType">string</property>
             <property name="distinct">true</property>
+            <simple-property-list name="defaultValue">
+                <value type="constant">lmdpdlaa</value>
```
ale-axelor, line 80: Please remove default password

PR #5628, opened 2020-07-14
Feature #29610 TEMPLATE : BirtTemplate

shve-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/message/TemplateMessageServiceBaseImpl.java
```diff
@@ -63,11 +63,14 @@ public TemplateMessageServiceBaseImpl(
   public Set<MetaFile> getMetaFiles(Template template) throws AxelorException, IOException {
 
     Set<MetaFile> metaFiles = super.getMetaFiles(template);
-    if (template.getBirtTemplate() == null) {
+    Set<BirtTemplate> birtTemplates = template.getBirtTemplateSet();
```
psh-axelor, line 66: Check null and empty both or just use CollectionUtils.isNotEmpty

axelor-base/src/main/java/com/axelor/apps/base/service/message/TemplateMessageServiceBaseImpl.java
```diff
@@ -63,11 +63,14 @@ public TemplateMessageServiceBaseImpl(
...
+    if (birtTemplates == null) {
```
psh-axelor, line 67: Check null and empty both or just use CollectionUtils.isNotEmpty

axelor-base/src/main/resources/data-init/input-config.xml
```diff
@@ -13,7 +13,9 @@
         <bind to="canExport" eval="can_export == 'x' ? 'true' : 'false'"/>
     </input>
 
-    <input file="base_template.csv" separator=";" type="com.axelor.apps.message.db.Template"  search="self.name = :name"  />
+    <input file="base_template.csv" separator=";" type="com.axelor.apps.message.db.Template"  search="self.name = :name"  >
```
psh-axelor, line 16: If we want to set birtTemplate in Template then birtTemplate should be imported first

Test with move below file import tags before importing Template:
<input file="base_birtTemplate.csv" separator=";" type="com.axelor.apps.base.db.BirtTemplate" search="self.importId = :importId"  />
<input file="base_birtTemplateParameter.csv" separator=";" type="com.axelor.apps.base.db.BirtTemplateParameter" search="self.importId = :importId"  />

PR #5728, opened 2020-08-18
RM30610 Changed an error message for stock move already invoiced

erg-axelor:


ale-axelor:
Please apply java format

PR #5748, opened 2020-08-20
RM30649 corrected wrong fr translation for contact

erg-axelor:


ale-axelor (changes requested):
This is a fix, so please move the changelog entry to "bug fixes" category.

PR #5750, opened 2020-08-20
RM30609 added the origin reference and the concerned product to the m…

erg-axelor:
…essage error appearing when estimated delivery date is null


ale-axelor (changes requested):
- The new error message is not translated, please translate it.
- Please apply spotless.

PR #5775, opened 2020-08-25
RM30690 : Add sequence to opportunity demo data

aol-axelor:


axelor-crm/src/main/java/com/axelor/csv/script/ImportOpportunity.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class ImportOpportunity {
+  public Object importOpportunity(Object bean, Map<String, Object> values) {
+    assert bean instanceof Opportunity;
+
+    Opportunity opportunity = (Opportunity) bean;
+
+    try {
+      Beans.get(OpportunityService.class).setSequence(opportunity);
+    } catch (AxelorException e) {
```
ale-axelor, line 18: Please change it so we catch and log every exception

PR #5776, opened 2020-08-25
Anomaly #25254: INVOICE : control to avoid cancelation of ventilated …

nsw-axelor:
…invoice


axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceController.java
```diff
@@ -176,6 +176,9 @@ public void cancel(ActionRequest request, ActionResponse response) throws Axelor
     if (invoice.getStatusSelect() == InvoiceRepository.STATUS_VENTILATED
         && invoice.getCompany().getAccountConfig() != null
         && !invoice.getCompany().getAccountConfig().getAllowCancelVentilatedInvoice()) {
+      response.setAlert(
```
vbh-axelor, line 179: You can use response.setError() method

vbh-axelor (changes requested):

PR #5794, opened 2020-08-27
RM26795 : Change visibility of product and product family on account …

aol-axelor:
…management


ale-axelor (changes requested):
You cannot mix the use of `showIf` and `hidden` attribute. Please put the condition of `showIf`in the action onload and add `hidden="true"` to the two fields so they are hidden during the page is loaded.

PR #5806, opened 2020-08-28
RM30859 Fixed the anomaly causing the report from sale invoice detail…

erg-axelor:
… to be empty


ale-axelor:
already merged in dev

PR #5830, opened 2020-09-08
RM30999 Changed action to balance move lines to java

erg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/MoveLineController.java
```diff
@@ -57,6 +57,30 @@ public void computeAnalyticDistribution(ActionRequest request, ActionResponse re
     }
   }
 
+  public void balanceCreditDebit(ActionRequest request, ActionResponse response) {
```
ale-axelor, line 60: Please add a try catch to trace exceptions and refactor the the moveLine update to be made into a service.

PR #5936, opened 2020-10-07
Anomaly #31786 Hard coded locale

shve-axelor:


axelor-admin/src/main/java/com/axelor/meta/service/MetaGroupMenuAssistantService.java
```diff
@@ -72,6 +73,8 @@
 
   @Inject private MetaFiles metaFiles;
 
+  @Inject private AppFilter appFilter;
```
psh-axelor, line 76: No need to inject, use it directly

PR #5941, opened 2020-10-07
Anomaly #31511 Sale order : No allocation for services

shve-axelor:


changelogs/unreleased/no-allocation-for-services.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Sale order : make allocateQty, requestQty, cancelReservation buttons of the sale order line and stock move line to readonly for product type service."
```
psh-axelor, line 2: Update change log as per latest instructions, do not use technical terms here

PR #5962, opened 2020-10-12
RM#31948: corrected sql error in fiscal year closing method

jdo-axelor:


ale-axelor (changes requested):
please use named query parameter instead of ordinal parameters to avoid future issues.
if possible describe the "bug" in the changelog (for example, fix NullPointerException when closing fiscal year)

PR #6014, opened 2020-10-26
Anomaly RM#32262 : PURCHASE REQUEST CREATOR : delete object

ple-axelor:


axelor-purchase/src/main/resources/views/MetaJsonField.xml
```diff
@@ -2,30 +2,10 @@
...
-	    <panel-include view="json-field-form-base" />
-    </form>
-    
-    <grid name="purchase-request-json-field-grid" title="Custom fields" model="com.axelor.meta.db.MetaJsonField">
-	  <field name="name"/>
-	  <field name="type"/>
-	</grid>
-    
+        
 	<action-attrs name="action-purchase-request-creator-set-json-field-default" >
```
vbh-axelor, line 6: you can also remove this action as it is not used anymore.

vbh-axelor (changes requested):

PR #6037, opened 2020-10-29
Anomaly #31585 Added missing fields in Inventory demo import

jbo-axelor:
Anomaly #31585


ale-axelor (changes requested):
Please add CHANGELOG

PR #6057, opened 2020-11-04
Anomaly #31589 : FORECAST GENERATOR : copy

lbl-axelor:


lbl-axelor:
Error on my part, the second commit corresponds to the correction of another ticket. 

I canceled these changes with the third commit and I will create another pull request corresponding to the correction of the other ticket

PR #6061, opened 2020-11-04
Feature #31604 Add new functional origin select values for moves

jbo-axelor:
Feature #31604


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveServiceImpl.java
```diff
@@ -163,7 +163,8 @@ public Move createMove(Invoice invoice) throws AxelorException {
               partner,
               invoice.getInvoiceDate(),
               invoice.getPaymentMode(),
-              MoveRepository.TECHNICAL_ORIGIN_AUTOMATIC);
+              MoveRepository.TECHNICAL_ORIGIN_AUTOMATIC,
+              0);
```
ale-axelor, line 167: I'm unsure but shouldn't it be -1 here ? or use everywhere 0 as the default empty value since this is the framework's convention.

jbo-axelor: Yes you're right, i changed moveLine creation to set nothing if functionalOriginSelect is -1 
    if (functionalOriginSelect != -1) {
      move.setFunctionalOriginSelect(functionalOriginSelect);
    }
But it's useless is 0 is already the default empty value

PR #6171, opened 2020-12-09
RM#23258 : Fix tax number computing

aol-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/PartnerServiceImpl.java
```diff
@@ -679,23 +679,32 @@ public String getPartnerDomain(Partner partner) {
 
   @Override
   public String getTaxNbrFromRegistrationCode(Partner partner) {
-    String regCode = partner.getRegistrationCode();
     String taxNbr = "";
 
+    String countryCode = partner.getMainAddress().getAddressL7Country().getAlpha2Code();
```
ale-axelor, line 684: a NPE could be thrown if the partner has no address, or if the country is missing on the address.

PR #6262, opened 2021-01-22
RM#34452 Helpdesk> Ticket dashboards - changes

hmi-axelor:


axelor-helpdesk/src/main/resources/views/Charts.xml
```diff
@@ -11,22 +11,42 @@
 	  	<dataset type="sql">
 			SELECT 
 				COUNT(*) AS total_row, 
-				'ALL' as _status
+				(CASE WHEN(translation.message_value = '') THEN
```
jpo-axelor, line 14: For null check, you can use CASE WHEN(COALESCE(translation.message_value,'')= '')
change it every where

PR #6519, opened 2021-04-13
Feature RM#36900 : Add barcode generation to tracking number

abo-axelor:


abo-axelor:
Still not the good branch name. Need to update ticket number.

PR #6520, opened 2021-04-13
Feature RM#37079 : Add barcode generation to tracking number

abo-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/TrackingNumberConfiguratorService.java
```diff
@@ -0,0 +1,31 @@
...
+import com.axelor.apps.base.db.Sequence;
+import com.axelor.apps.base.service.BarcodeGeneratorService;
+import com.axelor.apps.base.service.administration.SequenceService;
+import com.axelor.apps.base.service.app.AppBaseService;
+import com.axelor.apps.stock.db.TrackingNumberConfiguration;
+import com.axelor.exception.AxelorException;
+import com.axelor.inject.Beans;
+import com.google.inject.Inject;
+
+public class TrackingNumberConfiguratorService {
```
ale-axelor, line 12: Please create an interface for this service and rename this class TrackingNumberConfiguratorServiceImpl

axelor-stock/src/main/java/com/axelor/apps/stock/service/TrackingNumberService.java
```diff
@@ -134,6 +137,15 @@ public TrackingNumber createTrackingNumber(
     }
     trackingNumber.setTrackingNumberSeq(seq);
 
+    // In case of barcode generation, retrieve the one set on tracking number configuration
+    AppStock appStock = Beans.get(AppStockService.class).getAppStock();
```
ale-axelor, line 141: please avoid using conditional injection for services and inject from the constructor

axelor-stock/src/main/java/com/axelor/apps/stock/web/TrackingNumberConfigurationController.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public void checkSequenceAndBarcodeTypeConfigConsistency(
+      ActionRequest request, ActionResponse response) {
+    try {
+      TrackingNumberConfiguration config =
+          request.getContext().asType(TrackingNumberConfiguration.class);
+      Beans.get(TrackingNumberConfiguratorService.class)
+          .checkSequenceAndBarcodeTypeConfigConsistency(config);
+    } catch (Exception e) {
+      response.setError(e.getLocalizedMessage());
```
ale-axelor, line 21: Please use `TraceBackService.trace(response, e, ResponseMessageType.ERROR);` to trace the full exception and to return an error with the message.

axelor-base/src/main/java/com/axelor/apps/base/service/BarcodeGeneratorServiceImpl.java
```diff
@@ -357,4 +357,72 @@ public String addCharPaddingBits(int barcodeLength, String serialno, BarcodeForm
         barcodeFormat,
         barcodeLength);
   }
+
+  @Override
+  public boolean checkSerialNumberConsistency(
+      String serialno, BarcodeTypeConfig barcodeTypeConfig, boolean isPadding)
+      throws AxelorException {
+    if (serialno != null && barcodeTypeConfig != null) {
+      BarcodeFormat barcodeFormat = null;
```
ale-axelor, line 366: I think it would be a small improvement to not initialize it, and let the compiler checks if the variable is set for every possibility

axelor-stock/src/main/java/com/axelor/apps/stock/service/TrackingNumberConfiguratorService.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public boolean checkSequenceAndBarcodeTypeConfigConsistency(TrackingNumberConfiguration config)
+      throws AxelorException {
+    if (config.getBarcodeTypeConfig() != null
+        && config.getSequence() != null
+        && config.getUseTrackingNumberSeqAsSerialNbr()) {
+      Sequence sequence = config.getSequence();
+      String testSeq =
+          sequenceService.computeTestSeq(
+              sequence, Beans.get(AppBaseService.class).getTodayDate(sequence.getCompany()));
```
ale-axelor, line 25: `AppBaseService` should also be injected from the constructor

ale-axelor (changes requested):

axelor-stock/src/main/java/com/axelor/apps/stock/web/TrackingNumberConfigurationController.java
abo-axelor: This method is used like an action-validate, we don't really need to trace the error, we just want the user to have a error message if needed. This methods relies on consistency check between sequence and barcode type process which only throw exception when something wrong happend. Catching it and restore it to the user and block following actions is the behaviour that we want here. Also adding traceback after response.setError will display in double the error message to user.

PR #6527, opened 2021-04-19
Dev 33927

rth-axelor:
change of account-year-form in year-form in the panel related of production - year


axelor-base/src/main/resources/views/Period.xml
```diff
@@ -2,114 +2,145 @@
...
+			<field name="toDate" colSpan="3"
+				onChange="action-period-year-on-change-to-date-validate"
+				required="true" />
+			<field name="allowExpenseCreation"
+				if-module="axelor-human-resource"
+				if="__config__.app.isApp('expense')" />
+		</panel>
+		<panel-dashlet
+			action="action-view-period-linked-adjusting-moves-dashlet"
+			name="linkedAdjustingMovesPanel" />
```
ale-axelor, line 45: this panel should have hidden="true"

axelor-base/src/main/resources/views/Period.xml
```diff
@@ -2,114 +2,145 @@
...
 		<domain>self.period.id = :id AND self.adjustingMove = true</domain>
 	</action-view>
 
-	<action-attrs name="action-attrs-period-hide-linked-adjusting-moves-dashlet">
-		<attribute for="linkedAdjustingMovesPanel" name="hidden" expr="eval: __repo__(AdjustHistory).all().filter('self.period.id = ?', id).fetch().size() == 0"/>
+
+
+	<action-attrs
+		name="action-attrs-period-hide-linked-adjusting-moves-dashlet">
+		<attribute for="linkedAdjustingMovesPanel" name="hidden"
```
ale-axelor, line 90: manage the case where id is null

ale-axelor (changes requested):
Please remove all `onChange=save`

PR #6599, opened 2021-05-17
Anomaly RM#37851 : ACCOUNTINGBATCH : View conditions not set properly…

ple-axelor:
… and rework the method to get tradingNameSet

### DIFF
--- a/axelor-account/src/main/resources/views/AccountingBatch.xml
+++ b/axelor-account/src/main/resources/views/AccountingBatch.xml
@@ -13,28 +13,30 @@
     <field name="createdBy" form-view="user-form" grid-view="user-grid"/>
   </grid>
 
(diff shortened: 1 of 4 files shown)


axelor-account/src/main/resources/views/AccountingBatch.xml
```diff
@@ -114,6 +113,7 @@
 	<action-group name="action-accounting-batch-group-on-load">
 		<action name="action-accounting-batch-attrs-trading-names-management"/>
 		<action name="action-accounting-batch-attrs-display-trading-name-set"/>
+		<action name="save"/>
```
vbh-axelor, line 116: Remove this 'save' action from onLoad

vbh-axelor (changes requested):

PR #6617, opened 2021-05-26
Anomaly RM#37984 : Project and task major changes reverted

hta-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/db/repo/ProjectBusinessProjectRepository.java
```diff
@@ -0,0 +1,65 @@
...
+              company,
+              TraceBackRepository.CATEGORY_CONFIGURATION_ERROR,
+              I18n.get(IExceptionMessage.PROJECT_SEQUENCE_ERROR),
+              company.getName());
+        }
+
+        project.setCode(seq);
+      }
+    } catch (AxelorException e) {
+      throw new PersistenceException(e.getLocalizedMessage());
```
ale-axelor, line 61: Please change this line to
```suggestion
      throw new PersistenceException(e);
```
to avoid losing the stacktrace when an exception occurs. On top of that we can also trace the AxelorException before rethrowing it so we are sure to trace it.

PR #6648, opened 2021-06-08
Feature #37079 : Tracking number barcode - Fix barcode type config defaulting

abo-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/db/repo/TrackingNumberManagementRepository.java
```diff
@@ -82,6 +82,10 @@ public TrackingNumber save(TrackingNumber trackingNumber) {
       BarcodeTypeConfig barcodeTypeConfig = trackingNumber.getBarcodeTypeConfig();
       if (!appStock.getEditTrackingNumberBarcodeType()) {
         barcodeTypeConfig = appStock.getTrackingNumberBarcodeTypeConfig();
+      } else if (trackingNumber.getProduct() != null
+          && trackingNumber.getProduct().getTrackingNumberConfiguration() != null) {
+        barcodeTypeConfig =
+            trackingNumber.getProduct().getTrackingNumberConfiguration().getBarcodeTypeConfig();
```
ale-axelor, line 88: I don't get why we use the configuration from the product instead of the tracking number. Here whether EditTrackingNumberBarcodeType is true or not we will not use the barcodeTypeConfig stored in tracking number ?
Even if this is the wanted behavior, please rewrite it to be less confusing

abo-axelor: I have changed the code accordingly to improve readability.

PR #6704, opened 2021-06-28
  Feature #39243 Add respite date on debt recoveries to prevent customer account from being blocked before the given date

jbo-axelor:
Feature #39243


axelor-account/src/main/java/com/axelor/apps/account/service/batch/BatchBlockCustomersWithLatePayments.java
```diff
@@ -50,47 +52,60 @@ protected void process() {
...
-                .append(I18n.get("Blocking"))
-                .append(" ")
-                .append(partner.getFullName())
-                .append("</br>");
-            customersToBlock.add(partner.getId());
-            incrementDone();
+        if (debtRecovery
+                .getRespiteDateBeforeAccountBlocking()
+                .compareTo(appBaseService.getTodayDate(debtRecovery.getCompany()))
+            >= 1) {
```
ale-axelor, line 67: isn't it
```suggestion
            >= 0) {
```
or do we block user when we are at the respire date ? if so please change it to
```suggestion
            > 0) {
```

axelor-account/src/main/java/com/axelor/apps/account/service/batch/BatchBlockCustomersWithLatePayments.java
```diff
@@ -50,47 +52,60 @@ protected void process() {
...
-            log.debug("Blocking {}", partner.getFullName());
-            result
-                .append(I18n.get("Blocking"))
-                .append(" ")
-                .append(partner.getFullName())
-                .append("</br>");
-            customersToBlock.add(partner.getId());
-            incrementDone();
+        if (debtRecovery
+                .getRespiteDateBeforeAccountBlocking()
```
ale-axelor, line 65: This field is not required, so this line can throw a NPE

ale-axelor (changes requested):

PR #6706, opened 2021-06-28
Feature #39729 : Configurator manage products to consume

abo-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/configurator/ConfiguratorProdProcessServiceImpl.java
```diff
@@ -126,14 +146,16 @@ public ProdProcess generateProdProcessService(
...
+            isConsProOnOperation,
             product);
 
     List<ConfiguratorProdProcessLine> confLines =
         confProdProcess.getConfiguratorProdProcessLineList();
     if (confLines != null) {
       for (ConfiguratorProdProcessLine confLine : confLines) {
         ProdProcessLine generatedProdProcessLine =
-            confProdProcessLineService.generateProdProcessLine(confLine, attributes);
+            confProdProcessLineService.generateProdProcessLine(
```
ale-axelor, line 157: isConsProOnOperation is a Boolean (nullable), and the method `generateProdProcessLine` is using a primitive type boolean, so I think a NPE can be thrown here.

ale-axelor (changes requested):

PR #6742, opened 2021-07-13
39771 partner rework

rth-axelor:


fri-axelor:
Please,

* fix the mainActivity domain
* for the listener, the AOS convention is to put in repo package but the listener is a class on its own
* in messages_fr.csv, set the translation of MainActivity to "Code NAF"

PR #6783, opened 2021-08-02
RM#40232 Partner relations types : translatable and datainit

hmi-axelor:


changelogs/unreleased/partner-relations-types-set-translatable-true-for-link1Name-and-link2Name-and-add-demo-data.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Partner relations types : set translatable true for link1Name and link2Name and add demo data."
```
jpo-axelor, line 2: its data init (not demo data)

PR #6827, opened 2021-09-09
Anomaly #41310: Translation, currency and hide button

kra-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateService.java
```diff
@@ -240,6 +240,10 @@ public void validate(Move move, boolean updateCustomerAccount) throws AxelorExce
     if (move.getPeriod().getStatusSelect() == PeriodRepository.STATUS_ADJUSTING) {
       move.setAdjustingMove(true);
     }
+    // Setting currency code in case it not valued
+    if (move.getCompanyCurrency() != null && move.getCurrencyCode() == null) {
+      move.setCurrencyCode(move.getCompanyCurrency().getCode());
+    }
```
ale-axelor, line 246: This should be called inside `completeMoveLines` method

ale-axelor: Indeed, instead this code should be removed since on validation we will call `moveRepository.save(move);` which will call the exact same code

PR #6836, opened 2021-09-10
41364 - NPE management on moveline (onchange account with no partner)

sha-axelor:


sha-axelor:
Duplicate PR with 6898

PR #6843, opened 2021-09-14
Anomaly  RM#41151 - Base: Replacement of static country to dynamic

hde-axelor:


axelor-base/src/main/resources/views/Address.xml
```diff
@@ -93,7 +93,7 @@
 	</action-group>
 
 	<action-record name="default-address-record" model="com.axelor.apps.base.db.Address">
-		<field name="addressL7Country" expr="action:default-address-country-record" />
+		<field name="addressL7Country" expr="eval: __user__.activeCompany.address.addressL7Country"/>
```
gdu-axelor, line 96: Need to manage NPE if user doesn't have an active company, or an address

gdu-axelor (changes requested):

PR #6873, opened 2021-09-22
Remove of a default account

rth-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/AnalyticMoveLineServiceImpl.java
```diff
@@ -211,8 +211,6 @@ public AnalyticMoveLine computeAnalyticMoveLine(
       if (moveLine.getAccount().getAccountType() != null) {
```
gdu-axelor, line 211: No need to check not null value here.

gdu-axelor (changes requested):

PR #6891, opened 2021-09-28
40775 Add of an import function for analyticGrouping

rth-axelor:


axelor-account/src/main/java/com/axelor/csv/script/ImportAnalyticGrouping.java
```diff
@@ -0,0 +1,19 @@
...
+
+public class ImportAnalyticGrouping {
+
+  public Object importAnalyticGrouping(Object bean, Map<String, Object> values)
+      throws AxelorException {
+    assert bean instanceof AnalyticGrouping;
+    AnalyticGrouping analyticGrouping = (AnalyticGrouping) bean;
+    if (ObjectUtils.isEmpty(analyticGrouping.getName())) {
+      analyticGrouping.setFullName(analyticGrouping.getCode() + "_" + analyticGrouping.getName());
+    }
```
gdu-axelor, line 16: We should have only one way to compute the fullname. Means, we can define a java method an call it from xml view when we update an analyticGrouping and also call it from the xml config file to import. Else, it better to use xml action in xml view and in a bind in import config file.

gdu-axelor (changes requested):

PR #6905, opened 2021-09-30
42006 used generic moveLine creation for counterpartMoveLine

sha-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveCounterPartServiceImpl.java
```diff
@@ -16,45 +17,54 @@
...
+      MoveLineCreateService moveLineCreateService) {
     this.moveRepository = moveRepository;
     this.moveLineToolService = moveLineToolService;
+    this.moveLineCreateService = moveLineCreateService;
   }
 
   @Override
   @Transactional
-  public void generateCounterpartMoveLine(Move move) {
+  public void generateCounterpartMoveLine(Move move) throws Exception {
```
gdu-axelor, line 34: Need to manage rollback on the @Transactional annotation and throw the specific exceptions

gdu-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveCounterPartServiceImpl.java
gdu-axelor: Need to use specific exception instead of Exception

gdu-axelor (changes requested):

PR #6908, opened 2021-10-01
42011 - use standard functions to get accountingAccount from paymentMode

sha-axelor:
…

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveCounterPartServiceImpl.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveCounterPartServiceImpl.java
@@ -1,13 +1,14 @@
 package com.axelor.apps.account.service.move;
 
 import com.axelor.apps.account.db.Account;
-import com.axelor.apps.account.db.AccountManagement;
 import com.axelor.apps.account.db.AccountingSituation;
 import com.axelor.apps.account.db.Move;
 import com.axelor.apps.account.db.MoveLine;
 import com.axelor.apps.account.db.repo.JournalTypeRepository;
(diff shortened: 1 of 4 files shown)


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveCounterPartServiceImpl.java
```diff
@@ -16,23 +17,27 @@
...
+      PaymentModeService paymentModeService) {
     this.moveRepository = moveRepository;
     this.moveLineToolService = moveLineToolService;
+    this.paymentModeService = paymentModeService;
   }
 
   @Override
   @Transactional
-  public void generateCounterpartMoveLine(Move move) {
+  public void generateCounterpartMoveLine(Move move) throws AxelorException {
```
gdu-axelor, line 34: Need to manage the rollback on Transactional annotation

axelor-account/src/main/java/com/axelor/apps/account/service/payment/PaymentModeServiceImpl.java
```diff
@@ -78,6 +78,17 @@ public Account getPaymentModeAccount(
         paymentMode.getName());
   }
 
+  @Override
+  public Account getPaymentModeAccount(PaymentMode paymentMode, Company company)
+      throws AxelorException {
```
gdu-axelor, line 83: I'm not sure about this way. I think we have to get the account for the correct paymentMode, company AND the bankDetails because we can have two accountManagement lines with same company but different bankDetails and different account.

gdu-axelor (changes requested):

PR #6926, opened 2021-10-05
42006 fixed createMoveLineForAutoTax NPE on partner

sha-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -547,35 +547,28 @@ public MoveLine createMoveLineForAutoTax(
...
-      Account account,
-      Move move)
+  protected MoveLine createMoveLine(LocalDate date, TaxLine taxLine, Account account, Move move)
       throws AxelorException {
     MoveLine moveLine;
-    debit = debit.multiply(taxLine.getValue());
-    credit = credit.multiply(taxLine.getValue());
+    BigDecimal debit = BigDecimal.ZERO;
+    BigDecimal credit = BigDecimal.ZERO;
     boolean isDebit = debit.signum() > 0;
```
gdu-axelor, line 572: Why evaluate debit.signum if you fill the debit to 0 just before.

Also, I think it's better to create the moveline with the correct debit and credit computed before in the parent method

PR #6930, opened 2021-10-05
RM#42124 add minimun on totalTax and totalAmount

alu-axelor:


axelor-human-resource/src/main/resources/domains/ExpenseLine.xml
```diff
@@ -12,8 +12,8 @@
 	    <many-to-one name="expenseProduct" ref="com.axelor.apps.base.db.Product" title="Expense Type" required="true" />
 	    <date name="expenseDate" title="Expense date" required="true"/>
 	    <decimal name="untaxedAmount" title="Untaxed amount" readonly="true"/>
-	    <decimal name="totalTax" title="Total tax" />
-	    <decimal name="totalAmount" title="Total amount with tax" required="true"/>
+	    <decimal name="totalTax" title="Total tax" min="0 "/>
```
ale-axelor, line 15: Please remove space after `0`

ale-axelor (changes requested):
Please do not use action-validate and use validIf condition instead to avoid issues on editable grid. (be careful, to use number fields inside javascript condition, call `$number`: https://docs.axelor.com/adk/5.4/dev-guide/web-client/widgets.html#expressions

PR #7002, opened 2021-10-15
Features #42389: Unicity and readOnly

kra-axelor:


axelor-account/src/main/resources/views/Period.xml
```diff
@@ -25,4 +42,16 @@
...
+    <call class="com.axelor.apps.account.web.PeriodController" method="controlDates"/>
+  </action-method>
+
+  <action-attrs name="action-attrs-account-readonly-if-period-in-move">
+    <attribute name="readonly" for="fromDate"
+      expr="eval: (__repo__(Move).all().filter('self.period = ?', __this__).count() > 0) &amp;&amp; statusSelect == 1 &amp;&amp; year?.statusSelect == 1"/>
+    <attribute name="readonly" for="toDate"
+      expr="eval: (__repo__(Move).all().filter('self.period = ?', __this__).count() > 0) &amp;&amp; statusSelect == 1 &amp;&amp; year?.statusSelect == 1"/>
+    <attribute name="readonly" for="year"
+      expr="eval: (__repo__(Move).all().filter('self.period = ?', __this__).count() > 0) &amp;&amp; statusSelect == 1 &amp;&amp; year?.statusSelect == 1"/>
```
ale-axelor, line 55: I think this way, the query will be run three times. Should be moved to a java method to only run the query once and use the result for each field.

axelor-account/src/main/resources/views/Year.xml
```diff
@@ -94,4 +95,17 @@
...
+  </action-method>
+  <action-attrs name="action-attrs-account-fiscal-year-readonly-if-in-move">
+    <attribute name="readonly" for="fromDate"
+      expr="eval: (__repo__(Move).all().filter('self.period.id in (?)', __this__?.periodList?.collect{it.id}).count() > 0) &amp;&amp; statusSelect == 1 "/>
+    <attribute name="readonly" for="toDate"
+      expr="eval: (__repo__(Move).all().filter('self.period.id in (?)', __this__?.periodList?.collect{it.id}).count() > 0) &amp;&amp; statusSelect == 1 "/>
+    <attribute name="readonly" for="periodDurationSelect"
+      expr="eval: (__repo__(Move).all().filter('self.period.id in (?)', __this__?.periodList?.collect{it.id}).count() > 0) &amp;&amp; statusSelect == 1 "/>
+    <attribute name="readonly" for="generatePeriodsBtn"
+      expr="eval: (__repo__(Move).all().filter('self.period.id in (?)', __this__?.periodList?.collect{it.id}).count() > 0) &amp;&amp; statusSelect == 1 "/>
```
ale-axelor, line 109: I think this way, the query will be run three times. Should be moved to a java method to only run the query once and use the result for each field.

axelor-base/src/main/resources/views/PaymentMode.xml
```diff
@@ -72,6 +73,17 @@
...
 
+  <action-attrs name="action-attrs-base-set-readonly-fields-if-in-move">
+    <attribute name="readonly" for="name"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="code"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="typeSelect"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="inOutSelect"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
```
ale-axelor, line 84: Same as above

ale-axelor (changes requested):
Also please add a changelog for any change (if it is not too difficult, also please create a separate PR for each ticket)

PR #7015, opened 2021-10-18
42404 - fix NPE when we click on delete move without any move selected

alg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/MoveController.java
```diff
@@ -246,7 +246,7 @@ protected void removeOneMove(Move move, ActionResponse response) throws Exceptio
   public void deleteMultipleMoves(ActionRequest request, ActionResponse response) {
     try {
       List<Long> moveIds = (List<Long>) request.getContext().get("_ids");
-      if (!moveIds.isEmpty()) {
+      if (moveIds != null && !moveIds.isEmpty()) {
```
ssb-axelor, line 249: Better to use CollectionUtils.isEmpty of org.apache.commons.collections , it has a null-safe check: 

`if (CollectionUtils.isEmpty(moveIds) {`

PR #7104, opened 2021-11-08
Feature #42433: Payment mode Code to be unique + Fields to set readOn…

kra-axelor:
…ly if any Moves linked.


axelor-base/src/main/resources/views/PaymentMode.xml
```diff
@@ -92,4 +92,17 @@
...
+  <action-attrs name="action-attrs-base-set-readonly-fields-if-in-move"
+    if-module="axelor-account">
+    <attribute name="readonly" for="name"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="code"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="typeSelect"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
+    <attribute name="readonly" for="inOutSelect"
+      expr="eval: __repo__(Move).all().filter('self.paymentMode = ?', __this__).count() > 0"/>
```
ale-axelor, line 104: please factor this query in a java method instead

axelor-account/src/main/java/com/axelor/apps/account/web/PaymentModeController.java
```diff
@@ -0,0 +1,30 @@
...
+              .find(request.getContext().asType(PaymentMode.class).getId());
+      if (paymentMode != null) {
+        Boolean isInMove = Beans.get(PaymentModeControlService.class).isInMove(paymentMode);
+        response.setAttr("name", "readonly", isInMove);
+        response.setAttr("code", "readonly", isInMove);
+        response.setAttr("typeSelect", "readonly", isInMove);
+        response.setAttr("inOutSelect", "readonly", isInMove);
+      }
+    } catch (Exception e) {
+      TraceBackService.trace(e);
```
ale-axelor, line 27: please add the response in the traceback service to prevent the method from failing silently.

axelor-account/src/main/java/com/axelor/apps/account/service/PaymentModeControlServiceImpl.java
```diff
@@ -0,0 +1,24 @@
...
+  protected MoveRepository moveRepository;
+
+  @Inject
+  public PaymentModeControlServiceImpl(MoveRepository moveRepository) {
+
+    this.moveRepository = moveRepository;
+  }
+
+  @Override
+  public Boolean isInMove(PaymentMode paymentMode) {
```
ale-axelor, line 19: We should return an unboxed `boolean` here, so we can be sure that it cannot be null.

ale-axelor (changes requested):

PR #7109, opened 2021-11-09
Feature #42431: Code to be unique + fields to set readOnly if any analyticMoveLine linked.

kra-axelor:


ale-axelor (changes requested):
Please run i18n task on messages*.csv to fix EOL chars

PR #7161, opened 2021-11-19
43483 merge

rth-axelor:


ale-axelor (changes requested):
Many changes in views seem unrelated to this change and should not be included

PR #7178, opened 2021-11-23
Feature RM#33809 STRINGTOOL: used StringBuilder to concatenate IDs in getIdListString method

mr-axelor:


axelor-tool/src/main/java/com/axelor/apps/tool/StringTool.java
```diff
@@ -268,19 +269,21 @@ public static String capitalizeFirstLetter(String s) {
...
    */
   public static String getIdListString(Collection<? extends Model> collection) {
-    List<Long> idList = new ArrayList<>();
-    String idString;
+    StringBuilder idStringBuilder = new StringBuilder();
     if (CollectionUtils.isEmpty(collection)) {
-      idString = "0";
+      idStringBuilder.append("0");
     } else {
       for (Model item : collection) {
```
dtr-axelor, line 276:
```suggestion
      return "0";
    }
    
    for (Model item : collection) {
       ...
```

We can reduce cognitive complexity and return immediatly the value. As a consequence, we don't need an `else` block below.

dtr-axelor:
Open in favor of https://github.com/axelor/axelor-open-suite/pull/8101 after deletion of branch `wip-6.3`

PR #7191, opened 2021-11-25
RM#43681,43682,44343 : Add cut off modelization

aol-axelor:


ale-axelor (changes requested):
Please change the methods `applyCutOffDates`, I think the fact that we return a boolean if the dates were not applied is not clear enough. Please use a separate method for checking in the controller and for updating the invoice/move lines.

PR #7197, opened 2021-11-25
43774 - Stock detail by product : allow to see history for all locations

alg-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockHistoryServiceImpl.java
```diff
@@ -52,6 +54,17 @@ public StockHistoryServiceImpl(
       Long productId, Long companyId, Long stockLocationId, LocalDate beginDate, LocalDate endDate)
       throws AxelorException {
     List<StockHistoryLine> stockHistoryLineList = new ArrayList<>();
+    List<Long> stockLocationIdList = new ArrayList<>();
+    if (stockLocationId == null) {
+      stockLocationIdList.addAll(
+          Beans.get(StockLocationRepository.class).all()
+              .filter("self.typeSelect != 3 AND self.company.id = :company")
```
gdu-axelor, line 61: Better to use a parameter with repository static value instead of the hard coded value "3". Like that we can easily update the code if we override the selection values.

gdu-axelor (changes requested):

PR #7249, opened 2021-12-06
RM#44056 : INVOICE / INVOICETERM / PASSFORPAYMENT : Adapt PFP Features to fit with the implementation of PFP on invoiceTerm (MODEL)

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceTermServiceImpl.java
```diff
@@ -405,4 +408,18 @@ public boolean checkInvoiceTermDeletionConditions(Invoice invoice) {
     }
     return false;
   }
+
+  @Override
+  @Transactional(rollbackOn = {AxelorException.class, Exception.class})
```
ale-axelor, line 413: This method does not throw AxelorException, and `Exception.class` already includes every exception. Here we can keep only Exception or only write `@Transactional` since this method does not throw checked exception

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -99,4 +100,16 @@ public void initInvoiceTermFromInvoice(ActionRequest request, ActionResponse res
...
+  public void refusalToPay(ActionRequest request, ActionResponse response) {
+    InvoiceTerm invoiceTerm = request.getContext().asType(InvoiceTerm.class);
+    if (invoiceTerm.getInvoice() != null && invoiceTerm.getInvoice().getCompany() != null) {
+      Beans.get(InvoiceTermService.class)
+          .refusalToPay(
+              Beans.get(InvoiceTermRepository.class).find(invoiceTerm.getId()),
+              invoiceTerm.getReasonOfRefusalToPay(),
+              invoiceTerm.getReasonOfRefusalToPayStr());
+      response.setCanClose(true);
+    }
```
ale-axelor, line 113: exception should be catched and traced inside this controller

axelor-account/src/main/resources/views/InvoiceTerm.xml
```diff
@@ -139,4 +230,26 @@
...
+    <view-param name="popup-save" value="false"/>
+    <view-param name="forceEdit" value="true"/>
+    <context name="_showRecord" expr="eval: __this__.id"/>
+    <context name="_xApplicationType" expr="eval: _model"/>
+  </action-view>
+
+  <action-condition name="action-invoiceterm-refusal-to-pay-cancel-reason-check">
+    <check error="A cancel reason must be selected" field="cancelReason"
+      if="reasonOfRefusalToPay == null || reasonOfRefusalToPay == 0"/>
+  </action-condition>
```
ale-axelor, line 253: Please make the field required instead and add a check in java method on the field to avoid a NPE in case of UI error.

ale-axelor (changes requested):

PR #7251, opened 2021-12-06
43465 add of bap on invoice term

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceTermServiceImpl.java
```diff
@@ -414,4 +417,18 @@ public boolean checkInvoiceTermDeletionConditions(Invoice invoice) {
     }
     return false;
   }
+
+  @Override
+  @Transactional(rollbackOn = {AxelorException.class, Exception.class})
+  public void refusalToPay(
```
ale-axelor, line 423: This method does not throw AxelorException, and Exception.class already includes every exception. Here we can keep only Exception or only write @Transactional since this method does not throw checked exception.

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -99,4 +101,38 @@ public void initInvoiceTermFromInvoice(ActionRequest request, ActionResponse res
...
+      return;
+    }
+
+    response.setAttr(
+        "$isSelectedPfpValidatorEqualsPartnerPfpValidator",
+        "value",
+        invoiceTerm
+            .getPfpValidatorUser()
+            .equals(Beans.get(InvoiceService.class).getPfpValidatorUser(invoiceTerm.getInvoice())));
+  }
```
ale-axelor, line 137: Exception should be catched and traced inside these controllers

axelor-account/src/main/resources/views/InvoiceTerm.xml
```diff
@@ -154,4 +281,36 @@
...
+    <view-param name="popup-save" value="false"/>
+    <view-param name="forceEdit" value="true"/>
+    <context name="_showRecord" expr="eval: __this__.id"/>
+    <context name="_xApplicationType" expr="eval: _model"/>
+  </action-view>
+
+  <action-condition name="action-invoiceterm-refusal-to-pay-cancel-reason-check">
+    <check error="A cancel reason must be selected" field="cancelReason"
+      if="reasonOfRefusalToPay == null || reasonOfRefusalToPay == 0"/>
+  </action-condition>
```
ale-axelor, line 314: Please make the field required instead and add a check in java method on the field to avoid a NPE in case of UI error.

ale-axelor (changes requested):

PR #7286, opened 2021-12-10
RM#44222: exclude wasteStockMove in update produced StockMove process

jmi-axelor:


gdu-axelor:
Change log file is missing

PR #7372, opened 2021-12-28
44807 dueDate now updated when invoice term is paid

rth-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceToolService.java
```diff
@@ -60,6 +61,14 @@ public static LocalDate getDueDate(Invoice invoice) throws AxelorException {
...
   }
 
+  @Transactional
+  public void updateDueDate(Invoice invoice) throws AxelorException {
+    if (invoice != null) {
+      invoice.setDueDate(getDueDate(invoice));
+      Beans.get(InvoiceRepository.class).save(invoice);
+    }
+  }
+
```
ssb-axelor, line 71: The service invoiceToolService only contains methods that do some checks on the invoice, so no update.
Besides, when we update the invoice term payment in method updateInvoiceTermsPaidAmount , the save is not done yet at this level, and is handled by higher calling methods. 
So maybe there is no need to add this method. we could just add something like this in methods updateInvoiceTermsPaidAmount and updateInvoiceTermsAmountRemaining : 
```
Invoice invoice = invoiceTerm.getInvoice();
if (invoice != null) {
    invoice.setDueDate(invoiceToolService.getDueDate(invoice));
}
```

ssb-axelor (changes requested):

PR #7375, opened 2021-12-29
Anomaly #44816 INVENTORY : thow an exception for product not found wh…

dde-axelor:
…ile importing


axelor-stock/src/main/java/com/axelor/apps/stock/service/InventoryService.java
```diff
@@ -257,6 +257,14 @@ public Path importFile(Inventory inventory) throws AxelorException {
                 I18n.get(IExceptionMessage.INVENTORY_12) + " " + code);
           }
         }
+
```
gdu-axelor, line 260: Can use Else condition of the previous statement

gdu-axelor (changes requested):

PR #7398, opened 2022-01-06
45019 : Payment session : Add a search button to retrieve eligible in…

alg-axelor:
…voice terms and associate them to the current payment session


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceTermServiceImpl.java
```diff
@@ -442,4 +444,116 @@ public void refusalToPay(
...
+                    + " AND (self.moveLine != NULL AND (self.moveLine.move.partner.isCustomer = TRUE OR self.moveLine.move.partner.isSupplier = TRUE OR self.moveLine.move.partner.isEmployee = TRUE) OR self.moveLine = NULL AND (self.invoice.partner.isCustomer = TRUE OR self.invoice.partner.isSupplier = TRUE OR self.invoice.partner.isEmployee = TRUE))"
+                    + " AND ((self.moveLine != NULL AND self.moveLine.account.isRetrievedOnPaymentSession = TRUE) OR (self.moveLine = NULL AND self.invoice.partnerAccount.isRetrievedOnPaymentSession = TRUE))"
+                    + " AND (self.invoice.company.accountConfig.isManagePassedForPayment = FALSE OR ((self.invoice.company.accountConfig.isManagePassedForPayment = TRUE OR self.moveLine.move.company.accountConfig.isManagePassedForPayment = TRUE) AND (self.pfpValidateStatusSelect = 2 OR self.pfpValidateStatusSelect = 4)))"
+                    + " AND self.isPaid = FALSE AND self.amountRemaining > 0 AND self.paymentSession = NULL")
```
ale-axelor, line 459: `IS NULL`/`IS NOT NULL` must be used instead of `= NULL`/`!= NULL` here.
The full query should also be rewritten to be more understandable:
- At least, more line break must be used (maximum line length in our java code is 100 chars and we have a line with 300 chars)
- Then, maybe break down the query in multiple constants. (for example, one constant per line in the current query would allow to read more easily each part)

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceTermServiceImpl.java
```diff
@@ -442,4 +444,116 @@ public void refusalToPay(
...
+                paymentSession
+                    .getPaymentDate()
+                    .minusDays(paymentSession.getPaymentMode().getDaysMarginOnPaySession()))
+            .bind("currency", paymentSession.getCurrency())
+            .fetch();
+    eligibleInvoiceTermList.forEach(
+        invoiceTerm -> {
+          fillEligibleTerm(paymentSession, invoiceTerm);
+          invoiceTermRepo.save(invoiceTerm);
+          Beans.get(PaymentSessionRepository.class).save(paymentSession);
```
ale-axelor, line 478: This save should be unnecessary, as the invoiceTerm repo will save the paymentSession in the invoiceTerm. Please remove this line. If for any reason, the PaymentSessionRepository must be called, then use the injection in the constructor to avoid injecting a service in a loop.

ale-axelor (changes requested):

PR #7425, opened 2022-01-12
44325 fix of the move setter in moveline

rth-axelor:


jmi-axelor (approved):

jmi-axelor (changes requested):
There is a conflict, it looks like the line you changed has been deleted in wip-accounting

PR #7450, opened 2022-01-19
RM#45335 : Add email generation on payment session

aol-axelor:


axelor-account/src/main/resources/views/PaymentSession.xml
```diff
@@ -78,7 +78,8 @@
     </panel>
     <panel name="emailPanel" sidebar="true" showIf="id &amp;&amp; sessionTotalAmount > 0">
       <button name="sendEmailsBtn" title="Send emails payment notice"
-        showIf="statusSelect == 1" readonlyIf="hasEmailsSent == true" onClick=""/>
+        showIf="statusSelect == 1" readonlyIf="hasEmailsSent == true"
+        onClick="save,action-payment-session-method-send-emails,save"/>
```
ale-axelor, line 82: If I'm not mistaken, the last `save` should be removed as the controller method always calls `response.setReload(true);`

ale-axelor (changes requested):

PR #7464, opened 2022-01-24
Anomaly #45312 : ADDRESS : zip code not filled in demo data

bsu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -314,4 +320,15 @@ public void autocompleteAddress(Address address) {
       }
     }
   }
+
+  @Override
+  public String getZipCode(Address address) {
+    if (address.getAddressL6() == null) return null;
```
ale-axelor, line 326: Please add missing braces to follow our [code convention](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used)

PR #7518, opened 2022-02-03
45986 -PAYMENTSESSION : Clicking on button search after on new generation an error message (Save is missing)

alg-axelor:


axelor-account/src/main/resources/views/PaymentSession.xml
```diff
@@ -211,6 +211,11 @@
       method="searchEligibleTerms"/>
   </action-method>
 
+  <action-group name="action-group-payment-session-search-eligible-terms">
+    <action name="save" if="!id"/>
```
ale-axelor, line 215: Please remove if="!id", the form must always be saved to fetch the updated payment session inside the controller

PR #7530, opened 2022-02-08
Anomaly #45312 : ADDRESS : zip code not filled in demo data

bsu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/AddressService.java
```diff
@@ -101,4 +101,6 @@ public Address getAddress(
    * @param address
    */
   public void autocompleteAddress(Address address);
+
+  String getZipCode(Address address);
```
sos-axelor, line 105: Change access modifier to public

sos-axelor (changes requested):

PR #7541, opened 2022-02-09
46070 MOVELINE / ACCOUNT / ACCOUNTINGSITUATION : VAT System (Regime) …

rth-axelor:
…modelization and computation


axelor-account/src/main/java/com/axelor/apps/account/util/TaxAccountToolServiceImpl.java
```diff
@@ -0,0 +1,124 @@
...
+
+  protected AccountingSituationRepository accountingSituationRepository;
+
+  @Inject
+  public TaxAccountToolServiceImpl(AccountingSituationRepository accountingSituationRepository) {
+    this.accountingSituationRepository = accountingSituationRepository;
+  }
+
+  @Override
+  public Integer calculateVatSystem(
```
ale-axelor, line 29: better to return `int` so it is clearer that this method cannot return null

ale-axelor (changes requested):
Please check combination of showIf and requiredIf on fields, minding the fact that an integer select will have a default value

PR #7600, opened 2022-02-22
46227

rth-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/generator/tax/TaxInvoiceLine.java
```diff
@@ -28,28 +31,37 @@
...
    *
    * @return La liste des lignes de TVA de la facture.
+   * @throws AxelorException
    */
   @Override
-  public List<InvoiceLineTax> creates() {
+  public List<InvoiceLineTax> creates() throws AxelorException {
 
-    Map<TaxLine, InvoiceLineTax> map = new HashMap<>();
+    Map<List<Object>, InvoiceLineTax> map = new HashMap<>();
```
ale-axelor, line 64: Please do not use a list of object, instead use a collection that states explicitly what will be used as keys. You can create a nested class that contains a taxLine, or use a map of a map, like `Map<TaxLine, Map<Integer, InvoiceLineTax>>`

ale-axelor (changes requested):

PR #7603, opened 2022-02-23
Anomaly RM#43336 : Wrong translation in scheduler

ple-axelor:


changelogs/unreleased/fix-wrong-translation-in-schedular.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "MetaSchedule: Add title of field."
```
dtr-axelor, line 2:
```suggestion
title: "MetaSchedule: Fix the title for 'Service' field"
```
As quoted in https://github.com/axelor/axelor-open-suite/blob/v6.0.18/changelogs/README.md#changelog-is-about-application-usage:
> Changelog is about application usage

PR #7640, opened 2022-03-02
Anomaly #46673 App Base: default project unit

alo-axelor:


alo-axelor:
https://git.axelor.com/abs/abs-scripts/-/commit/c2bbf3f28caf22051a6df1c2ccd33931e5e194f0

Here is the script removing the unused column.

PR #7645, opened 2022-03-03
RM#46651 MAIL TEMPLATE ASSOCIATION: add to template context new keys …

hmi-axelor:
…and change of behavior


ale-axelor:
this change will be merged with https://github.com/axelor/axelor-open-suite/pull/8490

PR #7649, opened 2022-03-03
Anomaly #46928: call validate method when auto validation from invoice

kra-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/FixedAssetValidateServiceImpl.java
```diff
@@ -0,0 +1,111 @@
...
+          && !fixedAsset.getFixedAssetDerogatoryLineList().isEmpty()) {
+        fixedAssetDerogatoryLineService.clear(fixedAsset.getFixedAssetDerogatoryLineList());
+      }
+      if (fixedAsset.getIfrsFixedAssetLineList() != null
+          && !fixedAsset.getIfrsFixedAssetLineList().isEmpty()) {
+        fixedAssetLineService.clear(fixedAsset.getIfrsFixedAssetLineList());
+      }
+
+      if (!fixedAsset
+          .getDepreciationPlanSelect()
```
ale-axelor, line 70: can this field be null instead of empty ? I'm not sure because it is not required but in AOP a string can be nullable (the generated getter does not return an empty string. So I think it is better to check if the field is null before.

axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/FixedAssetValidateServiceImpl.java
```diff
@@ -0,0 +1,111 @@
...
+          .contains(FixedAssetRepository.DEPRECIATION_PLAN_NONE)) {
+        fixedAsset = fixedAssetGenerationService.generateAndComputeLines(fixedAsset);
+      } else {
+        fixedAsset.setNumberOfDepreciation(fixedAsset.getNumberOfDepreciation() - 1);
+      }
+
+      if (fixedAsset.getIsEqualToFiscalDepreciation()) {
+        fixedAsset.setAccountingValue(fixedAsset.getGrossValue());
+      } else if (fixedAsset
+          .getDepreciationPlanSelect()
```
ale-axelor, line 80: same issue as above

ale-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/FixedAssetValidateServiceImpl.java
kra-axelor: Should not be null today, but managed null value as it could be null in the futur

PR #7717, opened 2022-03-14
RM#47372: fix closing of a fiscal period with previous one still opened

ija-axelor:


jmi-axelor (changes requested):
please don't commit the translation files from axelor-account

PR #7722, opened 2022-03-15
RM#47321: fixing duplicates settings in journal copy

ija-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/JournalController.java
```diff
@@ -85,4 +85,26 @@ public void toggleStatus(ActionRequest request, ActionResponse response) {
...
+
+  public void changeConfig(ActionRequest request, ActionResponse response) {
+    try {
+      Journal journal = request.getContext().asType(Journal.class);
+
+      if (journal.getCompany().getAccountConfig() == null) {
+        response.setValue("authorizeSimulatedMove", false);
+        response.setValue("allowAccountingDaybook", false);
+      } else {
+        if (journal.getCompany().getAccountConfig().getIsActivateSimulatedMove() == null
```
jmi-axelor, line 97: You don't need to check if the IsActivateSimulatedMove is null, it's a boolean, not a Boolean, it can't be null

axelor-account/src/main/java/com/axelor/apps/account/web/JournalController.java
```diff
@@ -85,4 +85,26 @@ public void toggleStatus(ActionRequest request, ActionResponse response) {
       TraceBackService.trace(response, e);
     }
   }
+
+  public void changeConfig(ActionRequest request, ActionResponse response) {
+    try {
+      Journal journal = request.getContext().asType(Journal.class);
+
```
jmi-axelor, line 92: Could you declare a variable here to store journal.getCompany().getAccountConfig() and use it after? 
The code would be cleaner

axelor-account/src/main/java/com/axelor/apps/account/web/JournalController.java
```diff
@@ -85,4 +85,26 @@ public void toggleStatus(ActionRequest request, ActionResponse response) {
...
+      if (journal.getCompany().getAccountConfig() == null) {
+        response.setValue("authorizeSimulatedMove", false);
+        response.setValue("allowAccountingDaybook", false);
+      } else {
+        if (journal.getCompany().getAccountConfig().getIsActivateSimulatedMove() == null
+            || !journal.getCompany().getAccountConfig().getIsActivateSimulatedMove()) {
+          response.setValue("authorizeSimulatedMove", false);
```
jmi-axelor, line 99: Is it possible to just do:
response.setValue("authorizeSimulatedMove", journal.getCompany().getAccountConfig().getIsActivateSimulatedMove())
?

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/web/JournalController.java
ija-axelor: i can't do that if journal.getCompany().getAccountConfig() is null i will have a NPE

axelor-account/src/main/resources/views/Journal.xml
```diff
@@ -154,5 +155,7 @@
   <action-method name="action-journal-method-set-totals">
     <call class="com.axelor.apps.account.web.JournalController" method="computeBalance"/>
   </action-method>
-
+  <action-method name="action-journal-method-change-config-values">
+    <call class="com.axelor.apps.account.web.JournalController" method="changeConfig"/>
+  </action-method>
```
jmi-axelor, line 160: Now that I think about it, it would be so much easier to use an action-record like this:
``` xml
  <action-record name="action-journal-record-change-config-values"
    model="com.axelor.apps.account.db.Journal">
    <field name="authorizeSimulatedMove" expr="eval: company?.accountConfig?.isActivateSimulatedMove"/>
    <field name="allowAccountingDaybook" expr="eval: company?.accountConfig?.accountingDaybook"/>
  </action-record>
```
Also with your action-method, if the user delete the company field, the method will be called and they will get an NPE line 93.

jmi-axelor (changes requested):

PR #7745, opened 2022-03-18
Anomaly #47393 : Checking before changing status : CRM

alo-axelor:


axelor-crm/src/main/java/com/axelor/apps/crm/service/LeadService.java
```diff
@@ -69,13 +69,37 @@ public Lead convertLead(Lead lead, Partner partner, Partner contactPartner)
...
+   * Set the lead status to In Process.
+   *
+   * @param lead
+   * @throws AxelorException
+   */
+  void startLead(Lead lead) throws AxelorException;
+
+  /**
+   * Set the lead status to In Process.
+   *
```
kra-axelor, line 82: Can you add more precision ? By just reading the javadoc, this method is the same as `startLead` or `recycleLead`

axelor-crm/src/main/java/com/axelor/apps/crm/service/LeadServiceImpl.java
```diff
@@ -230,8 +242,63 @@ public boolean isThereDuplicateLead(Lead lead) {
...
+    if (lead.getStatusSelect() == null || !authorizedStatus.contains(lead.getStatusSelect())) {
+      throw new AxelorException(
+          TraceBackRepository.CATEGORY_INCONSISTENCY,
+          I18n.get(com.axelor.apps.crm.exception.IExceptionMessage.LEAD_ASSIGN_TO_ME_WRONG_STATUS));
+    }
+    lead.setUser(AuthUtils.getUser());
+    if (lead.getStatusSelect() == LeadRepository.LEAD_STATUS_NEW) {
+      lead.setStatusSelect(LeadRepository.LEAD_STATUS_ASSIGNED);
+    }
+    saveLead(lead);
```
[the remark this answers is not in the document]
alo-axelor: Done

axelor-crm/src/main/java/com/axelor/apps/crm/service/LeadService.java
alo-axelor: Done

axelor-crm/src/main/java/com/axelor/apps/crm/web/LeadController.java
```diff
@@ -190,4 +190,49 @@ public void loseLead(ActionRequest request, ActionResponse response) {
...
+    }
+  }
+
+  public void assignToMeMultipleLead(ActionRequest request, ActionResponse response) {
+    try {
+      LeadRepository leadRepo = Beans.get(LeadRepository.class);
+      for (Lead lead :
+          leadRepo.all().filter("id in ?1", request.getContext().get("_ids")).fetch()) {
+        Beans.get(LeadService.class)
+            .assignToMeLead(Beans.get(LeadRepository.class).find(lead.getId()));
```
kra-axelor, line 231: Sorry I didn't noticed this earlier. When calling `assignToMeLead` on multiple lead, the process will stop at the first exception and then the next leads will not be treated, but the previous ones would be correctly assigned.

Is it the intended behaviour ?

alo-axelor: I didn't think about it when I wrote it, but would it be correct to keep it that way ? Or we should cancel the correctly assigned leads when an exception occurs ?

kra-axelor: We have the same behaviour on fixed asset on mass validation so it is correct to keep it that way, I just wanted to be sure that this was the intended behaviour.

alo-axelor: Ok, I think I will keep it that way then.

kra-axelor (approved):

PR #7758, opened 2022-03-22
RM#46616: add refund to payment session and added automatic reconciliation between invoice and refund in session payment validation

jdo-axelor:
… invoice term paid amount


axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
```diff
@@ -185,6 +186,33 @@ public void reconcileMoveLines(List<MoveLine> moveLineList) {
     }
   }
 
+  @Override
+  @Transactional(rollbackOn = {Exception.class})
+  public void reconcileMoveLinesWithFullRollBack(List<MoveLine> moveLineList)
```
jmi-axelor, line 191: This method is never called, can it be deleted?

jmi-axelor (changes requested):

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #7765, opened 2022-03-22
RM#47595 : Fix NPE when creating invoice term from move line

aol-axelor:


ale-axelor (changes requested):
Please review the methods, a service method should not use Context objects as they are specific to the views and must only be called in Controller classes.

jmi-axelor:
Need to apply spotless on your branch

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -110,12 +126,26 @@ public void initInvoiceTermFromInvoice(ActionRequest request, ActionResponse res
           invoiceTermService.initCustomizedInvoiceTerm(invoice, invoiceTerm);
           response.setValues(invoiceTerm);
         } else {
-          MoveLine moveLine = ContextTool.getContextParent(request.getContext(), MoveLine.class, 1);
+          if (request.getContext().getParent().get("_model").toString().contains("MoveLine")) {
```
jmi-axelor, line 129: Why do you replace the ContexTool.getContextParent?

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -110,12 +126,26 @@ public void initInvoiceTermFromInvoice(ActionRequest request, ActionResponse res
...
+                      .getContext()
+                      .getParent()
+                      .getParent()
+                      .get("_model")
+                      .equals(Move.class.getName())) {
+                move = request.getContext().getParent().getParent().asType(Move.class);
+              }
 
-          if (moveLine != null) {
-            Move move = ContextTool.getContextParent(request.getContext(), Move.class, 2);
```
jmi-axelor, line 116: Same here

PR #7776, opened 2022-03-23
RM#47375 : Generate financial discount move lines from payment session

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateServiceImpl.java
```diff
@@ -641,7 +642,12 @@ protected void checkInactiveJournal(Move move) throws AxelorException {
 
   protected void validateVatSystem(Move move) throws AxelorException {
     if (!CollectionUtils.isEmpty(move.getMoveLineList())) {
-      if (isConfiguredVatSystem(move) && isConfigurationIssueOnVatSystem(move)) {
+      if ((move.getJournal().getJournalType().getTechnicalTypeSelect()
```
ale-axelor, line 645: Need to add a check to now if journal type is filled, else we will get a NPE since the field is not required.

ale-axelor (changes requested):

PR #7780, opened 2022-03-23
Dev 46678

dtr-axelor:


dtr-axelor:
To merge after https://github.com/axelor/axelor-open-suite/pull/7602

PR #7799, opened 2022-03-25
RM47824 MRP : show message on NPE on manufOrder element

erg-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/MrpServiceImpl.java
```diff
@@ -1482,6 +1482,13 @@ protected MrpLine createMrpLine(
             I18n.get(IExceptionMessage.MRP_NO_PRODUCT_UNIT),
             product.getFullName());
       }
+      if (product.getId() == null) {
```
gdu-axelor, line 1485: Exception should be throw if product.getId() is not defined in the  this.productMap because product.getId will never return a NPE.

gdu-axelor (changes requested):

PR #7821, opened 2022-03-29
RM47845 MRP take into account the alternatives BOM

erg-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -431,4 +431,23 @@ public BillOfMaterial getDefaultBOM(Product originalProduct, Company company)
 
     return billOfMaterial;
   }
+
+  @Override
+  public List<BillOfMaterial> getAlternativesBOM(Product originalProduct, Company company)
+      throws AxelorException {
+    if (company == null) {
```
gdu-axelor, line 438: company is required on the mrp.stockLocation. No need to use the active company

gdu-axelor (changes requested):

PR #7888, opened 2022-04-13
US 89 - Identification of 'Value for Collection' type accounting acco…

cgu-axelor:
Issue #47940
US 89 - Identification of "Value for Collection" type accounting accounts


jmi-axelor (commented):
This PR is already included in the https://github.com/axelor/axelor-open-suite/pull/7890. 
@cgu-axelor can you confirm and if it is, can you close this one?

PR #7898, opened 2022-04-14
Anomaly RM#48159:ADDRESS / APPBASE : Geoname import on empty database…

vva-axelor:
… (without demodata) seems no longer operational


jmi-axelor:
This PR can be closed, new PR 8019

PR #7900, opened 2022-04-14
Feature #48383 : Data backup : demo data for anonymization

alo-axelor:


axelor-admin/src/main/java/com/axelor/apps/base/service/app/AnonymizeServiceImpl.java
```diff
@@ -0,0 +1,84 @@
...
+        return BigDecimal.ZERO;
+
+      case DATE:
+        return LocalDate.of(1970, 1, 1);
+
+      case TIME:
+        return LocalTime.of(0, 0, 0);
+
+      case DATETIME:
+        return LocalDateTime.of(LocalDate.of(1970, 1, 1), LocalTime.of(0, 0, 0));
```
mch-axelor, line 50: Could be better to use LocalDateTime.MIN or LocalDateTime.MAX

PR #7939, opened 2022-04-25
48784 PaymentVoucher : bank details set when partner change and payme…

rth-axelor:
…nt mode contains only one bank details


ssb-axelor:
this PR can be closed. a development was made within ticket #47990

PR #7959, opened 2022-04-27
Anomaly #48673: Changed error message when stock location is present …

kra-axelor:
…but not useable for outsourcing


axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderStockMoveService.java
```diff
@@ -142,6 +143,27 @@ public StockMove _createToConsumeStockMove(ManufOrder manufOrder, Company compan
             ? company.getStockConfig().getOutsourcingReceiptStockLocation()
             : getDefaultStockLocation(manufOrder, company, STOCK_LOCATION_IN);
 
+    if (fromStockLocation == null) {
```
kra-axelor, line 146: I'm not sure about adding these check in this method. Depends on wether we allow source stock location and destination stock location to be null or not.

kra-axelor, line 146: I'm not sure about adding these check in this method. Depends on wether we allow source stock location and destination stock location to be null or not.

ale-axelor: the method to plan the created stock move is always called, so it's better to add the check here so we can have a better error message

PR #7963, opened 2022-04-27
44883 Fix of invoice grid to have only the lines according to the pre…

rth-axelor:
…sent filter


axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceController.java
```diff
@@ -368,7 +369,8 @@ public void showInvoice(ActionRequest request, ActionResponse response) {
                         .collect(Collectors.toList()));
         fileLink = Beans.get(InvoicePrintService.class).printInvoices(ids);
         title = I18n.get("Invoices");
-      } else if (context.get("id") != null) {
+      } else if (context.get("id") != null
+          && Invoice.class.toString().equals(context.getContextClass().toString())) {
```
jmi-axelor, line 373: Don't need to use toString()

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #7991, opened 2022-04-28
RM#48803: fix NPE in fillMoveWithFinancialDiscount

jmi-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentValidateServiceImpl.java
```diff
@@ -335,45 +337,37 @@ public Move fillMoveWithFinancialDiscount(
...
-
-      move =
-          fillMoveLinesWithFinancialDiscount(
-              invoicePayment,
-              move,
-              customerAccount,
-              purchAccount,
-              accountConfigService.getAccountConfig(company).getPurchFinancialDiscountAccount());
-
+      financialDiscountTax = accountConfig.getPurchFinancialDiscountTax();
```
gdu-axelor, line 343: Please use accountServiceConfig to check not null constraint

axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentValidateServiceImpl.java
```diff
@@ -335,45 +337,37 @@ public Move fillMoveWithFinancialDiscount(
...
+    }
 
-      move =
-          fillMoveLinesWithFinancialDiscount(
-              invoicePayment,
-              move,
-              customerAccount,
-              saleAccount,
-              accountConfigService.getAccountConfig(company).getSaleFinancialDiscountAccount());
+    if (account == null) {
```
gdu-axelor, line 363: Please, add a method into AccountManagmentService to get the correct value. Also, if null throw an exception instead of return a partial move

gdu-axelor (changes requested):

PR #8029, opened 2022-05-03
49113 add of an action group in account period

rth-axelor:


axelor-account/src/main/resources/views/Period.xml
```diff
@@ -66,6 +66,12 @@
     <call class="com.axelor.apps.account.web.PeriodController" method="setReadOnly"/>
   </action-method>
 
+  <action-group name="action-period-group-onload">
```
jmi-axelor, line 69: Shouldn't we also call it in account-period-year-form?

jmi-axelor (approved):

PR #8052, opened 2022-05-05
Anomaly #49215: Fixed saleOrderLine price that was not recomputed on …

kra-axelor:
…click of update pricing scale + also updated SaleOrder price

### DIFF
--- a/axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
+++ b/axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
@@ -300,6 +300,13 @@ public class SaleOrderLineServiceImpl implements SaleOrderLineService {
     return line;
   }
 
+  public void resetPrice(SaleOrderLine line) {
+    if (!line.getEnableFreezeFields()) {
+      line.setPrice(null);
+    }
+    line.setInTaxPrice(null);
+  }
+
   @Override
   public Map<String, BigDecimal> computeValues(SaleOrder saleOrder, SaleOrderLine saleOrderLine)
       throws AxelorException {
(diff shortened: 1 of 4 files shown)


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
```diff
@@ -300,6 +300,13 @@ public SaleOrderLine resetProductInformation(SaleOrderLine line) {
     return line;
   }
 
+  public void resetPrice(SaleOrderLine line) {
```
ale-axelor, line 303: please add `@Override` annotation

axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
```diff
@@ -300,6 +300,13 @@ public SaleOrderLine resetProductInformation(SaleOrderLine line) {
...
+    if (!line.getEnableFreezeFields()) {
+      line.setPrice(null);
+    }
+    line.setInTaxPrice(null);
```
ale-axelor, line 307: Not sure about this, if fields are frozen, then this will empty inTaxPrice and it will not be updated by fillPrice, emptying the inTaxPrice value while we want to keep the original value ?

ale-axelor (changes requested):

axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderLineServiceImpl.java
kra-axelor: Yes you're right, fixed it

PR #8102, opened 2022-05-11
RM#49406: fix AccountingSituation views

jmi-axelor:


axelor-account/src/main/resources/views/AccountingSituation.xml
```diff
@@ -251,7 +251,7 @@
   <action-record name="action-record-manage-vat-system-onnew"
     model="com.axelor.apps.account.db.AccountingSituation">
     <field name="vatSystemSelect" expr="eval: 1"
-      if="partner != null &amp;&amp; (partner.isCustomer != null &amp;&amp; partner.isCustomer) || (partner.isSupplier != null &amp;&amp; partner.isSupplier)"/>
+      if="partner?.isCustomer || partner?.isSupplier"/>
```
gdu-axelor, line 254: Should use __parent__ if appears on onNew from Partner form.

gdu-axelor (changes requested):

axelor-account/src/main/resources/views/AccountingSituation.xml
jmi-axelor: I will create a ticket in AOS sigma

PR #8107, opened 2022-05-12
 Feature #41571: Manage attrs field for pricing rule 

kra-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingComputer.java
```diff
@@ -126,7 +127,7 @@ public void apply() throws AxelorException {
     for (int counter = 0; counter < MAX_ITERATION; counter++) {
 
       Optional<Pricing> optChildPricing = getNextPricing(currentPricing);
-      if (optChildPricing.flatMap(this::applyPricing).isPresent()) {
+      if (optChildPricing.isPresent() && applyPricing(optChildPricing.get()).isPresent()) {
```
ale-axelor, line 130: flatMap does not work here ?

axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingComputer.java
```diff
@@ -197,24 +200,50 @@ protected void computeResultFormulaAndApply(Pricing pricing, PricingLine pricing
...
-                LOG.debug(
-                    "Adding result temp variable {} in context",
-                    resultPricingRule.getTempVarName());
-                putInContext(resultPricingRule.getTempVarName(), result);
-              }
-              notifyResultPricingRule(resultPricingRule, result);
-              notifyFieldToPopulate(fieldToPopulate);
-            });
+    for (PricingRule resultPricingRule : resultPricingRuleList) {
+      if (resultPricingRule != null) {
```
ale-axelor, line 204: I'm surprised we could have a list with null element here. I don't know if it can happen but we should always avoid adding null element in a list

axelor-base/src/main/resources/views/PricingRule.xml
```diff
@@ -26,13 +27,30 @@
...
+  </action-attrs>
+
+  <action-attrs name="action-pricing-rule-attrs-display-meta-json-field">
+    <attribute name="hidden" for="metaJsonField"
+      expr="eval:__this__?.fieldToPopulate.json == false"
+      if="eval:__this__?.fieldToPopulate != null"/>
+    <attribute name="required" for="metaJsonField"
+      expr="eval:__this__?.fieldToPopulate.json == true" if="eval:__this__?.fieldToPopulate != null"/>
+    <attribute name="value" for="metaJsonField" expr="eval:null"
+      if="eval:__this__?.fieldToPopulate== null || __this__?.fieldToPopulate.json == false"/>
```
ale-axelor, line 53: be careful when using safe dereferencing with boolean comparison since `null == false` will return `false`. But I don't know if this cause an issue since I suspect that `__this__` will never be null

ale-axelor (changes requested):

axelor-base/src/main/java/com/axelor/apps/base/service/MetaJsonAttrsService.java
```diff
@@ -0,0 +1,179 @@
...
+        LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    static Map.Entry<String, Object> adaptValueForMap(MetaJsonField metaJsonField, Object value)
+        throws AxelorException {
+
+      Objects.requireNonNull(value);
+      String wantedType = MetaTool.jsonTypeToType(metaJsonField.getType());
+      String fieldName = metaJsonField.getName();
+      logger.debug("Adapting value {} into {}", value, wantedType);
+      switch (wantedType) {
```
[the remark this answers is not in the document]
kra-axelor: Hmmm, do you mean as static value in MetaJsonFieldRepository ? I could not find them

PR #8150, opened 2022-05-13
49566 fix of npe in bank reconciliation service when query is null

rth-axelor:


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationService.java
```diff
@@ -210,13 +210,17 @@ public void generateMovesAutoAccounting(BankReconciliation bankReconciliation)
...
-          if (Boolean.TRUE.equals(
-              new GroovyScriptHelper(scriptContext)
-                  .eval(
-                      bankStatementRule
-                          .getBankStatementQuery()
-                          .getQuery()
-                          .replaceAll("%s", "\"" + bankStatementRule.getSearchLabel() + "\"")))) {
+          if (bankStatementRule != null
+              && bankStatementRule.getBankStatementQuery() != null
+              && bankStatementRule.getBankStatementQuery().getQuery() != null
```
jmi-axelor, line 215: Could you use com.google.common.base.Strings.isNullOrEmpty instead of != null, it's better when we have a string.

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #8172, opened 2022-05-17
Feature #47325 : Refactor IExceptionMessage

alo-axelor:


axelor-admin/src/main/java/com/axelor/apps/base/exceptions/IExceptionMessages.java
```diff
@@ -17,6 +17,7 @@
  */
 package com.axelor.apps.base.exceptions;
 
+@Deprecated
```
dtr-axelor, line 20: @alo-axelor I think it would be better to also add the javadoc tag `@deprecated` to say which class should now be used. Same for all the other classes you deprecated.

alo-axelor: Good idea, I will do this.

kra-axelor (approved):

PR #8177, opened 2022-05-18
49703 change of message in accounting for vat system error

rth-axelor:


axelor-account/src/main/java/com/axelor/apps/account/exception/IExceptionMessage.java
```diff
@@ -1124,7 +1124,7 @@ public interface IExceptionMessage {
       "Tax lines can't be computed due to missing value in the field vat system configuration for the account %s." /*)*/;
 
   static final String TAX_MOVELINE_VAT_SYSTEM_DEFAULT = /*$$(*/
-      "At least one tax accounting move line associated to this move is not correctly set in terms of vat system settings. In order to avoid troubles in the tax declaration, we advise you to use the button generate tax while entering manually a move to avoid such misconfiguration. The vat system value can be changed on the tax move line in the tab 'Others'" /*)*/;
+      "At least one tax accounting move line associated to the move %s is not correctly set in terms of vat system settings. In order to avoid troubles in the tax declaration, we advise you to use the button generate tax while entering manually a move to avoid such misconfiguration. The vat system value can be changed on the tax move line in the tab 'Others'" /*)*/;
```
jmi-axelor, line 1127: It doesn't work, the move is never set in the exception, so the user will see exactly the following message with the "%s" inside: 
"At least one tax accounting move line associated to the move %s is not correctly set in terms of vat system settings. In order to avoid troubles in the tax declaration, we advise you to use the button generate tax while entering manually a move to avoid such misconfiguration. The vat system value can be changed on the tax move line in the tab 'Others'"

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #8193, opened 2022-05-19
Anomaly #49664 : generated bankOrder has a different bankdetails

abo-axelor:


jdo-axelor:
Corrected on https://github.com/axelor/axelor-open-suite/pull/8388

PR #8211, opened 2022-05-23
Anomaly #49211: Changed context to use interface AnalyticLine instead

kra-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/analytic/AnalyticMoveLineServiceImpl.java
```diff
@@ -229,83 +228,40 @@ public AnalyticMoveLine computeAnalyticMoveLine(
...
-    } else if (InvoiceLine.class.equals(parent.getContextClass())) {
-      InvoiceLine line = parent.asType(InvoiceLine.class);
-      if (line.getAccount() != null && line.getAccount().getCompany() != null) {
-        return appBaseService.getTodayDate(line.getAccount().getCompany());
       }
     }
+    if (parent.getAccount() != null && parent.getAccount().getCompany() != null) {
+      return appBaseService.getTodayDate(parent.getAccount().getCompany());
+    }
     return LocalDate.now();
```
ale-axelor, line 265: appBaseService.getTodayDate(Company company) is a method that can be called if company is null.
So LocalDate.now() should not be called but appBaseService should also be used.

ale-axelor, line 265: appBaseService.getTodayDate(Company company) is a method that can be called if company is null.
So LocalDate.now() should not be called but appBaseService should also be used.

PR #8235, opened 2022-05-25
Anomaly #50023 Don't set onChange element for non editable custom fields

dtr-axelor:
Signed-off-by: David Trochel <d.trochel@axelor.com>


dtr-axelor:
@ale-axelor I think that we should also refactor some other parts of the code to not rely on hard-coded strings for `MetaJsonField.type`.

I checked quickly where `MetaJsonField.getType()` and I noticed for example this method: https://github.com/axelor/axelor-open-suite/blob/master/axelor-tool/src/main/java/com/axelor/apps/tool/MetaTool.java#L83

But ideally, these constants should be defined on a lower level (ie AOP).

PR #8242, opened 2022-05-27
Anomaly #46495 : Product : Computation of salePrice from costPrice doesn't use unit conversion

alo-axelor:


axelor-base/src/main/java/com/axelor/apps/base/web/ProductCompanyController.java
```diff
@@ -0,0 +1,26 @@
...
+    if (productCompany.getAutoUpdateSalePrice()) {
+      response.setValue(
+          "salePrice",
+          Beans.get(ProductService.class)
+              .computeSalePrice(
+                  productCompany.getManagPriceCoef(),
+                  productCompany.getCostPrice(),
+                  productCompany.getProduct(),
+                  productCompany.getCompany()));
+    }
```
ale-axelor, line 24: please use a try catch to catch Exception

ale-axelor (changes requested):
We also have to manage the case where autoUpdateSalePrice is not managed per company

PR #8245, opened 2022-05-30
Feature #29874: Added content currency for AccountingReport1, AccountingReport7 and AccountingReport14

kra-axelor:


axelor-account/src/main/resources/views/Selects.xml
```diff
@@ -664,4 +664,9 @@
     <option value="1">Direct bank LCR</option>
   </selection>
 
+  <selection name="account.accounting.report.content.currency.select">
+    <option value="1">Company's currency</option>
```
ale-axelor, line 668: Please rename this option to "Company currency" to match other titles

axelor-account/src/main/resources/views/Selects.xml
```diff
@@ -664,4 +664,9 @@
...
+    <option value="2">Origin's currency</option>
```
ale-axelor, line 669: Please rename this option to "Original currency"

ale-axelor (changes requested):

PR #8256, opened 2022-05-31
Feature #50161 - Stock API : add API for Supplier arrivals

lme-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveServiceImpl.java
```diff
@@ -1429,4 +1432,57 @@ public void updateStockMoveMobility(StockMove stockMove, BigDecimal movedQty, Un
...
+    }
+    newLine.setQty(qty);
+    newLine.setRealQty(realQty);
+    newLine.setUnit(unit);
+    newLine.setConformitySelect(conformitySelect);
+
+    stockMoveLineService.compute(newLine, stockMove);
+
+    stockMoveLineRepo.save(newLine);
+  }
```
ale-axelor, line 1487: This method is only creating and saving a new stock move. This method must be moved in StockMoveLineService and called "createStockMoveLine", and maybe calls existing `createStockMoveLine` methods.

ale-axelor (changes requested):

PR #8296, opened 2022-06-07
50296 refacto to avoid npe when movelines are created with a null acc…

rth-axelor:
…ount


jmi-axelor (approved):

axelor-account/src/main/java/com/axelor/apps/account/service/AccountManagementServiceAccountImpl.java
```diff
@@ -222,4 +223,16 @@ protected FixedAssetCategory getProductFixedAssetCategory(
 
     return fixedAssetCategory;
   }
+
+  @Override
+  public Account getCashAccount(AccountManagement accountManagement, PaymentMode paymentMode)
+      throws AxelorException {
+    if (accountManagement != null && accountManagement.getCashAccount() == null) {
```
jmi-axelor, line 230: If accountManagement is null, we get an NPE, please add a if condition before to check if accountManagement is null and return null. And also in this new if condition, check if paymentMode is null.

jmi-axelor (changes requested):

PR #8306, opened 2022-06-08
Anomaly #46931: Fixed first depreciation date on generation from invoice + manage null value of disposalAmount

kra-axelor:


jmi-axelor (changes requested):
Everything is ok, just could you add a changelog file please?

jmi-axelor (approved):

PR #8323, opened 2022-06-09
RM#7978 Translate every debug message in english

alo-axelor:


kra-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/debtrecovery/DebtRecoverySessionService.java
```diff
@@ -115,15 +115,15 @@ public DebtRecovery debtRecoverySession(DebtRecovery debtRecovery) throws Axelor
     if ((appAccountService.getTodayDate(debtRecovery.getCompany()).isAfter(referenceDate)
             || appAccountService.getTodayDate(debtRecovery.getCompany()).isEqual(referenceDate))
         && balanceDueDebtRecovery.compareTo(BigDecimal.ZERO) > 0) {
-      log.debug(
-          "Si la date actuelle est égale ou ultérieur à la date de référence et le solde due relançable positif");
+      // If the current date is equal or after the reference date and the balance due debt recovery
```
kra-axelor, line 118: Sorry what I said was not clear. What I meant was that the debug messages were not appropriate and that they looked like java comments. We still want the log.debug, but the messages must be changed, and not be conditional and should give more informations (by displaying the value of the fields)
For example `log.debug("Current date {} is after reference date {} and balance due debt = {}", currentDate, referenceDate, balanceDueDebt)`

kra-axelor (approved):

PR #8332, opened 2022-06-13
RM#50590 : Improve invoice term on move line change

aol-axelor:


axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -596,8 +597,8 @@
         <button name="showInvoiceTermsBtn" title="Show related invoice term(s)"
           onClick="save,action-move-line-view-show-related-invoice-terms"
           readonlyIf="!invoiceTermList.length"/>
-        <field name="invoiceTermList" colSpan="12" form-view="invoice-term-move-line-form"
-          grid-view="invoice-term-move-line-grid"/>
+        <field name="invoiceTermList" colSpan="12" onChange="save"
```
gdu-axelor, line 600: we can't save a moveLine each time we update the invoiceTermList, else we will have an account move not correctly balanced on the daybook status.

axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -535,7 +535,8 @@
       </field>
       <panel name="datesPanel" colSpan="12">
         <field name="date" onChange="account-move-line-group-date-onchange"/>
-        <field name="dueDate" showIf="dueDate"/>
+        <field name="dueDate" showIf="dueDate"
+          onChange="action-move-line-validate-invoice-term-alert"/>
```
gdu-axelor, line 539: dueDate should be displayed even if null.

gdu-axelor (changes requested):

PR #8389, opened 2022-06-22
RM#51125 Compute $availableQty only when it's necessary

dtr-axelor:


dtr-axelor:
@ale-axelor I can create another PR for 2fea27743deae181caa7db934c89c6a0624fa42c and b164101c8578350a76e23a53dcfc658b75ad2b39 if you want, these commits are just some refactor on some code called by the processes I tested, but they are not really required.

PR #8404, opened 2022-06-27
Anomaly #51259 : Stock Correction: enable to do a stock correction for product with no quanty on the stock location

alo-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockCorrectionServiceImpl.java
```diff
@@ -134,9 +151,10 @@ public StockMove generateStockMove(StockCorrection stockCorrection) throws Axelo
               stockCorrection.getTrackingNumber());
     }
 
-    BigDecimal realQty = stockCorrection.getRealQty();
-    Product product = stockCorrection.getProduct();
-    TrackingNumber trackingNumber = stockCorrection.getTrackingNumber();
+    if (stockLocationLine == null) {
+      stockLocationLine = stockLocationLineService.createLocationLine(toStockLocation, product);
+    }
+
```
ale-axelor, line 157: Please instead call earlier the dedicated methods in StockLocationLineService `getOrCreateStockLocationLine` and `getOrCreateDetailLocationLine` as this current modification will not create the stock location with a tracking number

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockCorrectionServiceImpl.java
```diff
@@ -190,19 +208,16 @@ public StockMove generateStockMove(StockCorrection stockCorrection) throws Axelo
...
+        null,
+        null,
+        company,
+        fromStockLocation,
+        toStockLocation,
+        null,
+        null,
+        null,
+        StockMoveRepository.TYPE_INTERNAL);
   }
```
ale-axelor, line 221: this method should at least not be public, please move it to protected.
If it is called only one time, it is better to remove this method and directly call the stock move service

ale-axelor (changes requested):

ale-axelor (changes requested):

PR #8483, opened 2022-07-07
RM#51835 : Fix rounding issue when ventilating multi currency invoice

aol-axelor:


jmi-axelor:
Approved but must be reverted when multi currency is managed on InvoiceTerms

PR #8498, opened 2022-07-11
Anomaly #52042 Optimize query to reduce process time on update stock history batch

jbo-axelor:
Anomaly #52042


ale-axelor:
moved to #8917

PR #8585, opened 2022-07-20
Anomaly #52272: Added constant for fix constraint issue configurator

kra-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/db/repo/ConfiguratorCreatorSaleRepository.java
```diff
@@ -33,6 +33,8 @@
 
 public class ConfiguratorCreatorSaleRepository extends ConfiguratorCreatorRepository {
 
+  protected static final String AXELOR_TMP_FIX_CONSTRAINT = "$AXELORTMP";
```
dtr-axelor, line 36: The constant is duplicated in 3 classes, this approach doesn't improve a lot the maintainability. I recommend making the constant public in one of the class so that other classes can use it. As a consequence, they won't need to declare the constant anymore.

dtr-axelor (changes requested):

axelor-sale/src/main/java/com/axelor/apps/sale/db/repo/ConfiguratorCreatorSaleRepository.java
kra-axelor: Indeed, fixed it. Thank you for the feedback

dtr-axelor: That's better, thank you.

However, I don't think the repo is the best location where to declare the constant because it's not related to data access. But I will let @ale-axelor give his opinion.

dtr-axelor:
@ale-axelor why did you close it? This PR is still useful

PR #8626, opened 2022-07-25
52395 add of required of on invoice bank details field to match invoi…

rth-axelor:
…ce term's field condition


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceTermServiceImpl.java
```diff
@@ -326,6 +326,14 @@ public InvoiceTerm initCustomizedInvoiceTerm(Invoice invoice, InvoiceTerm invoic
...
     if (invoice.getStatusSelect() == InvoiceRepository.STATUS_VENTILATED) {
       MoveLine moveLine = getExistingInvoiceTermMoveLine(invoice);
+      if (moveLine == null && !CollectionUtils.isEmpty(invoice.getMove().getMoveLineList())) {
+        for (MoveLine ml : invoice.getMove().getMoveLineList()) {
+          if (ml.getAccount().getHasInvoiceTerm()) {
+            moveLine = ml;
+            ml.addInvoiceTermListItem(invoiceTerm);
+          }
+        }
+      }
```
ale-axelor, line 336: Please add a new method with this code. Since everything is in the same method, we have multiple move line variables which is confusing

ale-axelor (changes requested):

PR #8630, opened 2022-07-26
Anomaly RM#52645: NPE on ventilate expense

hka-axelor:


ale-axelor:
This is a good change but the feature was still not working correctly after, opened a new PR to refactor everything and use the correct methods: #8713

PR #8692, opened 2022-07-29
52926 add of 2 fields in accounting batch to configure cut off move

rth-axelor:


axelor-supplychain/src/main/resources/views/AccountingBatch.xml
```diff
@@ -3,6 +3,32 @@
...
+    </extend>
+  </form>
+
+  <action-group name="action-group-accounting-batch-onchange-accounting-type"
+    id="supplychain-action-group-accounting-batch-onchange-accounting-type">
+    <action name="action-accounting-batch-attrs-show-label"/>
+    <action name="action-record-accounting-batch-set-expenses-incomes"/>
+    <action name="action-attrs-accounting-batch-cutoff-move-status-domain"/>
+    <action name="action-method-accounting-batch-set-default-accounts"/>
+  </action-group>
```
aol-axelor, line 30: Please create a new action group calling the extended one, so that if the extended one is changed there is no need to also change that new one.

axelor-supplychain/src/main/resources/views/AccountingBatch.xml
```diff
@@ -3,6 +3,32 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://axelor.com/xml/ns/object-views http://axelor.com/xml/ns/object-views/object-views_5.4.xsd">
 
+  <form name="accounting-batch-form" id="supplychain-accounting-batch-form"
+    title="Accounting Batch" model="com.axelor.apps.account.db.AccountingBatch" extension="true">
+    <extend target="//field[@name='company']">
+      <attribute name="onChange"
+        value="action-accounting-batch-attrs-set-empty-year-and-accountset-value,action-method-accounting-batch-set-default-accounts"/>
```
aol-axelor, line 10: Please create an action group

axelor-supplychain/src/main/resources/views/AccountingBatch.xml
```diff
@@ -3,6 +3,32 @@
...
+    </extend>
+    <extend target="//field[@name='miscOpeJournal']">
+      <insert position="before">
+        <field name="forecastedInvCustAccount" requiredIf="accountingCutOffTypeSelect &lt;= 2"
+          domain="self.statusSelect = 1 AND self.company = :company" grid-view="account-grid"
+          form-view="account-form" colSpan="6" showIf="accountingCutOffTypeSelect &lt;= 2"/>
+        <field name="forecastedInvSuppAccount" requiredIf="accountingCutOffTypeSelect &lt;= 2"
+          domain="self.statusSelect = 1 AND self.company = :company" grid-view="account-grid"
+          form-view="account-form" colSpan="6" showIf="accountingCutOffTypeSelect &lt;= 2"/>
```
aol-axelor, line 19: There is a `showIf` on the panel containing `miscOpeJournal` so maybe its condition should be added to the `requiredIf` to prevent any issue

axelor-account/src/main/resources/views/AccountingBatch.xml
```diff
@@ -163,7 +163,7 @@
         <field name="moveDescription" requiredIf="actionSelect == 26"/>
         <panel colSpan="6" itemSpan="12">
           <field name="moveDate" requiredIf="actionSelect == 26"
-            validIf="$moment(moveDate).date() == $moment(moveDate).endOf('month').date()"
+            validIf="actionSelect != 26 || $moment(moveDate).date() == $moment(moveDate).endOf('month').date()"
```
ale-axelor, line 166: is this change wanted ? it seems that it is not related to the issue.

axelor-supplychain/src/main/resources/views/AccountingBatch.xml
```diff
@@ -3,6 +3,41 @@
...
+    <extend target="//field[@name='miscOpeJournal']">
+      <insert position="before">
+        <field name="forecastedInvCustAccount"
+          requiredIf="actionSelect == 26 &amp;&amp; accountingCutOffTypeSelect &lt;= 2"
+          domain="self.statusSelect = 1 AND self.company = :company" grid-view="account-grid"
+          form-view="account-form" colSpan="6" showIf="accountingCutOffTypeSelect &lt;= 2"/>
+        <field name="forecastedInvSuppAccount"
+          requiredIf="actionSelect == 26 &amp;&amp; accountingCutOffTypeSelect &lt;= 2"
+          domain="self.statusSelect = 1 AND self.company = :company" grid-view="account-grid"
+          form-view="account-form" colSpan="6" showIf="accountingCutOffTypeSelect &lt;= 2"/>
```
ale-axelor, line 25: the requiredIf and showIf condition is the same for both fields, is this correct ? I was expecting something like accountingCutOffTypeSelect = 2 for forecastedInvCustAccount and accountingCutOffTypeSelect = 1 for forecastedInvSuppAccount.
Also shouldn't it be hidden if the batch is not the cut-off ?

ale-axelor (changes requested):
Also please remove `axelor-*/src/main/resources/i18n/messages_{de,es,it,nl,pl,pt,ru}.csv` from the list of changes.

axelor-account/src/main/resources/views/AccountingBatch.xml
rth-axelor: Yes, this fix is to be on wip-6.4 so can be done here

axelor-supplychain/src/main/resources/views/AccountingBatch.xml
rth-axelor: We need both, process will choose one if the stockmove is a sale or a purchase stockmove.

It's already hidden if the batch is not the cut-off directly in the panel

PR #8694, opened 2022-08-01
RM#52872: MOVE: corrected not displaying simulate button in move form

jdo-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -428,6 +431,12 @@
       expr="eval: journal?.journalType?.technicalTypeSelect"/>
   </action-record>
 
+  <action-record name="action-move-record-set-is-activate-simulated-moves"
+    model="com.axelor.apps.account.db.Move">
+    <field name="$isActivateSimulatedMoves"
+      expr="eval:__user__.activeCompany?.accountConfig?.isActivateSimulatedMove"/>
```
gdu-axelor, line 437: Should depends of move.journal.authorizeSimulatedMove

Also, better to use an action-attrs instead of action-record and showIf condition

gdu-axelor (changes requested):

PR #8706, opened 2022-08-02
RM#52297: JOURNAL: corrected fields not read only in journal although…

jdo-axelor:
… moves are linked

### DIFF
--- a/axelor-account/src/main/resources/views/Journal.xml
+++ b/axelor-account/src/main/resources/views/Journal.xml
@@ -18,24 +18,24 @@
   </grid>
 
   <form name="journal-form" title="Journal" model="com.axelor.apps.account.db.Journal"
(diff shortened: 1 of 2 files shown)


axelor-account/src/main/resources/views/Journal.xml
```diff
@@ -18,7 +18,7 @@
   </grid>
 
   <form name="journal-form" title="Journal" model="com.axelor.apps.account.db.Journal"
-    onNew="account-journal-record-default,action-method-account-journal-readonly-if-journal-in-move"
+    onNew="account-journal-record-default,action-method-account-journal-readonly-if-journal-in-move,action-attrs-account-journal-readonly-if-account-in-move-line"
```
aol-axelor, line 21: Please create an `action-group` for better maintainability

jmi-axelor (approved):

PR #8719, opened 2022-08-04
RM#53116: corrected behavior of ducplicate dectection origin if partn…

jdo-axelor:
…er is null, now display other move references and change title of move origin field


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveToolService.java
```diff
@@ -191,7 +191,7 @@ boolean isTemporarilyClosurePeriodManage(Period period, Journal journal, User us
 
   boolean checkMoveLinesCutOffDates(Move move);
 
-  boolean checkMoveOriginIsDuplicated(Move move) throws AxelorException;
+  List<Move> checkMoveOriginIsDuplicated(Move move) throws AxelorException;
```
aol-axelor, line 194: Maybe the method name should be changed as well, to something like `getMovesWithDuplicatedOrigin`

PR #8798, opened 2022-08-22
RM#53556 : Fix payment session global computation

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/PaymentModeServiceImpl.java
```diff
@@ -181,11 +181,8 @@ public Account getPaymentModeAccount(
     throw new AxelorException(
         paymentMode,
         TraceBackRepository.CATEGORY_CONFIGURATION_ERROR,
-        I18n.get("Company")
-            + " : %s, "
-            + I18n.get("Payment mode")
-            + " : %s: "
-            + I18n.get(IExceptionMessage.PAYMENT_MODE_1),
+        I18n.get(IExceptionMessage.PAYMENT_MODE_CASH_ACCOUNT),
+        global ? I18n.get("global ") : "",
```
jmi-axelor, line 185: I think it's better to do something like: I18n.get("global") + " "
Because the I18n.get is doing a trim so the space at the end will be removed

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #8810, opened 2022-08-23
RM#53339 Rewrite BankStatementCreateService#computeName method

alo-axelor:


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankstatement/BankStatementCreateService.java
```diff
@@ -52,37 +52,30 @@ public BankStatement createBankStatement(
     return bankStatement;
   }
 
-  protected String computeName(BankStatement bankStatement) {
-
-    String name = "";
+  public String computeName(BankStatement bankStatement) {
```
kra-axelor, line 55: StringBuilder is good, but I think it is more appropriate to use StringJoiner to avoid if statement such as `   if (!builder.toString().isEmpty()) {
        builder.append("-");
      }`

kra-axelor (changes requested):

PR #8823, opened 2022-08-25
Anomaly RM#53691:ANALYTIC RULES _ Possibility to select an analytic a…

vva-axelor:
…ccount that is not linked to the same company

### DIFF
--- a/axelor-account/src/main/resources/views/AnalyticRules.xml
+++ b/axelor-account/src/main/resources/views/AnalyticRules.xml
@@ -22,7 +22,7 @@
       <field name="toAccount" onSelect="action-on-select-to-account-set-domain"
         requiredIf="company != null" readonlyIf="company == null"/>
       <field name="analyticAccountSet" colSpan="12" readonlyIf="company == null"
-        domain="(self.analyticAxis.company = :company OR self.analyticAxis.company IS NULL OR self.company = :company OR self.company IS NULL) AND self.statusSelect = 1"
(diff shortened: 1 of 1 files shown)


axelor-account/src/main/resources/views/AnalyticRules.xml
```diff
@@ -22,7 +22,7 @@
       <field name="toAccount" onSelect="action-on-select-to-account-set-domain"
         requiredIf="company != null" readonlyIf="company == null"/>
       <field name="analyticAccountSet" colSpan="12" readonlyIf="company == null"
-        domain="(self.analyticAxis.company = :company OR self.analyticAxis.company IS NULL OR self.company = :company OR self.company IS NULL) AND self.statusSelect = 1"
+        domain="(self.analyticAxis.company = :company OR self.analyticAxis.company IS NULL AND self.company = :company OR self.company IS NULL) AND self.statusSelect = 1"
```
jmi-axelor, line 25: Coudl you add some brackets? 
Because when I try to understand the behavior, I can't, I don't get if you want to do :
`((self.analyticAxis.company = :company OR self.analyticAxis.company IS NULL) AND (self.company = :company OR self.company IS NULL))`
or
`(self.analyticAxis.company = :company OR (self.analyticAxis.company IS NULL AND self.company = :company) OR self.company IS NULL)`
or something else ...

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #8824, opened 2022-08-25
Anomaly #53702: manage first service date null

kra-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/AbstractFixedAssetLineComputationServiceImpl.java
```diff
@@ -485,4 +488,12 @@ protected BigDecimal getAccountingValue(FixedAssetLine fixedAssetLine) {
         ? fixedAssetLine.getCorrectedAccountingValue()
         : fixedAssetLine.getAccountingValue();
   }
+
+  protected LocalDate getFirstServiceDate(FixedAsset fixedAsset) {
```
jmi-axelor, line 492: This method is never used, maybe use it everywhere you do a :
```
    LocalDate firstServiceDate =
        fixedAsset.getFirstServiceDate() == null
            ? fixedAsset.getAcquisitionDate()
            : fixedAsset.getFirstServiceDate();
```
And set it as a public static method, so you don't have to inject the service all the time? 
Or just delete this method

jmi-axelor (changes requested):
Also could you add a changelog file?

kra-axelor:
> Also could you add a changelog file?

I don't think it needs a changelog as firstServiceDate has been added in 6.3 and 6.3 is not released yet

jmi-axelor (approved):

PR #8835, opened 2022-08-26
Anomaly RM#53732:MOVE / ANALYTIC DISTRIBUTION _ Possibily to select a…

vva-axelor:
…n analytic account that is not linked to the same company

### DIFF
--- a/axelor-account/src/main/resources/views/AnalyticMoveLine.xml
+++ b/axelor-account/src/main/resources/views/AnalyticMoveLine.xml
@@ -253,11 +253,14 @@
 
   <action-attrs name="action-attrs-domain-onselect-analytic-account">
     <attribute name="domain" for="analyticAccount"
-      expr="eval: &quot;self.statusSelect = 1 AND self.analyticAxis.id = ${analyticAxis.id}  AND self.company.id = ${analyticAxis.company.id}&quot;"
+      expr="eval: &quot;self.statusSelect = 1 AND self.analyticAxis.id = ${analyticAxis?.id} AND self.company.id = ${analyticAxis?.company?.id}&quot;"
       if="analyticAxis &amp;&amp; analyticAxis.company"/>
     <attribute name="domain" for="analyticAccount"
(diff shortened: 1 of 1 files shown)


axelor-account/src/main/resources/views/AnalyticMoveLine.xml
```diff
@@ -242,11 +242,14 @@
 
   <action-attrs name="action-attrs-domain-onselect-analytic-account">
     <attribute name="domain" for="analyticAccount"
-      expr="eval: &quot;self.statusSelect = 1 AND self.analyticAxis.id = ${analyticAxis.id}  AND self.company.id = ${analyticAxis.company.id}&quot;"
+      expr="eval: &quot;self.statusSelect = 1 AND (self.analyticAxis.id = ${analyticAxis?.id} AND (self.company IS NULL OR self.company.id = ${analyticAxis?.company?.id}))&quot;"
```
jmi-axelor, line 245: We don't want to get analyticAccounts with a null company, but here we'll get them with: self.company IS NULL
Please remove it

axelor-account/src/main/resources/views/AnalyticMoveLine.xml
```diff
@@ -242,11 +242,14 @@
...
       if="analyticAxis &amp;&amp; analyticAxis.company"/>
     <attribute name="domain" for="analyticAccount"
-      expr="eval: &quot;self.statusSelect = 1 AND self.analyticAxis.id = ${analyticAxis.id}&quot;"
+      expr="eval: &quot;self.statusSelect = 1 AND (self.analyticAxis.id = ${analyticAxis?.id} AND (self.company IS NULL OR self.company.id = ${__parent__?.parentContext?.company?.id?:__parent__?.move?.company?.id}))&quot;"
```
jmi-axelor, line 248: Same here, please remove the "self.company IS NULL"

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #8843, opened 2022-08-26
53520 : improve methode accountingMultipleMoves

alg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/MoveController.java
```diff
@@ -145,30 +145,15 @@ public void generateReverse(ActionRequest request, ActionResponse response) {
 
   @SuppressWarnings("unchecked")
   public void validateMultipleMoves(ActionRequest request, ActionResponse response) {
-    List<Long> moveIds = (List<Long>) request.getContext().get("_ids");
+    List<Integer> moveIds = (List<Integer>) request.getContext().get("_ids");
```
jmi-axelor, line 148: Why did you change it from List<Long> to List<Integer>?

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/web/MoveController.java
alg-axelor: Because the list comes from the context, and somehow, if I keep the List<Long>, it throws a cast exception, and in the debugger, the List<Long> contains integers. This solution is used in a lot of methods in the app.

jmi-axelor (approved):

PR #8931, opened 2022-09-09
54144 management of error messages in partner when bank detail is inv…

rth-axelor:
…alid


axelor-account/src/main/resources/views/Partner.xml
```diff
@@ -45,9 +45,8 @@
                 ]]></viewer>
         <editor onNew="action-set-default-partner" x-show-on-new="false">
           <field name="ownerName"/>
-          <field name="iban" onChange="action-group-account-bankdetails-iban-onchange"/>
-          <field name="bank" onChange="action-group-account-bankdetails-iban-onchange"
```
jmi-axelor, line 49: Code is ok but I'm not sure we want to remove the onchange on these 2 fields

PR #8985, opened 2022-09-16
52041 - ANALYTIC DITRIBUTION - add filter for analyticAxis given acco…

alg-axelor:
…untConfig of the company


axelor-account/src/main/resources/views/AnalyticMoveLine.xml
```diff
@@ -126,9 +126,11 @@
     <panel name="mainPanel">
       <field name="analyticJournal" grid-view="analytic-journal-grid"
         form-view="analytic-journal-form"
-        onSelect="action-attrs-domain-onselect-journal-analytic-move-line"/>
+        onSelect="action-attrs-domain-onselect-journal-analytic-move-line"
+        onChange="action-method-domain-onselect-axis-analytic-move-line"/>
```
jmi-axelor, line 130: I think it's useless to call an action that will change a domain of another field on an onChange. 
The same action will be called anyway when we try to select something for this other field.

axelor-account/src/main/resources/views/AnalyticMoveLine.xml
```diff
@@ -126,9 +126,11 @@
...
       <field name="analyticAxis" form-view="analytic-axis-form" grid-view="analytic-axis-grid"
-        onSelect="action-attrs-domain-onselect-axis-analytic-move-line"/>
+        onChange="action-attrs-domain-onselect-journal-analytic-move-line"
```
jmi-axelor, line 132: I think it's useless to call an action that will change a domain of another field on an onChange. 
The same action will be called anyway when we try to select something for this other field.

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9021, opened 2022-09-22
Anomaly #54734: Fixed not persisted Stock history line

kra-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockHistoryServiceImpl.java
```diff
@@ -55,18 +56,31 @@ public class StockHistoryServiceImpl implements StockHistoryService {
...
+      StockLocationRepository stockLocationRepository,
+      StockHistoryLineManagementRepository stockHistoryLineRepository) {
     this.stockMoveLineRepository = stockMoveLineRepository;
     this.unitConversionService = unitConversionService;
     this.stockLocationRepository = stockLocationRepository;
+    this.stockHistoryLineRepository = stockHistoryLineRepository;
   }
 
+  @Override
   @Transactional
```
ale-axelor, line 74: Please update transactional annotation to rollback on Exception

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockHistoryServiceImpl.java
```diff
@@ -55,18 +56,31 @@ public class StockHistoryServiceImpl implements StockHistoryService {
...
+  public List<StockHistoryLine> computeAndSaveStockHistoryLineList(
+      Long productId, Long companyId, Long stockLocationId, LocalDate beginDate, LocalDate endDate)
+      throws AxelorException {
+
+    return stockHistoryLineRepository.save(
+        this.computeStockHistoryLineList(
```
ale-axelor, line 80: Please define a new method `createStockHistoryLine` (that will include the call to `new StockHistoryLine();`). It could also improve the whole PR, by allowing to call the save() directly in this method/after this method.

ale-axelor (changes requested):

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockHistoryServiceImpl.java
kra-axelor: I am not sure about this change, `computeStockHistoryLineList` is also used in another context where we don't want to persist the StockHistoryLines (which is the reason why I made another method). Also the method also creates specifics StockHistoryLine (which can not be included in a loop, and are filled differently)

PR #9024, opened 2022-09-22
RM#53865 Add null checks for activeCompany

alo-axelor:


axelor-crm/src/main/java/com/axelor/apps/crm/message/MessageServiceCrmImpl.java
```diff
@@ -59,38 +60,29 @@ public Message createMessage(Event event)
...
+    Company company = event.getUser().getActiveCompany();
 
-    switch (event.getTypeSelect()) {
-      case EventRepository.TYPE_EVENT:
-        template =
-            Beans.get(CrmConfigService.class)
-                .getCrmConfig(event.getUser().getActiveCompany())
-                .getEventTemplate();
-        break;
+    if (company != null) {
```
ale-axelor, line 65: I think this method must get a company, it looks like the parent method will throw a NPE if a message was not generated. So please throw an exception if the company is not found.

ale-axelor (changes requested):

axelor-crm/src/main/java/com/axelor/apps/crm/message/MessageServiceCrmImpl.java
```diff
@@ -59,38 +63,33 @@ public Message createMessage(Event event)
...
 
-      default:
-        break;
+        default:
+          break;
+      }
+    } else {
+      throw new AxelorException(
+          TraceBackRepository.CATEGORY_CONFIGURATION_ERROR,
+          I18n.get(BaseExceptionMessage.NO_ACTIVE_COMPANY));
```
ale-axelor, line 92: Please change the message on this method to "Please set an active company for user %s" and set the name of the user set in the event

PR #9072, opened 2022-10-04
RM#55013 BankOrder : Introduce booleans to avoid duplicating move generation or bank communication 

mga-axelor:
Contains a lot of refactors in different commits, do not squash commits


ale-axelor:
see #9099

PR #9081, opened 2022-10-06
54991 - SEQUENCE : implement a way to throw an understandable excepti…

alg-axelor:
…on when the generated sequence value is duplicated


axelor-base/src/main/java/com/axelor/apps/base/service/administration/SequenceService.java
```diff
@@ -182,13 +190,40 @@ public String getSequenceNumber(Sequence sequence, LocalDate refDate) {
...
     return nextSeq;
   }
 
+  protected void isSequenceAlreadyExisting(Class objectClass, String fieldName, String nextSeq)
+      throws AxelorException {
+    String table = objectClass.getSimpleName();
+    boolean isSequenceAlreadyExisting =
+        CollectionUtils.isNotEmpty(
+            JPA.em()
+                .createQuery(
```
jmi-axelor, line 211: You use createQuery so fieldName should be the field's Java name and not SQL name. 
Example, in SyncContactService, you set fieldName to "partner_seq", it should be "partnerSeq"

axelor-base/src/main/java/com/axelor/apps/base/service/administration/SequenceService.java
```diff
@@ -182,13 +190,40 @@ public String getSequenceNumber(Sequence sequence, LocalDate refDate) {
...
+                .setParameter("nextSeq", nextSeq)
+                .setLockMode(LockModeType.PESSIMISTIC_WRITE)
+                .setFlushMode(FlushModeType.COMMIT)
+                .getResultList());
+    if (isSequenceAlreadyExisting) {
+      throw new AxelorException(
+          TraceBackRepository.CATEGORY_CONFIGURATION_ERROR,
+          I18n.get(BaseExceptionMessage.SEQUENCE_ALREADY_EXISTS),
+          nextSeq,
+          table);
```
jmi-axelor, line 223: We can have multiple Sequences for a Move, so please display the name of the Sequence instead of the table.

jmi-axelor (changes requested):
I'll change the Redmine issue from an anomaly to a feature, so please add a changelog file in your PR

jmi-axelor (approved):

PR #9098, opened 2022-10-10
RM#55221 Production API : Add subManufOrderId of consumed products in…

alo-axelor:
… manufOrderResponse


ale-axelor (changes requested):

axelor-production/src/main/java/com/axelor/apps/production/rest/ManufOrderProductRestServiceImpl.java
```diff
@@ -228,4 +242,16 @@ public void updateProdProductQty(ProdProduct prodProduct, BigDecimal qty) {
       prodProduct.setQty(qty);
     }
   }
+
+  protected ManufOrder getChildManufOrder(List<ManufOrder> childrenManufOrder, Product product) {
```
ale-axelor, line 246: Please simplify this method, you do not have to check if the list childrenManufOrder is empty because the stream will work on empty stream (and will simply return an empty optional, that you are converting to `null`

ale-axelor (changes requested):

PR #9125, opened 2022-10-14
55541 - RECONCILIATION : unreconciliation process after validating

alg-axelor:


axelor-bank-payment/src/main/resources/views/BankReconciliationLine.xml
```diff
@@ -6,9 +6,9 @@
   <grid name="bank-reconciliation-line-grid" title="Bank reconciliation Lines"
     model="com.axelor.apps.bankpayment.db.BankReconciliationLine" canDelete="false" editable="false"
     canNew="false">
-    <button onClick="action-bank-reconciliation-line-method-set-selected" name="select"
+    <button onClick="save,action-bank-reconciliation-line-method-set-selected" name="select"
       title="Select / Unselect" icon="fa-check-circle" showIf="!isSelectedBankReconciliation"/>
-    <button onClick="action-bank-reconciliation-line-method-set-selected" name="select"
+    <button onClick="save,action-bank-reconciliation-line-method-set-selected" name="select"
```
gdu-axelor, line 11: Why a save is introduce here ?

axelor-bank-payment/src/main/resources/views/BankReconciliation.xml
```diff
@@ -144,6 +147,17 @@
         title="Computed balance in accounting and Ending Balance of the statement should be equal to validate the Statement"
         showIf="computedBalance != movesTheoreticalBalance" colSpan="12" css="label-danger"/>
     </panel>
+    <panel colSpan="12" showIf="corrected" title="Correction">
+      <field name="comments"
+        showIf="statusSelect == 3 || statusSelect == 2 &amp;&amp; comments != null"
```
gdu-axelor, line 152: On the status "Under correction", the field "comments" should be displayed even if empty

axelor-bank-payment/src/main/resources/domains/BankReconciliation.xml
```diff
@@ -35,6 +35,10 @@
     <many-to-one name="validatedByUser" title="Validated By"
       ref="com.axelor.auth.db.User" readonly="true"/>
     <date name="validatedDate" title="Validated On" readonly="true"/>
+    <boolean name="corrected" title="Corrected" default="false"/>
+    <date name="correctedDate" readonly="true"/>
```
gdu-axelor, line 39: Better to define a dateTime field.

axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/web/BankReconciliationController.java
```diff
@@ -490,4 +490,43 @@ public void computeTotalOfSelectedMoveLines(ActionRequest request, ActionRespons
...
+
+  public void correctedLabelFill(ActionRequest request, ActionResponse response) {
+    try {
+      BankReconciliation bankReconciliation = request.getContext().asType(BankReconciliation.class);
+      if (bankReconciliation.getCorrected()) {
+        response.setAttr(
+            "correctedLabel",
+            "title",
+            Beans.get(BankReconciliationService.class)
+                .getCorrectedLabel(bankReconciliation.getCorrectedDate(), request.getUser()));
```
gdu-axelor, line 514: Not sure about the way to get the user. The user to take into account is the bankReconciliation.correctedUser not the current user.

gdu-axelor (changes requested):

axelor-bank-payment/src/main/resources/views/BankReconciliationLine.xml
alg-axelor: because the correct button does not save the reconciliation as the comments field is mandatory. So when clicking on correct and then clicking on select, the reconciliation return into validated status

jmi-axelor (approved):

PR #9141, opened 2022-10-18
55542 - RECONCILIATION : dynamically update ongoing balance - the amo…

alg-axelor:
…unts are wrong when the lines don't match

### DIFF
--- a/axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationService.java
+++ b/axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationService.java
@@ -523,25 +523,34 @@ public class BankReconciliationService {
   protected BigDecimal computeStatementOngoingReconciledLineBalance(
       BigDecimal statementOngoingReconciledBalance, BankReconciliationLine brl) {
     if (!brl.getIsPosted() && !Strings.isNullOrEmpty(brl.getPostedNbr())) {
-      statementOngoingReconciledBalance =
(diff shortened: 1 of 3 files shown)


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationService.java
```diff
@@ -523,25 +523,33 @@ protected BigDecimal computeStatementUnreconciledLineBalance(
   protected BigDecimal computeStatementOngoingReconciledLineBalance(
       BigDecimal statementOngoingReconciledBalance, BankReconciliationLine brl) {
     if (!brl.getIsPosted() && !Strings.isNullOrEmpty(brl.getPostedNbr())) {
-      statementOngoingReconciledBalance =
-          statementOngoingReconciledBalance.subtract(brl.getDebit());
-      statementOngoingReconciledBalance = statementOngoingReconciledBalance.add(brl.getCredit());
+      String query = "self.moveLine = %s";
+      query = query.replace("%s", brl.getMoveLine().getId().toString());
+      List<BankReconciliationLine> bankReconciliationLines =
+          bankReconciliationLineRepository.all().filter(query).fetch();
```
jmi-axelor, line 529: You can just write:
bankReconciliationLineRepository.all().filter("self.moveLine.id = ?1", brl.getMoveLine().getId())

And it's strange that "self.moveLine = id" doesn't throw an exception

jmi-axelor (changes requested):

PR #9272, opened 2022-11-14
57066 - ANNUAL CLOSURE _ Wrong batch action to execute

alg-axelor:


axelor-supplychain/src/main/resources/views/ClosureAssistantLine.xml
```diff
@@ -1,26 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<object-views xmlns="http://axelor.com/xml/ns/object-views"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://axelor.com/xml/ns/object-views http://axelor.com/xml/ns/object-views/object-views_5.4.xsd">
-
-
-  <action-group
-    id="supplychain-account-closure-assistant-line-open-view-link-to-action-group"
-    name="account-closure-assistant-line-open-view-link-to-action-group">
```
jmi-axelor, line 9: Can you create an SQL script in abs-script to delete this action please? Otherwise it will not be deleted automatically and we'll still have this override, so the action supplychain-account-closure-assistant-line-open-view-link-to-action-method will still be called.

jmi-axelor (changes requested):

PR #9309, opened 2022-11-21
Anomaly #57335: Fixed double reload invoice ventilate or validate or …

kra-axelor:
…change invoice line list


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceFinancialDiscountServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+      resetFinancialDiscountInformations(invoice);
+    }
+  }
+
+  protected BigDecimal computeFinancialDiscountTotalAmount(
+      Invoice invoice, FinancialDiscount financialDiscount) {
+    return financialDiscount
+        .getDiscountRate()
+        .multiply(invoice.getInTaxTotal())
+        .divide(new BigDecimal(100));
```
ale-axelor, line 46: It is not a huge issue especially in the current code, but it might be better to set a scale here (equal to the scale of the field `financialDiscountTotalAmount`

ale-axelor (changes requested):

PR #9313, opened 2022-11-22
ContractLine : prorata and discount per line

erg-axelor:


axelor-contract/src/main/resources/domains/Invoice.xml
```diff
@@ -7,7 +7,7 @@
 
   <entity name="Invoice" lang="java" cacheable="true">
 
-    <many-to-one name="contract" title="Contract"
+    <many-to-many name="contract" title="Contract"
```
ale-axelor, line 10: Please change the name to contractSet and the title

axelor-contract/src/main/resources/views/Contract.xml
```diff
@@ -64,7 +67,8 @@
             <editor x-viewer="true" x-show-on-new="true" x-show-titles="true">
               <panel colSpan="6">
                 <panel hideIf="statusSelect >= 3" colSpan="12">
-                  <field name="supposedActivationDate"/>
+                  <field name="supposedActivationDate"
+                    onChange="action-method-contract-version-check-supposed-activation-date,save,action-method-contract-version-update-contract-lines"/>
```
ale-axelor, line 71: please check this, we should only save, then call a single action method, not call a second action method.

ale-axelor (changes requested):

PR #9338, opened 2022-11-24
57532 Adding class variable for JPA.clear occurence limit

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateServiceImpl.java
```diff
@@ -76,7 +76,7 @@
 @RequestScoped
 public class MoveValidateServiceImpl implements MoveValidateService {
   private final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-
+  private final int jpaLimit = 20;
```
jmi-axelor, line 79: It must be protected if we want to be able to override it in another class

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
```diff
@@ -65,6 +65,7 @@
 public class MoveLineServiceImpl implements MoveLineService {
 
   private final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  private final int jpaLimit = 20;
```
jmi-axelor, line 68: It must be protected if we want to be able to override it in another class

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9342, opened 2022-11-25
57515 - Fix creation of new invoiceTerm without invoiceId

paa-axelor:


jmi-axelor (approved):

axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -659,7 +659,7 @@
       <panel name="invoiceTermFromInvoicePanel" title="Term(s)"
         showIf="account.hasInvoiceTerm  &amp;&amp; move.invoice != null" colSpan="12">
         <button name="showInvoiceTermsFromInvoiceBtn" title="Show related invoice term(s)"
-          onClick="save,action-move-line-view-show-related-invoice-terms-from-invoice"
+          onClick="save,action-move-line-view-show-related-invoice-terms-from-invoice,close"
```
ale-axelor, line 662: This change does not cause issue when we call this action from move line menu form ?

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -138,6 +137,15 @@ public void initInvoiceTerm(ActionRequest request, ActionResponse response) {
...
         }
+      } else if (request.getContext().get("_invoiceId") != null) {
+        invoice =
+            Beans.get(InvoiceRepository.class)
+                .find(Long.valueOf((Integer) request.getContext().get("_invoiceId")));
+      }
+
+      if (invoice != null) {
+        invoiceTermService.initCustomizedInvoiceTerm(invoice, invoiceTerm);
+        response.setValues(invoiceTerm);
```
ale-axelor, line 148: Please refactor the whole `initInvoiceTerm` method as this is becoming too complex.
Please change it so we avoid nested `if` conditions: first retrieve invoice/move, then depending on which variable is null, call the right service.
Also I don't get why we call `moveLine.setMove(move);` so please check if this is really necessary.

ale-axelor (changes requested):

axelor-account/src/main/resources/views/MoveLine.xml
paa-axelor: Create an issue on move-line-move-fom and can create confusion for user, reverted

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
paa-axelor: initInvoiceTerm method refactored and remotion of `moveLine.setMove(move);`

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceTermController.java
```diff
@@ -119,25 +121,22 @@ public void initInvoiceTerm(ActionRequest request, ActionResponse response) {
...
-              moveLine.setMove(move);
-            }
-          }
-        }
+      if (invoice == null && moveLine != null) {
+        Move move = ContextTool.getContextParent(request.getContext(), Move.class, 2);
+        invoiceTermService.initCustomizedInvoiceTerm(moveLine, invoiceTerm, move);
+      } else if (invoice != null) {
+        invoiceTermService.initCustomizedInvoiceTerm(invoice, invoiceTerm);
+        response.setValues(invoiceTerm);
```
ale-axelor, line 139: I think you can remove the setValues() here, since it will always be called at the end

paa-axelor: Solved

PR #9355, opened 2022-11-30
Anomaly RM#57735 [Accounting] missing column on accounting reports

pda-axelor:


ale-axelor:
#9367 was correctly merged into wip branch, so this can be closed.

PR #9359, opened 2022-12-01
RM#57837: Invoice payment corrected discount fields not displayed in …

jdo-axelor:
…invoice payment


axelor-account/src/main/resources/views/InvoicePayment.xml
```diff
@@ -61,9 +61,9 @@
     <panel name="discountPanel" title="Financial discount" showIf="applyFinancialDiscount"
       if="__config__.app.getApp('account')?.getManageFinancialDiscount()">
       <field name="applyFinancialDiscount" readonly="true"/>
-      <field name="$totalAmountWithFinancialDiscount" type="decimal"
+      <field name="totalAmountWithFinancialDiscount" type="decimal"
```
ale-axelor, line 64: Please remove the `type` attribute since this is not a dummy field anymore

jdo-axelor: Thanks it's corrected

PR #9426, opened 2022-12-08
58187 - Fix filter by company on journal field

paa-axelor:


axelor-account/src/main/resources/views/AccountManagement.xml
```diff
@@ -184,4 +186,8 @@
       if="company?.defaultBankDetails?.active"/>
   </action-attrs>
 
+  <action-record name="action-account-management-fill-company"
+    model="com.axelor.apps.account.db.AccountManagement">
+    <field name="company" expr="eval: journal.company" if="journal != null"/>
```
jmi-axelor, line 191: You can use "?" in a groovy expression to avoid NPE, so the `if="journal != null"` is useless if you do:
`<field name="company" expr="eval: journal?.company"/>`

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9499, opened 2022-12-19
RM#57659 Fix invoiceState when saleOrder total amount was zero

alo-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderInvoiceServiceImpl.java
```diff
@@ -953,6 +953,34 @@ public void displayErrorMessageBtnGenerateInvoice(SaleOrder saleOrder) throws Ax
     }
   }
 
+  @Override
+  public int getInvoicingState(SaleOrder saleOrder) {
+    int invoicingState = 0;
```
ale-axelor, line 958: Please use static variables like we do for selections instead of hardcoded numbers

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SaleOrderInvoiceServiceImpl.java
```diff
@@ -953,6 +953,34 @@ public void displayErrorMessageBtnGenerateInvoice(SaleOrder saleOrder) throws Ax
...
+      } else {
+        invoicingState = 1;
+      }
+    }
+
+    return invoicingState;
+  }
+
+  protected boolean atLeastOneInvoiceIsVentilated(SaleOrder saleOrder) {
+    return getInvoices(saleOrder).stream().anyMatch(invoice -> invoice.getStatusSelect() == 3);
```
ale-axelor, line 981: Please change this to a `.count()` query and use static variable from InvoiceRepository instead of writing `3`

ale-axelor (changes requested):

PR #9500, opened 2022-12-19
58494 - Set default value when we create a new fixedAsset

paa-axelor:


axelor-account/src/main/resources/domains/FixedAsset.xml
```diff
@@ -123,9 +123,9 @@
       name="associatedFixedAssetsSet" title="Associated Fixed Assets Set"/>
     <string name="comments" large="true" multiline="true" title="Comments"/>
     <integer name="disposalTypeSelect" title="Disposal Type"
-      selection="account.fixed.asset.disposal.type.select" default="1"/>
+      selection="account.fixed.asset.disposal.type.select" default="0"/>
     <integer name="disposalQtySelect" title="Disposal Quantity select"
-      selection="account.fixed.asset.disposal.quantity.select" default="1"/>
+      selection="account.fixed.asset.disposal.quantity.select" default="0"/>
```
jmi-axelor, line 128: You just need to remove the default="1", integers' default value is already 0

axelor-account/src/main/resources/views/FixedAsset.xml
```diff
@@ -434,6 +434,8 @@
     <field name="firstServiceDate" expr="eval: __config__.date"/>
     <field name="firstDepreciationDate" expr="eval: __config__.date"/>
     <field name="originSelect" expr="eval:'1'"/>
+    <field name="disposalTypeSelect" expr="eval: 0"/>
+    <field name="disposalQtySelect" expr="eval: 0"/>
```
jmi-axelor, line 438: You don't need to add this, integer's default value is already 0
But we need an SQL script to change the default value in the database

jmi-axelor (changes requested):

axelor-account/src/main/resources/views/FixedAsset.xml
paa-axelor: Fixed

jmi-axelor (approved):

PR #9502, opened 2022-12-20
RM58228 Move template : add partner on the generated move

erg-axelor:


jmi-axelor (changes requested):
Please add a changelog file

jmi-axelor (approved):

PR #9505, opened 2022-12-20
RM57783 Blank partner on InvoiceTerm view form

erg-axelor:


jmi-axelor (changes requested):
Please add a changelog file

PR #9524, opened 2022-12-21
RM#58610 Change Generate Proposal wizard for MRP

alo-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/MrpServiceImpl.java
```diff
@@ -1504,29 +1504,86 @@ protected void copyMrpLineOrigins(MrpLine mrpLine, List<MrpLineOrigin> mrpLineOr
...
-      if (!mrpLine.getProposalGenerated()) {
-        mrpLineService.generateProposal(
-            mrpLine, purchaseOrders, purchaseOrdersPerSupplier, isProposalPerSupplier);
+  protected List<MrpLine> getAllMrpLines(Mrp mrp) throws AxelorException {
+    List<MrpLine> mrpLineList =
+        mrpLineRepository
+            .all()
+            .filter("self.mrp.id = :id AND self.proposalGenerated = false")
+            .bind("id", mrp.getId())
+            .bind("elementSelect", MrpLineTypeRepository.ELEMENT_PURCHASE_PROPOSAL)
```
ale-axelor, line 1551: this bind is not used in the filter.
I'm not sure about this method. Most MRP have a lot of lines, here we fetch all lines and not only the proposals, which can cause huge performance issues. We should at least filter to get only proposal lines, maybe manage JPA cache.

ale-axelor (changes requested):

PR #9525, opened 2022-12-21
Anomaly #55469: Fixed BOM priority

kra-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -449,6 +449,49 @@ public List<BillOfMaterial> getAlternativesBOM(Product originalProduct, Company
...
+            BillOfMaterialRepository.STATUS_APPLICABLE)
+        .order("id")
+        .fetchOne();
+  }
+
+  @Override
+  public BillOfMaterial getBOM(Product originalProduct, Company company) throws AxelorException {
+
+    if (company == null) {
+      company = AuthUtils.getUser().getActiveCompany();
```
ale-axelor, line 468: be careful, AuthUtils.getUser() can return null, and the activeCompany on an user can also be null.

ale-axelor (changes requested):

axelor-base/src/main/java/com/axelor/apps/base/service/ProductCompanyServiceImpl.java
```diff
@@ -99,4 +113,49 @@ private Product findAppropriateProductCompany(
...
+    }
+
+    Product product = null;
+
+    if (company != null && originalProduct.getProductCompanyList() != null) {
+      for (ProductCompany productCompany : originalProduct.getProductCompanyList()) {
+        if (company.equals(productCompany.getCompany())) {
+          Set<MetaField> companySpecificFields =
+              appBaseService.getAppBase().getCompanySpecificProductFieldsSet();
+          for (MetaField field : companySpecificFields) {
```
ale-axelor, line 148: Why do you need to do this check inside the loop ? fieldName is available at the start of the method. Please rewrite this method as it is not optimized at all and is difficult to understand

PR #9554, opened 2022-12-26
58723 - Fix deletion of analytic axis on moveline and remove popup-sa…

paa-axelor:
…ve aparition when we try to close move-line-move-from from move without modification


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -225,12 +225,14 @@
         colSpan="12" height="6"
         showIf="(statusSelect == 1 || statusSelect == 2 || statusSelect == 5) &amp;&amp; journal &amp;&amp; currency &amp;&amp; company &amp;&amp; date &amp;&amp; period"
         canNew="true" canRemove="true" grid-view="move-line-move-grid"
-        form-view="move-line-move-form" onChange="action-move-group-move-line-list-onchange"/>
+        form-view="move-line-move-form" onChange="action-move-group-move-line-list-onchange"
+        x-call-onSave="false"/>
```
jmi-axelor, line 229: I don't understand why we need this `x-call-onSave="false"`, there is no onSave in move-line-move-form

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9567, opened 2022-12-30
RM#58364 New config for new invoice statements

alo-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -741,6 +741,10 @@ InvoiceSet.length > 0))"
           <field name="legalNotice" requiredIf="financialDiscount" widget="html" colSpan="12"
             showTitle="false"/>
         </panel>
+        <panel name="invoiceProductStatementPanel" title="Invoice Product Statement"
+          colSpan="12" if="__user__.activeCompany?.accountConfig?.displayItemsCategoriesOnPrinting">
```
ale-axelor, line 745: we should check the company from the invoice here, not from the user, so please use company from the invoice and use a `showIf` condition instead of `if`

axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -2692,4 +2698,15 @@ InvoiceSet.length > 0))"
       method="showSupplierInvoiceLines"/>
   </action-method>
 
+  <action-method name="action-invoice-method-get-invoice-product-statement">
+    <call class="com.axelor.apps.account.web.InvoiceController"
+      method="getInvoiceProductStatement"/>
+  </action-method>
```
ale-axelor, line 2704: I disagree with this implementation, please remove the action method call where it is not necessary (actually I think the action-method can be completely removed) and instead make the change in java code so existing services will call the new InvoiceProductStatementService.

ale-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/print/InvoiceProductStatementServiceImpl.java
```diff
@@ -0,0 +1,121 @@
...
+          getInvoiceProductStatements(invoice);
+      return computeStatement(productTypes, invoiceProductStatementList);
+    }
+    return "";
+  }
+
+  protected Set<InvoiceProductStatement> getInvoiceProductStatements(Invoice invoice) {
+    AccountConfig accountConfig = accountConfigRepository.findByCompany(invoice.getCompany());
+    return accountConfig.getStatementsForItemsCategoriesSet();
+  }
```
ale-axelor, line 52: please accountConfigService instead

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/print/InvoiceProductStatementServiceImpl.java
```diff
@@ -0,0 +1,121 @@
...
+  protected AccountConfigRepository accountConfigRepository;
+
+  @Inject
+  public InvoiceProductStatementServiceImpl(AccountConfigRepository accountConfigRepository) {
+    this.accountConfigRepository = accountConfigRepository;
+  }
+
+  @Override
+  public String getInvoiceProductStatement(Invoice invoice) {
+    AccountConfig accountConfig = accountConfigRepository.findByCompany(invoice.getCompany());
```
ale-axelor, line 28: please use accountConfigService instead

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/print/InvoiceProductStatementServiceImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    return statement;
+  }
+
+  protected String getStatementForOneType(
+      Set<String> productTypes, Set<InvoiceProductStatement> invoiceProductStatementList) {
+    if (productTypes.size() != 1) {
+      return "";
+    }
+    int min = getTypeListMinSize(invoiceProductStatementList);
+    List<String> statementSet = new ArrayList<>();
```
ale-axelor, line 84: this variable should be called `statementList`

ale-axelor (changes requested):

PR #9629, opened 2023-01-11
RM#59169: corrected accounting report 2055 wrong column A

jdo-axelor:


jdo-axelor:
Do not merge before 58753

PR #9640, opened 2023-01-12
58882 - Add condition when we have an payment condition with an line …

paa-axelor:
…with hold back option

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/db/repo/MoveManagementRepository.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/db/repo/MoveManagementRepository.java
@@ -184,7 +184,12 @@ public class MoveManagementRepository extends MoveRepository {
               && CollectionUtils.isNotEmpty(moveLine.getInvoiceTermList())) {
             if (moveLine.getInvoiceTermList().stream()
                 .allMatch(invoiceTermService::isNotReadonly)) {
-              moveLine.clearInvoiceTermList();
+              if (moveLine.getInvoiceTermList().stream()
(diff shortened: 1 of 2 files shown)


axelor-account/src/main/java/com/axelor/apps/account/db/repo/MoveManagementRepository.java
```diff
@@ -184,7 +184,12 @@ public Move save(Move move) {
               && CollectionUtils.isNotEmpty(moveLine.getInvoiceTermList())) {
             if (moveLine.getInvoiceTermList().stream()
                 .allMatch(invoiceTermService::isNotReadonly)) {
-              moveLine.clearInvoiceTermList();
+              if (moveLine.getInvoiceTermList().stream()
```
jmi-axelor, line 187: why do you add a if and not just do:
```
if(moveLine.getInvoiceTermList().stream()
                .allMatch(invoiceTermService::isNotReadonly) 
&& moveLine.getInvoiceTermList().stream()
.filter(invoiceTerm -> invoiceTerm.getIsHoldBack())
                      .count()
                  == 0)
```
Because we should also throw the AxelorException if there is an InvoiceTerm with isHoldBack = true

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/db/repo/MoveManagementRepository.java
paa-axelor: Fixed

jmi-axelor (approved):

PR #9648, opened 2023-01-13
56571 Forecast Line: fix company bank details filter

rth-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -2050,6 +2050,11 @@ InvoiceSet.length > 0))"
          (${(partner?.partnerAddressList?.findAll{it.isDeliveryAddr}.collect{it.address.id}+[null]).join(',')})&quot;"/>
   </action-attrs>
 
+  <action-attrs name="action-invoice-attrs-domain-company-bank-details">
+    <attribute name="domain" for="companyBankDetails"
+      expr="eval: &quot;self.company.id = ${company.id} AND self.active = true&quot;"/>
```
jmi-axelor, line 2055: can you add a "?" to avoid any NPE? 
`${company?.id}`

axelor-purchase/src/main/resources/views/PurchaseOrder.xml
```diff
@@ -664,6 +664,11 @@
     <attribute name="value" for="stockLocation.address" expr="eval: stockLocation.address"/>
   </action-attrs>
 
+  <action-attrs name="action-purchase-order-attrs-domain-company-bank-details">
+    <attribute name="domain" for="companyBankDetails"
+      expr="eval: &quot;self.company.id = ${company.id} AND self.active = true&quot;"/>
```
jmi-axelor, line 669: can you add a "?" to avoid any NPE? 
`${company?.id}`

jmi-axelor (changes requested):

jmi-axelor (approved):

axelor-cash-management/src/main/java/com/axelor/apps/cash/management/service/ForecastRecapServiceImpl.java
```diff
@@ -237,7 +237,7 @@ protected void populateWithForecastLineType(
             .bind("toDate", forecastRecap.getToDate())
             .bind("statusSelectList", statusSelectList)
             .bind(
-                "bankDetailsSet",
+                "bankDetails",
                 CollectionUtils.isEmpty(forecastRecap.getBankDetailsSet())
                     ? null
                     : forecastRecap.getBankDetailsSet())
```
jmi-axelor, line 243: I'm not sure this will work because in getFilter() method, there is : 
`AND (:bankDetails IS NULL OR self.companyBankDetails = :bankDetails)`
So the parameter bankDetails can't be a Set, it must be a single bankDetails

So I think this query must be changed to allow to use the forecastRecap.getBankDetailsSet()

jmi-axelor (changes requested):

PR #9665, opened 2023-01-17
Anomaly RM#59326 Account config: financial discount should not be req…

pda-axelor:
…uired fields


ale-axelor:
merged to dev branch 691aa167bc26c8726fef1fd7bba7882407f11b9c

PR #9690, opened 2023-01-20
59499 - Set date of first depreciation at end of month when the perio…

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/FixedAssetGenerationServiceImpl.java
```diff
@@ -114,6 +115,16 @@ public FixedAsset generateAndComputeLines(FixedAsset fixedAsset) throws AxelorEx
       generateAndComputeFiscalFixedAssetLines(fixedAsset);
       generateAndComputeFixedAssetDerogatoryLines(fixedAsset);
       generateAndComputeIfrsFixedAssetLines(fixedAsset);
+
+      if (fixedAsset.getFixedAssetCategory() != null
+          && fixedAsset.getFixedAssetCategory().getPeriodicityTypeSelect() != null
```
jmi-axelor, line 120: You don't need to check if an integer can be null

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/fixedasset/FixedAssetGenerationServiceImpl.java
paa-axelor: Resolved

jmi-axelor (approved):

PR #9700, opened 2023-01-23
59656 Cut off batch - the duration computation include the 1st day

rth-axelor:


jmi-axelor:
Merged in aos-sigma fork

PR #9747, opened 2023-01-31
Anomaly #59891 Batch bill of exchange : raise exception when the invoice's bank detail is inactive

jbo-axelor:
Anomaly #59891


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/batch/BatchBillOfExchange.java
```diff
@@ -124,27 +125,44 @@ protected void process() {
...
               e, "billOfExchangeBatch: create lcr accounting move", batch.getId());
-          break;
+          continue;
         }
       }
       JPA.clear();
     }
   }
 
-  @Transactional(rollbackOn = {Exception.class})
```
ale-axelor, line 145: why do we need to remove this transactional ? We need at least a transaction for the unitary process, else on error we could generate moves without updating the invoice.

PR #9778, opened 2023-02-03
60213 - Remove save on move when we change the date

paa-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -671,7 +671,6 @@
     <action name="action-move-method-set-moveline-dates"/>
     <action name="action-move-attrs-date-change"/>
     <action name="action-move-method-display-and-compute-due-date"/>
-    <action name="action-account-move-group-onsave"/>
```
jmi-axelor, line 674: You need to call every action in action-account-move-group-onsave that is linked to the Move.date
Per example, I think the action "action-move-method-update-invoice-terms" uses Move.date or even "action-move-method-update-due-date".

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9787, opened 2023-02-06
60191 FORECAST GENERATOR: Add weekly periodicity in forecast generation

rth-axelor:


jmi-axelor:
Merged in aos-sigma fork

PR #9831, opened 2023-02-10
60462 - Fill origin date with move date when we create, generate a mo…

paa-axelor:
…ve or when we change move date


jmi-axelor (approved):

jmi-axelor:
Merged in aos-sigma fork

PR #9849, opened 2023-02-14
60658 - Hide analytic panel on MoveLine/InvoiceLine/MoveTemplateLine/SaleOrderLine/PurchaseOrderLine

paa-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -2036,6 +2038,11 @@ InvoiceSet.length > 0))"
       if="_operationTypeSelect == 3"/>
   </action-attrs>
 
+  <action-attrs name="action-invoice-attrs-set-hidden-analytic-accounting-panel">
```
jmi-axelor, line 2041: This action must also be called when we change the company

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #9852, opened 2023-02-15
60650 Payment Session: add partner set field to filter partners linke…

rth-axelor:
…d to these invoice terms


jmi-axelor:
merged in aos-sigma

PR #9874, opened 2023-02-20
60781 add of new field in partner for compensation and modify of paym…

rth-axelor:
…ent session process


jmi-axelor:
Merged in aos-sigma fork

PR #9918, opened 2023-02-27
RM#60957: Move Line: corrected duplicated analytic line in move line …

jdo-axelor:
…generated from template

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveTemplateService.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveTemplateService.java
@@ -17,6 +17,7 @@
  */
 package com.axelor.apps.account.service.move;
 
+import com.axelor.apps.account.db.AnalyticMoveLine;
 import com.axelor.apps.account.db.Move;
 import com.axelor.apps.account.db.MoveLine;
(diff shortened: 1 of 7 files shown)


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveTemplateService.java
```diff
@@ -229,7 +229,7 @@ public List<Long> generateMove(MoveTemplate moveTemplate, List<HashMap<String, O
                     .divide(hundred, RoundingMode.HALF_UP);
 
             MoveLine moveLine =
-                moveLineCreateService.createMoveLine(
+                moveLineCreateService.createMoveLineWithoutAnalytic(
```
jmi-axelor, line 232: Don't create new method but please do something similar as here: https://github.com/axelor/axelor-open-suite/pull/9928/files#diff-9bd17286991e8863c210e5ff0adab519b8f7b312a770223865ad92a57456e3cdR469

jmi-axelor (changes requested):
Please check in all the code and fix all the similar cases to this one and the https://github.com/axelor/axelor-open-suite/pull/9928

jmi-axelor (approved):

PR #10049, opened 2023-03-23
Feature RM#61945: Date format: add a new locale per company

bsu-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/project/ProjectPlanningTimeComputeNameServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+import com.axelor.apps.project.db.Project;
+import com.google.inject.Inject;
+import java.time.LocalDate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProjectPlanningTimeComputeNameServiceImpl
+    implements ProjectPlanningTimeComputeNameService {
+
+  @Inject protected DateService dateService;
```
ale-axelor, line 31: please use injection by constructor

axelor-human-resource/src/main/java/com/axelor/apps/hr/service/timesheet/TimesheetComputeNameServiceImpl.java
```diff
@@ -17,13 +17,19 @@
...
 import com.axelor.apps.hr.db.Timesheet;
+import com.google.inject.Inject;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class TimesheetComputeNameServiceImpl implements TimesheetComputeNameService {
 
+  @Inject protected DateService dateService;
```
ale-axelor, line 31: Please use injection by constructor here

axelor-base/src/test/java/com/axelor/apps/base/test/TestCompanyDateService.java
```diff
@@ -0,0 +1,61 @@
...
+    companyDateService = new CompanyDateService();
+    usCompany = prepareCompany(Locale.US.toLanguageTag());
+    frCompany = prepareCompany(Locale.FRANCE.toLanguageTag());
+  }
+
+  @Test
+  public void testDateFormatter() throws AxelorException {
+    Assert.assertEquals(getDateTimeFormat(M_D_YYYY), getCompanyDateformat(usCompany));
+    Assert.assertEquals(getDateTimeFormat(DD_MM_YYYY), getCompanyDateformat(frCompany));
+  }
```
ale-axelor, line 36: please split so we can have one assert by method (for example here: `testDateFormatterUs()` `testDateFormatterFr()`) as this is a good practice to make one assert per method.

Please do the same for the other method

axelor-base/src/test/java/com/axelor/apps/base/test/TestCompanyDateService.java
```diff
@@ -0,0 +1,61 @@
...
+
+  protected Company prepareCompany(String languageTag) {
+    Company company = new Company();
+    company.setLocale(languageTag);
+    return company;
+  }
+
+  protected String getDateTimeFormat(String pattern) {
+    return DateTimeFormatter.ofPattern(pattern).format(LocalDateTime.now());
```
ale-axelor, line 51: I think it will not necessary cause issue, but to follow best practice we should avoid having a test that is different on every call. Please use a fixed LocalDate and LocalDateTime for the test.

ale-axelor (changes requested):

PR #10138, opened 2023-04-05
62407 - INVOICETERM / PAYMENTVOUCHER : blocking message when trying t…

alg-axelor:
…o pay


jmi-axelor (approved):

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveLineInvoiceTermServiceImpl.java
```diff
@@ -80,10 +80,21 @@ public void generateDefaultInvoiceTerm(
       throws AxelorException {
     Move move = moveLine.getMove();
 
+    boolean containsHoldback =
+        move.getPaymentCondition().getPaymentConditionLineList().stream()
```
jmi-axelor, line 84: We get a double NPE now, if move.paymentCondition is null and if paymentCondition.paymentConditionLineList is null
Before it was not the case because it was done after the:
```
else if (move.getPaymentCondition() == null
        || CollectionUtils.isEmpty(move.getPaymentCondition().getPaymentConditionLineList())) {
```

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #10140, opened 2023-04-06
RM61816 : Removed bank-payment dependency in supplychain module

erg-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/AccountingBatchSupplyChainService.java
```diff
@@ -1,32 +0,0 @@
...
-import com.axelor.apps.bankpayment.service.batch.AccountingBatchBankPaymentService;
-import com.axelor.apps.base.db.Batch;
-import com.axelor.apps.supplychain.service.batch.BatchAccountingCutOffSupplyChain;
-import com.axelor.inject.Beans;
-
-public class AccountingBatchSupplyChainService extends AccountingBatchBankPaymentService {
-
-  @Override
-  public Batch accountingCutOff(AccountingBatch accountingBatch) {
-    return Beans.get(BatchAccountingCutOffSupplyChain.class).run(accountingBatch);
```
jmi-axelor, line 30: You cannot just simply remove this without adding anything, we need this override, but maybe you need to find another way to do it

jmi-axelor (changes requested):

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/AccountingBatchSupplyChainService.java
erg-axelor: No need to find another way to do it, it's already done since BatchAccountingCutOffSupplyChain extends and is binded to BatchAccountingCutOff, when we call the .run on BatchAccountingCutOff it'll use the one from BatchAccountingCutOffSupplyChain which override it.

PR #10144, opened 2023-04-06
62867 - Fix counterpart and invoice terms generation when we have the same financial account for supplier and holdback

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveLineInvoiceTermServiceImpl.java
```diff
@@ -187,6 +188,12 @@ protected MoveLine computeInvoiceTermWithHoldback(
       throws AxelorException {
     MoveLine holdbackMoveLine = this.getHoldbackMoveLine(moveLine, move, holdbackAccount);
 
+    if (moveLine.getAccount().equals(holdbackAccount)
+        && paymentConditionLine.getIsHoldback()
+        && this.isHoldbackAlreadyGenerated(move, holdbackAccount)) {
+      holdbackMoveLine = null;
+    }
```
aol-axelor, line 195: Please set the value of `holdbackMoveLine` if condition is false instead of setting a value and then setting it to null if condition is true

PR #10200, opened 2023-04-17
RM#58442: INVOICE / INVOICE PAYMENT / INVOICE TERM: corrected financial discount not updatable on invoice-term, on invoice payment and early on invoice.

jdo-axelor:


axelor-account/src/main/resources/views/InvoicePayment.xml
```diff
@@ -65,12 +65,17 @@
         showIf="applyFinancialDiscount &amp;&amp; totalAmountWithFinancialDiscount != null"
         readonly="true"/>
       <panel name="financialDiscountAmountPanel"
-        showIf="applyFinancialDiscount &amp;&amp; !$isMultiCurrency" readonly="true" colSpan="12">
-        <field name="financialDiscount"/>
+        showIf="applyFinancialDiscount &amp;&amp; !$isMultiCurrency" readonlyIf="statusSelect > 0"
+        colSpan="12">
+        <field name="financialDiscount"
+          onChange="action-invoice-payment-method-update-discount"/>
```
jmi-axelor, line 71: I tried to pay a customer Invoice with an InvoicePayment. The invoice had a 10% financialDiscount, the total before discount was 84€.
So when I tried to pay it with an InvoicePayment, the default computed amount for the InvoicePayment was 75.6€ (84€ - 10%), there was a financialDiscount on the InvoicePayment, I removed it and tried to pay the whole Invoice, so 84€ and I get the following error message:
"Le montant payé est supérieur au(x) montant(s) restant à payer sur le(s) échéance(s)."
For me, if we decide to remove the financialDiscount on the InvoicePayment, we should be able to pay the whole Invoice without the discount. 

Also when we delete the financialDiscount, set applyFinancialDiscount to false. And put applyFinancialDiscount outside of the discountPanel and remove the `readonly="true"`

axelor-account/src/main/resources/views/InvoiceTerm.xml
```diff
@@ -239,8 +239,9 @@
       <panel name="financialDiscountPanel" title="Financial discount" colSpan="12"
         itemSpan="4" if="__config__.app.getApp('account').getManageFinancialDiscount()"
         hidden="true" showIf="applyFinancialDiscount &amp;&amp; $showFinancialDiscount"
-        readonly="true">
-        <field name="financialDiscount"/>
+        readonlyIf="amount != amountRemaining || isPaid">
+        <field name="financialDiscount"
+          onChange="action-invoice-term-method-update-financial-discount"/>
```
jmi-axelor, line 244: when you delete the financialDiscount, set applyFinancialDiscount  to false

jmi-axelor (changes requested):

axelor-account/src/main/resources/views/InvoicePayment.xml
```diff
@@ -65,12 +65,17 @@
...
         <field name="financialDiscountDeadlineDate"/>
```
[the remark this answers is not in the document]
jdo-axelor: Specification ticket, "Also, when I register a payment (InvoicePayment), I should be able to remove or change the selected financial discount", the readonly if if is on the panel

PR #10212, opened 2023-04-18
RM#6266 Currency : number of decimal

alo-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceLineServiceImpl.java
```diff
@@ -368,14 +376,16 @@ public void compute(Invoice invoice, InvoiceLine invoiceLine) throws AxelorExcep
       invoiceLine.setTaxCode(invoiceLine.getTaxLine().getTax().getCode());
     }
 
+    BigDecimal price = qty.multiply(priceDiscounted).setScale(scale, RoundingMode.HALF_UP);
+
```
gdu-axelor, line 380: Better to use another name for this variable. Price means unit price not a total

gdu-axelor (changes requested):

PR #10230, opened 2023-04-18
62546 - Invoice: invoice terms not equals to 100%

alg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentToolServiceImpl.java
```diff
@@ -248,15 +248,29 @@ public void checkConditionBeforeSave(InvoicePayment invoicePayment) throws Axelo
...
+            .reduce(BigDecimal::add)
+            .orElse(BigDecimal.ZERO);
+    if (CollectionUtils.isNotEmpty(invoiceTermList)) {
+      if (!startCurrency.equals(targetCurrency)) {
+        amount =
+            currencyService.getAmountCurrencyConvertedAtDate(
+                startCurrency, targetCurrency, amount, date);
+      }
+    }
+    return amount;
```
jmi-axelor, line 273: We never redo a conversion, we have 2 fields on an InvoiceTerm to manage multi currency:
amountRemaining = the remaining amount in the currency on the InvoiceTerm
companyAmountRemaining = the remaining amount in the currency on the company

That's all you need, so instead of doing the conversion, you need to get the right remaining amount, amountRemaining or companyAmountRemaining depending on the payment's currency.
And the payment's currency will always be equal to the InvoiceTerm currency or the company currency because in AOS, per example we don't allow to do a payment in £ if the InvoiceTerm currency is $ and the company currency is €.

axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentValidateServiceImpl.java
```diff
@@ -467,6 +467,7 @@ public Move fillMove(
         move.addMoveLineListItem(financialDiscountVatMoveLine);
       }
     }
+    move.setInvoice(invoice);
```
jmi-axelor, line 470: We only set the invoice field on a Move if it's a Move generated from an Invoice, here is not the case, it's a payment Move.

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveLineControlServiceImpl.java
```diff
@@ -111,7 +111,7 @@ public void validateMoveLine(MoveLine moveLine) throws AxelorException {
 
     if (CollectionUtils.isNotEmpty(moveLine.getInvoiceTermList())) {
       List<InvoiceTerm> invoiceTermList = moveLine.getInvoiceTermList();
-      Invoice invoiceAttached = invoiceTermList.get(0).getInvoice();
+      Invoice invoiceAttached = moveLine.getMove().getInvoice();
```
jmi-axelor, line 114: You need to find another way to fix this, we need to keep `invoiceTermList.get(0)` and also because in another comment I ask you to remove the invoice you set on the Move, so this fix won't work anymore.

jmi-axelor (changes requested):

jmi-axelor (changes requested):
Please check my comments and we'll talk about it Tuesday afternoon when I come back

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateServiceImpl.java
```diff
@@ -328,17 +328,22 @@ protected void checkMoveLineInvoiceTermBalance(Move move) throws AxelorException
...
-                  .map(InvoiceTerm::getAmount)
+                  .map(InvoiceTerm::getCompanyAmount)
                   .reduce(BigDecimal.ZERO, BigDecimal::add);
 
-          if (totalMoveLineInvoiceTerm.compareTo(moveLine.getDebit().max(moveLine.getCredit()))
-              != 0) {
+          BigDecimal moveLineAmount =
+              isCompanyCurrency
+                  ? moveLine.getCurrencyAmount()
+                  : moveLine.getDebit().max(moveLine.getCredit());
```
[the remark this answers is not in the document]
alg-axelor: Right

jmi-axelor (approved):

PR #10233, opened 2023-04-19
Feature RM#63232: HR FILES: Link files to trainingSkill

sda-axelor:


axelor-human-resource/src/main/resources/domains/EmployeeFile.xml
```diff
@@ -12,6 +12,18 @@
     <date name="recordDate" title="Record date"/>
     <many-to-one name="employee" ref="com.axelor.apps.hr.db.Employee"/>
     <date name="expirationDate" title="Expiration date"/>
+    <string name="fullName" namecolumn="true" search="fileType,metaFile" title="Full name">
+    	<![CDATA[
+    		String fullName = "";
+    		if(fileType != null){
+    			fullName = fileType.getName();
+    		}
+    		if(metaFile != null){
```
ale-axelor, line 21: please also check here if `fileType.getName()` and `metaFile.getFileName()` are not null, we need to be safe with getter methods since the method is often called.

axelor-talent/src/main/resources/views/TrainingSkill.xml
```diff
@@ -16,12 +17,32 @@
...
         onChange="action-training-skills-record-set-end-of-validity-date"/>
       <field name="endOfValidityDate"/>
     </panel>
   </form>
 
+  <action-attrs name="action-employee-file-attrs-set-employee">
+    <attribute name="value" for="employee" expr="eval: __this__?.employee"
+      if="__this__?.employee != null"/>
+    <attribute name="value" for="employee" expr="eval: __parent__?._parent"
+      if="eval: __this__?.employee == null &amp;&amp;  __parent__?.getContextClass() ==  TrainingSkill.class &amp;&amp; __parent__?.getParent()?.getContextClass() == Employee.class"/>
```
ale-axelor, line 43: does `if` accepts `eval:` ? (if it works, it's fine)

ale-axelor (changes requested):

PR #10277, opened 2023-04-26
Dev 63446

alg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoiceTermPaymentServiceImpl.java
```diff
@@ -230,13 +241,14 @@ protected InvoiceTermPayment initInvoiceTermPayment(
...
     invoiceTermPayment.setInvoiceTerm(invoiceTermToPay);
-    invoiceTermPayment.setPaidAmount(paidAmount);
     manageInvoiceTermFinancialDiscount(
         invoiceTermPayment, invoiceTermToPay, applyFinancialDiscount);
 
     boolean isCompanyCurrency =
         invoicePayment == null
             || invoicePayment.getCurrency().equals(invoiceTermToPay.getCompany().getCurrency());
+    invoiceTermPayment.setPaidAmount(
+        isCompanyCurrency ? this.computePaidAmount(invoiceTermToPay, paidAmount) : paidAmount);
```
jmi-axelor, line 251: I don't understand why you call computePaidAmount if isCompanyCurrency is true. 
If it's true, then we don't need to do the conversion you're doing and the ratio will always be 1, because  invoiceTerm.getAmount() = invoiceTerm.getCompanyAmount()

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoiceTermPaymentServiceImpl.java
jmi-axelor: Also we need to set the paidAmount on the invoiceTermPayment before calling the method manageInvoiceTermFinancialDiscount, why did you move it after?

jmi-axelor (changes requested):

PR #9745, opened 2023-01-30
59822 - Improve JPA cache management on payment session

paa-axelor:

PR #10325, opened 2023-05-04
RM#63785 : fix infinite loop in payment session cancel

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/paymentsession/PaymentSessionCancelServiceImpl.java
```diff
@@ -64,11 +64,14 @@ public void cancelInvoiceTerms(PaymentSession paymentSession) {
     List<InvoiceTerm> invoiceTermList;
     Query<InvoiceTerm> invoiceTermQuery =
         invoiceTermRepo.all().filter("self.paymentSession = ?", paymentSession).order("id");
+    int offset = 0;
 
-    while (!(invoiceTermList = invoiceTermQuery.fetch(jpaLimit)).isEmpty()) {
+    while (!(invoiceTermList = invoiceTermQuery.fetch(jpaLimit, offset)).isEmpty()) {
       for (InvoiceTerm invoiceTerm : invoiceTermList) {
         this.cancelInvoiceTerm(invoiceTerm);
+        offset++;
```
jmi-axelor, line 72: You can't put an offset here, some InvoiceTerms will be ignored if you do this.
The method cancelInvoiceTerm removes the paymentSession on the InvoiceTerm and it has a @Transactional, so when you fetch the query again, you'll only get the remaining attached InvoiceTerms, not the ones you previously removed.
I tested the #9745 and it works fine without the offset when you cancel a PaymentSession directly so you need to find why it doesn't work when we cancel a BankOrder

changelogs/unreleased/63785.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Inventory : unit from lines can not be changed anymore"
```
jmi-axelor, line 2: What's this?

jmi-axelor (changes requested):

PR #10343, opened 2023-05-09
RM#64108 Correct NPE when selecting partner and Company is empty

nbu-axelor:
When creating an account move, when you select a partner and filter the partners by the company, we check that the company is not null. Added a condition in filterPartner method in MoveViewHelperServiceImpl.

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveViewHelperServiceImpl.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveViewHelperServiceImpl.java
@@ -29,19 +29,22 @@ public class MoveViewHelperServiceImpl implements MoveViewHelperService {
 
   @Override
   public String filterPartner(Company company, Journal journal) {
-    Long companyId = company.getId();
-    String domain = "self.isContact = false AND " + companyId + " member of self.companySet";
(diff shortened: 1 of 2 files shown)


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveViewHelperServiceImpl.java
```diff
@@ -29,19 +29,22 @@ public class MoveViewHelperServiceImpl implements MoveViewHelperService {
...
-      String[] partnerSet = journal.getCompatiblePartnerTypeSelect().split(", ");
-      String lastPartner = partnerSet[partnerSet.length - 1];
-      for (String partner : partnerSet) {
-        domain += "self." + partner + " = true";
-        if (!partner.equals(lastPartner)) {
-          domain += " OR ";
+    String domain = "self.isContact = false";
+    if (company != null) {
+      Long companyId = company.getId();
+      domain = " AND " + companyId + " member of self.companySet";
```
ssb-axelor, line 35: There is a missing + that triggers an error when company is empty 
Also, maybe it's not necessary to store companyId in a variable, we can use it directly in query
```suggestion
      domain += " AND " + company.getId() + " member of self.companySet";
```

PR #10396, opened 2023-05-15
RM#63787: ACCOUNTING REPORT DAS 2: corrected export not exporting if …

jdo-axelor:
…n4ds code is missing.


axelor-account/src/main/java/com/axelor/apps/account/service/move/PaymentMoveLineDistributionServiceImpl.java
```diff
@@ -80,6 +80,9 @@ public void generatePaymentMoveLineDistributionList(Move move, Reconcile reconci
           new PaymentMoveLineDistribution(
               move.getPartner(), reconcile, moveLine, move, moveLine.getTaxLine());
 
+      if (moveLine.getAccount().getServiceType().getN4dsCode() == null) {
```
aol-axelor, line 83: I can save an account without a service type so maybe there is a potential NPE here

PR #10403, opened 2023-05-16
60269 - Fix unbalanced move when we generate a move with a percent mo…

paa-axelor:
…ve template


axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -839,6 +840,33 @@ public MoveLine createMoveLineForAutoTax(
...
+          moveLineList.stream()
+              .filter(
+                  ml ->
+                      ml.getAccount()
+                          .getAccountType()
+                          .getTechnicalTypeSelect()
+                          .equals(AccountTypeRepository.TYPE_RECEIVABLE))
+              .map(it -> it.getDebit().add(it.getCredit()))
+              .reduce(BigDecimal::add)
+              .orElse(BigDecimal.ZERO);
```
aol-axelor, line 867: Please create a method with an account type as parameter to compute those

paa-axelor: Resolved

jmi-axelor (approved):

PR #10431, opened 2023-05-22
64495 - Add check tax amount on mass entry control

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateServiceImpl.java
```diff
@@ -804,7 +804,7 @@ public void checkMoveLinesPartner(Move move) throws AxelorException {
     }
   }
 
-  protected void checkTaxAmount(Move move) throws AxelorException {
+  public void checkTaxAmount(Move move) throws AxelorException {
```
jmi-axelor, line 807: Don't you need to add a @Override ?

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveValidateServiceImpl.java
paa-axelor: Resolved

PR #10495, opened 2023-05-30
64817 - ANALYTICRULES : delete wrong method to retrieve AnalyticRules

alg-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/AccountService.java
```diff
@@ -274,6 +278,23 @@ public Account fillAccountCodeOnImport(Account account, int lineNo) throws Axelo
...
+    if (CollectionUtils.isNotEmpty(analyticRulesIdsList)) {
+      for (Long analyticRuleId : analyticRulesIdsList) {
+        AnalyticRules analyticRule = analyticRulesRepository.find(analyticRuleId);
+        if (analyticRule.getAnalyticAccountSet() != null) {
+          for (AnalyticAccount analyticAccount : analyticRule.getAnalyticAccountSet()) {
+            analyticAccountList.add(analyticAccount);
+          }
+        }
+      }
+    }
```
aol-axelor, line 294: Isn't is possible to directly fetch those accounts instead of two for loops ? You could put the query part of `getRulesIds` in a method and use it in an `analyticAccount` query in a new method for example.
Also I think there can be duplicates in the list.

PR #10531, opened 2023-06-02
65009 - Fix auto set of functional origin at move generation in mass entry process

paa-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -429,7 +429,7 @@
             onSelect="action-move-mass-entry-attrs-set-company-bank-details-domain"
             form-view="bank-details-form" grid-view="bank-details-grid"/>
           <field name="functionalOriginSelect" readonlyIf="statusSelect > 2" colSpan="3"
-            hideIf="journal.authorizedFunctionalOriginSelect.split(',').length == 1 &amp;&amp; journal.authorizedFunctionalOriginSelect == functionalOriginSelect || journal.authorizedFunctionalOriginSelect == null &amp;&amp; functionalOriginSelect == null"
+            hideIf="(journal.authorizedFunctionalOriginSelect.split(',').length == 1 &amp;&amp; journal.authorizedFunctionalOriginSelect == functionalOriginSelect) || (journal.authorizedFunctionalOriginSelect == null &amp;&amp; functionalOriginSelect == null) || journal == null || journal.authorizedFunctionalOriginSelect.split(',').length == null"
```
aol-axelor, line 432: `variable == null` can be replaced by `!variable` in javascript. Please do that here so that condition can be a bit smaller

PR #10533, opened 2023-06-02
RM#64965 Expense: VAT per Expense line

nbu-axelor:
Added boolean field in 'product-expense' to know if the move line will be generated without a taxLine, the field 'totalTax' will not be accessible. If the product is configured with VAT in percent we will auto calculate the 'totalTax' with 'totalAmount' and the tax from the product. Also if the product on the expenseLine is configured with VAT in percent, when we ventilate we do not check the tax amount.


axelor-human-resource/src/main/resources/i18n/messages_fr.csv
```diff
@@ -1,6 +1,9 @@
...
 "%d expenses treated successfully,","%d notes de frais traitées avec succès.",,
+"- Tax in amount : Tax amount must be entered manually on the expense line <br> - No tax : Tax amount entry is disabled on the expense line and can't be entered <br> - Tax in percentage : Tax amount is computed automatically based on configured tax rate","- TVA en montant : Le montant TVA doit être saisi manuellement dans la ligne de note de frais <br>
+                                                                                                                                                                                                                                                                - Aucune TVA : Le montant TVA est désactivé dans la ligne de note de frais et ne peut être saisi <br>
+                                                                                                                                                                                                                                                                - TVA en % : Le montant TVA est calculé en fonction du taux de TVA paramétré",,
```
ale-axelor, line 6: please put everything on the same line for the translation like the key

ale-axelor (changes requested):

axelor-human-resource/src/main/java/com/axelor/apps/hr/web/expense/ExpenseController.java
```diff
@@ -96,6 +97,21 @@ public class ExpenseController {
 
   private final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+  public void autoCalculateTotalTaxFromProductAndTotalAmount(
+      ActionRequest request, ActionResponse response) {
+    ExpenseLine expenseLine = request.getContext().asType(ExpenseLine.class);
+    if (expenseLine != null) {
+      if (expenseLine.getExpenseProduct().getTaxComputeModeSelect() == ProductRepository.NO_TAX) {
```
ale-axelor, line 104: I think this can throw a NPE if expenseProduct is null, can you check ?

Also I think you can move this part to the service method, as it is a functional check.

axelor-human-resource/src/main/java/com/axelor/apps/hr/service/expense/ExpenseServiceImpl.java
```diff
@@ -204,6 +208,64 @@ public ExpenseLine computeAnalyticDistribution(ExpenseLine expenseLine) {
     return expenseLine;
   }
 
+  @Override
+  public ExpenseLine getTotalTaxFromProductAndTotalAmount(ExpenseLine expenseLine) {
```
ale-axelor, line 212: this method is still too complex, please split it. Eventually it would be interesting to create a separate service for this method (for example `ExpenseLineComputationService`)

ale-axelor (changes requested):

PR #10537, opened 2023-06-02
65098 - Fix when we change journal, functional origin isn't reset if …

paa-axelor:
…we to an journal without authorized functional origin


axelor-account/src/main/java/com/axelor/apps/account/service/move/record/MoveRecordSetServiceImpl.java
```diff
@@ -181,6 +181,8 @@ public void setFunctionalOriginSelect(Move move) {
       } else {
         move.setFunctionalOriginSelect(null);
       }
+    } else {
+      move.setFunctionalOriginSelect(null);
```
aol-axelor, line 185: Please use a single `if` condition if both `else` code is the same

axelor-account/src/main/resources/views/Move.xml
```diff
@@ -90,7 +90,7 @@
           <field name="functionalOriginSelect"
             readonlyIf="statusSelect != 1 &amp;&amp; statusSelect != 2 &amp;&amp; statusSelect != 5"
             colSpan="3"
-            hideIf="journal.authorizedFunctionalOriginSelect.split(',').length == 1 &amp;&amp; journal.authorizedFunctionalOriginSelect == functionalOriginSelect || journal.authorizedFunctionalOriginSelect == null &amp;&amp; functionalOriginSelect == null"/>
+            hideIf="(journal.authorizedFunctionalOriginSelect.split(',').length == 1 &amp;&amp; journal.authorizedFunctionalOriginSelect == functionalOriginSelect) || (journal.authorizedFunctionalOriginSelect == null &amp;&amp; functionalOriginSelect == null) || journal == null || journal.authorizedFunctionalOriginSelect.split(',').length == null"/>
```
aol-axelor, line 93: Please use `!variable` instead of `variable != null` to make this a bit shorter

changelogs/unreleased/65098.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Move : Fix when we change journal, functional origin isn't reset if we change to an journal without authorized functional origin"
```
aol-axelor, line 2: Please correct typo `an` followed by a word not starting with a vowel

PR #10589, opened 2023-06-09
Feature#61974

Kyouis:
Ajout de limite supérieur et inférieur pour le accountingBatch quant au calcul de la some totale des moves


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/batch/BatchAccountingCutOffSupplyChain.java
```diff
@@ -117,6 +124,15 @@ protected void _processStockMovesByQuery(AccountingBatch accountingBatch) {
       for (StockMove stockMove : stockMoveList) {
         ++offset;
 
+        try {
```
kra-axelor, line 127: You can't make this `try...catch` here, or if you want to do so, you must trace the exception and increment the anomalies (by calling `incrementAnomaly()`) but you can't stop the batch (it is a batch after all).

A easier solution would be to call this method in `_processStockMove` at the very start ( since `checkPriceLimit` is more like a filter than a check, just return false if checkPriceLimit returns false), it would works as a filter and the exception would be already managed so you will have nothing to do on this part.

kra-axelor (changes requested):

PR #10603, opened 2023-06-12
65008 add immediate lcr management without bank order

rth-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/payment/paymentsession/PaymentSessionValidateServiceImpl.java
```diff
@@ -1111,4 +1143,259 @@ protected void recomputeAmountPaid(MoveLine moveLine) {
...
+      return null;
+    }
+
+    if (moveLine.getCredit().signum() != 0) {
+      moveLine.setCredit(moveLine.getCredit().add(invoiceTerm.getAmountPaid()));
+
+    } else {
+      moveLine.setDebit(moveLine.getDebit().add(invoiceTerm.getAmountPaid()));
+    }
+    moveLine.setCurrencyAmount(moveLine.getCurrencyAmount().add(invoiceTerm.getAmountPaid()));
```
aol-axelor, line 1398: I don't know if this is supposed to work with multi currency but if it does there might be an issue here

PR #10611, opened 2023-06-13
RM-61974 Add upper and lower limit to accounting batch

jma-axelor:


axelor-supplychain/src/main/resources/views/AccountingBatch.xml
```diff
@@ -9,6 +9,11 @@
       <attribute name="onChange"
         value="action-supplychain-accounting-batch-group-onchange-company"/>
     </extend>
+    <extend target="//field[@name='company']">
+      <insert position="after">
+
```
kra-axelor, line 14: Nothing to insert ?

kra-axelor (changes requested):

kra-axelor (changes requested):
Also, if you wish you can specify which I18n translations to generate with ./gradlew I18n -p [moduleLocation]  to not generate every translations from every module.
For example ./gradlew I18n -p modules/axelor-open-suite/axelor-account to generate translations only for account

PR #10629, opened 2023-06-15
65419 - MOVE : NPE when clicking on daybook button

alg-axelor:


changelogs/unreleased/65419.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Move : Fix NPE when we click on daybook button with a moveLine that contains no invoiceTerm"
```
aol-axelor, line 2: Please don't use technical terms like `NPE` or `invoiceTerm` in changelog (`invoice term` is fine though)

PR #10718, opened 2023-06-26
65418 - INVOICE : record not totally saved after validation

alg-axelor:


axelor-account/src/main/resources/views/Invoice.xml
```diff
@@ -1742,6 +1742,11 @@ InvoiceSet.length > 0))"
       if="invoiceTermList &amp;&amp; !invoiceTermList.isEmpty()"/>
   </action-group>
 
+  <action-group name="action-group-invoice-set-discount-values">
+    <action name="action-record-invoice-set-discount-values"/>
+    <action name="save"/>
+  </action-group>
```
aol-axelor, line 1748: I think it would be better to check if you can move that action before an already existing save instead of adding a new action group and save.

PR #10733, opened 2023-06-27
66178 - Fix moveLine Panel and error Panel when we remove journal or company

paa-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -456,14 +456,14 @@
       </panel>
     </panel>
 
-    <panel>
+    <panel hideIf="!journal">
```
aol-axelor, line 459: I think a panel containing a panel related is useless. You can remove it and add the `hideIf` condition to the panel related

PR #10753, opened 2023-06-29
RM#66402 : set invoice line tax from account when required

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -497,6 +497,7 @@ public List<MoveLine> createMoveLines(
         analyticLineService.printAnalyticAccount(moveLine, move.getCompany());
 
         TaxLine taxLine = invoiceLine.getTaxLine();
+
```
jmi-axelor, line 500: remove this please

axelor-account/src/main/resources/views/InvoiceLine.xml
```diff
@@ -526,6 +527,12 @@
       expr="eval: account?.manageCutOffPeriod ? __parent__?.cutOffEndDate : null"/>
   </action-record>
 
+  <action-record name="action-invoice-line-record-set-tax"
+    model="com.axelor.apps.account.db.InvoiceLine">
+    <field name="taxLine" expr="eval: account?.defaultTax?.activeTaxLine"
+      if="!taxLine &amp;&amp; account &amp;&amp; account.isTaxRequiredOnMoveLine"/>
```
jmi-axelor, line 533: why only when the tax is required on MoveLine? for me we should do it all the time if the taxLine is empty and there is a defaultTax on the selected account

jmi-axelor (changes requested):

PR #10797, opened 2023-07-05
RM#66603 : Improve custom accounting report error handling

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/custom/AccountingReportValuePercentageServiceImpl.java
```diff
@@ -77,9 +75,7 @@ public void createPercentageValue(
       if (accountingReport.getTraceAnomalies()
           && StringUtils.notEmpty(column.getPercentageBaseColumn())) {
         this.traceException(
-            new AxelorException(
-                TraceBackRepository.CATEGORY_INCONSISTENCY,
-                I18n.get(AccountExceptionMessage.CUSTOM_REPORT_WRONG_PERCENTAGE_BASE_COLUMN)),
+            AccountExceptionMessage.CUSTOM_REPORT_WRONG_PERCENTAGE_BASE_COLUMN,
```
jmi-axelor, line 78: I18n is missing right? Because the error message is not translated when I test

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/custom/AccountingReportValuePercentageServiceImpl.java
aol-axelor: I put I18n call in `traceException` method. Here I just forgot to add the translation.

jmi-axelor (approved):

PR #10808, opened 2023-07-05
66642 - Refacto printAnalyticAccount and create a method to retrieve …

paa-axelor:
…attrs map of axis analytic account

### DIFF
--- a/changelogs/unreleased/66642.yml
+++ b/changelogs/unreleased/66642.yml
@@ -1,0 +1,3 @@
+---
+title: "Analytic Line/Analytic Move Line : Refacto Analytic line service"
+type: Feature


axelor-account/src/main/java/com/axelor/apps/account/service/analytic/AnalyticLineService.java
```diff
@@ -39,11 +40,16 @@ public interface AnalyticLineService {
   List<Long> getAxisDomains(AnalyticLine line, Company company, int position)
       throws AxelorException;
 
+  List<Long> getAnalyticAccountIdList(Company company, int position) throws AxelorException;
```
aol-axelor, line 43: Maybe that method should be in `AnalyticAccountService` instead

axelor-account/src/main/java/com/axelor/apps/account/service/analytic/AnalyticLineService.java
```diff
@@ -39,11 +40,16 @@ public interface AnalyticLineService {
...
   boolean isAxisRequired(AnalyticLine line, Company company, int position) throws AxelorException;
 
   AnalyticLine checkAnalyticLineForAxis(AnalyticLine line);
 
   AnalyticLine printAnalyticAccount(AnalyticLine line, Company company) throws AxelorException;
 
   List<Long> getAnalyticAccountsByAxis(AnalyticLine line, AnalyticAxis analyticAxis);
+
+  Map<String, Map<String, Object>> getAnalyticAxisAttrsMap(
+      Company company, int startAxisPosition, int endAxisPosition) throws AxelorException;
```
aol-axelor, line 54: Same as before, but maybe in a `AnalyticAttrsService` or something like that
Also, its content is similar to what is done on moves for example. Can you check if it can be factorized ?

PR #10813, opened 2023-07-06
66713 fix modifications at onLoad of budget

rth-axelor:


axelor-budget/src/main/resources/views/Budget.xml
```diff
@@ -351,6 +342,16 @@
     <attribute for="code" name="focus" expr="true"/>
   </action-attrs>
 
+  <action-record name="action-budget-record-onchange-budget-level"
+    model="com.axelor.apps.budget.db.Budget">
+    <field name="fromDate"
+      expr="eval:_fromDate != null ? _fromDate : _parent?.fromDate != null ? _parent?.fromDate : fromDate"/>
+    <field name="toDate"
+      expr="eval:_toDate != null ? _toDate : _parent?.toDate != null ? _parent?.toDate : toDate"/>
```
aol-axelor, line 350: Please use `__parent__` instead of `_parent`. Also you can check `!= null` with `!` to make condition shorter

axelor-budget/src/main/resources/views/Budget.xml
```diff
@@ -351,6 +342,16 @@
     <attribute for="code" name="focus" expr="true"/>
   </action-attrs>
 
+  <action-record name="action-budget-record-onchange-budget-level"
```
aol-axelor, line 345: Please don't call a record action `onchange`

PR #10824, opened 2023-07-07
RM59759 ACCOUNTING REPORT : added move statuses filter

erg-axelor:


axelor-account/src/main/resources/views/AccountingReport.xml
```diff
@@ -185,9 +188,6 @@
           showIf="reportType.typeSelect == 3000 &amp;&amp; reportType.comparison == 3"/>
         <field name="displayDeltaYearPercentage"
           showIf="reportType.typeSelect == 3000 &amp;&amp; reportType.comparison == 3"/>
-        <field name="displaySimulatedMove"
```
jmi-axelor, line 188: if this field is not used anymore, please delete it in the domain and create a SQL script

jmi-axelor (changes requested):

PR #10834, opened 2023-07-07
RM#66827 SaleOrderLine : Project not emptied on copy

dar-axelor:


changelogs/unreleased/66827.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "SaleOrderLine : Project not emptied on copy"
```
kra-axelor, line 2: Please make the title more explicit (can be interpreted as, "made a fix so project are not emptied on copy")
For example: "SaleOrderLine: Fixed a bug where project was not emptied on copy"

kra-axelor (changes requested):

PR #10837, opened 2023-07-10
66742 - Set readonly move form when period is closed or doesn't exist

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/record/MoveRecordSetServiceImpl.java
```diff
@@ -70,11 +70,15 @@ public MoveRecordSetServiceImpl(
   }
 
   @Override
-  public void setPeriod(Move move) throws AxelorException {
-    if (move.getDate() != null && move.getCompany() != null) {
-      move.setPeriod(
-          periodService.getActivePeriod(
-              move.getDate(), move.getCompany(), YearRepository.TYPE_FISCAL));
+  public void setPeriod(Move move) {
```
jmi-axelor, line 73: This method is used on the Move's onNew, on the Move company's onChange and on the Move date's onChange
The exception should be thrown on the Move date's onChange!

jmi-axelor (changes requested):

PR #10901, opened 2023-07-19
Anomaly RM#67228: MANUF ORDER : saleOrderSet filtered on one partner

bsu-axelor:


changelogs/unreleased/fix-sale-order-set-filtered-on-one-partner.yml
```diff
@@ -0,0 +1,3 @@
+---
+title : "MANUF ORDER : saleOrderSet filtered on one partner"
```
kra-axelor, line 2: The title need to be explicit on what the fix did (it will appear on changelog, so final users need to understand what it did)
For example it could be "MANUF ORDER: Fixed a bug where sale order set was emptied on change of client partner"

PR #10933, opened 2023-07-25
Anomaly RM#67253:Improve modelization when using a PDF viewer in Empl…

vva-axelor:
…oyee and JobApplication


ale-axelor:
merged in #10967

PR #10941, opened 2023-07-26
67074 InvoiceLine : fix empty error message when changing taxLine to …

rth-axelor:
…empty


changelogs/unreleased/67074.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "InvoiceLine : fix npe when changing taxLine to empty"
```
aol-axelor, line 2: Please don't use technical terms like `npe` or `taxLine` in a changelog

PR #10943, opened 2023-07-26
Anomaly RM#67497 EMPLOYEE : value of statusSelect with hire and leave…

jka-axelor:
… date empty


axelor-human-resource/src/main/java/com/axelor/apps/hr/db/repo/EmployeeHRRepository.java
```diff
@@ -169,4 +157,31 @@ public static boolean isEmployeeFormerNewOrArchived(Employee employee) {
...
+        && employee.getHireDate() != null
+        && employee.getHireDate().compareTo(today.minusDays(30)) > 0) {
+      employeeStatus = "new";
+    } else if (leavingDate != null && leavingDate.compareTo(today) < 0) {
+      employeeStatus = "former";
+    } else {
+      employeeStatus = "active";
+    }
+    return employeeStatus;
+  }
```
alo-axelor, line 186: Please create a new service EmployeeComputeStatusService for this method. It will be called with a Beans.get() in the repository.

axelor-human-resource/src/main/java/com/axelor/apps/hr/db/repo/EmployeeHRRepository.java
```diff
@@ -169,4 +157,31 @@ public static boolean isEmployeeFormerNewOrArchived(Employee employee) {
...
+                : AuthUtils.getUser().getActiveCompany());
+
+    if (employee.getLeavingDate() == null
+        && employee.getHireDate() != null
+        && employee.getHireDate().compareTo(today.minusDays(30)) > 0) {
+      employeeStatus = "new";
+    } else if (leavingDate != null && leavingDate.compareTo(today) < 0) {
+      employeeStatus = "former";
+    } else {
+      employeeStatus = "active";
```
alo-axelor, line 183: These strings can be defined in constants in the new service.

alo-axelor (changes requested):

axelor-human-resource/src/main/java/com/axelor/apps/hr/db/repo/EmployeeHRRepository.java
```diff
@@ -24,39 +24,29 @@
...
+import com.google.inject.Inject;
 import java.time.LocalDate;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
 public class EmployeeHRRepository extends EmployeeRepository {
 
+  @Inject protected AppBaseService appBaseService;
+  @Inject protected EmployeeComputeStatusService employeeComputeStatusService;
```
alo-axelor, line 41: AppBaseService is not used in the repo, only in EmployeeComputeStatusService. Use Beans.get instead of injection.

alo-axelor (changes requested):

axelor-human-resource/src/main/java/com/axelor/apps/hr/db/repo/EmployeeHRRepository.java
cma-axelor: Can you please tell why Beans.get should be used instead of field injection ?
Is it due to any conventions ?
Thanks

alo-axelor: We generally want to use contructor injection and avoid Beans.get in services. But in repositories, since they are called in services, we want to avoid having injections in them. Preventing circulary injection. Also, like that we have a clearer (not optimal) structure :
Controller -> Services -> Repositories -> Beans.get and not Controller -> Services -> Repositories -> Services

PR #10952, opened 2023-07-27
Anomaly RM#67507 Manuf order plannification methode not launched

sro-axelor:


changelogs/unreleased/fix-manuf-order-plannification-methode-not-launched.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Manuf order plannification methode not launched"
```
kra-axelor, line 2: Title not explicit enough. It can be for example "Manuf Order: Fixed a bug some plannifications processes were not launched" (Processes can even be detailed I think, but should be ok anyway)

kra-axelor (changes requested):

PR #10969, opened 2023-07-31
Anomaly RM#67621 JOURNAL : Balance HTML tag no longer computed

sro-axelor:


changelogs/unreleased/fix-journal-balance-html-tag-no-longer-computed.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "JOURNAL : Balance HTML tag no longer computed"
```
kra-axelor, line 2: Not explicit enough (can be interpretated as "Balance is no longer computed")
Title can be: "JOURNAL: Balance HTML tag is now correcly computed"

kra-axelor (changes requested):

PR #10982, opened 2023-08-01
RM#66722 : fix stock details by product productionIndicators visibility AND reserved qty visibility

oma-axelor:


changelogs/unreleased/66722.yml
```diff
@@ -0,0 +1,5 @@
+title: "fix stock details by product productionIndicators visibility"
```
kra-axelor, line 1: Change title to add the 'object' affected, can be like this "Stock details: fix stock details ..."

kra-axelor (changes requested):

PR #10989, opened 2023-08-02
RM#65249 : translate and add helper for deb config in tax number

aol-axelor:


axelor-supplychain/src/main/resources/i18n/messages_fr.csv
```diff
@@ -241,7 +241,7 @@
 "Ignore end date","Ignorer la date de fin",,
 "Include element without date","Inclure les évènements sans date",,
 "Include elements with a date later than MRP end date.","Inclure les éléments dont la date est postérieure à la date de fin du CBN.",,
-"Include in DEB",,,
+"Include in DEB","Include dans la DEB",,
```
jmi-axelor, line 244: it's not "Include dans la DEB" but "Inclure dans la DEB"

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #10995, opened 2023-08-02
67702 - Fix date update on move lines

paa-axelor:


jmi-axelor (approved):

ale-axelor (changes requested):
I disagree with this implementation, having many booleans just to manage code logic makes the code difficult to read. Please refactor this method using submethods instead of booleans.

PR #10999, opened 2023-08-03
#67258: Business Project : Improve task management and reporting, Add a new section

cce-axelor:


kra-axelor (changes requested):
Also missing changelog

PR #11012, opened 2023-08-07
RM#66627 : add production planification capacity configuration and re…

oma-axelor:
…-add infinite capacity features


ale-axelor (changes requested):
Please:
- Create a new service for methods related to operation order planning: OperationOrderPlanningService and move the corresponding methods inside (including the method `plan(OperationOrder operationOrder, Long cumulatedDuration`).
- Add a new method `plan(OperationOrder operationOrder)` in OperationOrderWorkflowService which will call the method `operationOrderPlanningService.plan(OperationOrder operationOrder, Long cumulatedDuration)` .
- When methods are not called from outside their service, please set their visibility to `protected`.

axelor-production/src/main/resources/domains/ProductionConfig.xml
```diff
@@ -25,13 +25,19 @@
...
 
     <finder-method name="findByCompany" using="company"/>
 
     <extra-code>
 		<![CDATA[
    			public static final int REALIZE_START = 0;
    			public static final int REALIZE_FINISH = 1;
+
+   			public static final int PRODUCTION_CAPACITY_INFINITE = 1;
+   			public static final int PRODUCTION_CAPACITY_FINITE = 2;
```
ale-axelor, line 40: Please rename these variables to `INFINITE_CAPACITY_SCHEDULING`` and `FINITE_CAPACITY_SCHEDULING`

PR #11013, opened 2023-08-07
Possibility to choose between previous and next scale

pud-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingComputer.java
```diff
@@ -146,7 +146,13 @@ public void apply() throws AxelorException {
     LOG.debug("Treating pricing childs of {}", this.pricing);
     for (int counter = 0; counter < MAX_ITERATION; counter++) {
 
-      Optional<Pricing> optChildPricing = getNextPricing(currentPricing);
+      Optional<Pricing> optChildPricing = Optional.empty();
+
+      if (pricing.getLinkedPricing() == null) {
```
kra-axelor, line 151: I might be wrong, but is there a infinite loop for case with boolean is switched for"previousPricing" ?
**First iteration** for this case is ok, since linkedPricing of the root will be null, so for next iteration it will fetch pricing where linkedPricing = currentPricing (so its child) and then start second iteration with it
**On second iteration**, linkedPricing will not be null (pointed to the pricing of the first iteration because it is its parent) and so nextPricing will actually be determined as its current parent, repeating then the first iteration

A possible solution would be to add a boolean in init of PricingComputer (to know wether its "previous" or "next" mode) and then depending on the value, in PricingComputer you can either call `getPreviousPricing` or `Optional.ofNullable(currentPricing.getLinkedPricing())` to get the next pricing. (You can even keep the commented code above but you will have to add condition)

axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingComputer.java
```diff
@@ -132,10 +132,10 @@ public void apply() throws AxelorException {
     if (context == null || pricing == null || model == null || product == null) {
       throw new IllegalStateException("This instance has not been correctly initialized");
     }
-    if (pricing.getPreviousPricing() != null) {
-      throw new IllegalStateException(
-          "This method can only be called with root pricing (pricing with not previous pricing)");
-    }
+//    if (pricing.getLinkedPricing() != null) {
```
kra-axelor, line 135: Do not keep commented code

kra-axelor (changes requested):

axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingComputer.java
pud-axelor: @kra-axelor 
so i don't think it will be infinite looping

kra-axelor: Aah yes indeed, you are right

PR #11042, opened 2023-08-11
66646 - Add axis analytic account on purchase order line

paa-axelor:


changelogs/unreleased/66646.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "PurchaseOrder / SaleOrder : Manage analytic accounts per axis"
```
aol-axelor, line 2: Was there no changelog for when sale order where concerned ? This ticket is only about purchase orders. Also please add a space before `Order`

paa-axelor: Resolved

PR #11068, opened 2023-08-23
 RM#66728 : refactor stock computeIndicators method orchestration

oma-axelor:


changelogs/unreleased/66728.yml
```diff
@@ -0,0 +1,2 @@
+title: "Stock : refactor stock computeIndicators method orchestration"
```
kra-axelor, line 1: This changelog is more for functionnal than technical, so things as refactor should not be noted in changelog.
However since this work concerns a bug with indicators, the changelog can be "Fixed a bug where indicators were not always computed correctly." Or something like that

kra-axelor (changes requested):

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationServiceImpl.java
```diff
@@ -90,89 +103,84 @@ protected List<StockLocation> getNonVirtualStockLocations(Long companyId) {
...
+            + " FROM StockLocationLine sll"
+            + " WHERE (( (0 IN (%2$s)) AND sll.stockLocation.typeSelect != :stockLocationTypeSelectVirtual AND (sll.stockLocation.company.id = :companyId OR COALESCE(:companyId,0) = 0 ) )"
+            + " OR (sll.stockLocation.id in (%2$s) AND 0 NOT IN (%2$s) ))"
+            + " AND :productId is not null AND sll.product.id = :productId AND sll.product.stockManaged is TRUE GROUP BY sll.unit.id";
 
-      if (productUnit != null && !productUnit.equals(stockLocationLineUnit)) {
-        qty =
-            unitConversionService.convert(
-                stockLocationLineUnit, productUnit, qty, qty.scale(), product);
+    List<Tuple> sumOfQtyPerUnitList =
```
ale-axelor, line 121: I don't get why it is a List<Tuple> and not a Map<Long, BigDecimal>, can you check ?

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationServiceImpl.java
```diff
@@ -90,89 +103,84 @@ protected List<StockLocation> getNonVirtualStockLocations(Long companyId) {
...
-      Unit stockLocationLineUnit = stockLocationLine.getUnit();
-      qty =
-          qtyType.equals("real")
-              ? stockLocationLine.getCurrentQty()
-              : stockLocationLine.getFutureQty();
+    String query =
+        "SELECT sll.unit.id, sum(sll.%1$s)"
+            + " FROM StockLocationLine sll"
+            + " WHERE (( (0 IN (%2$s)) AND sll.stockLocation.typeSelect != :stockLocationTypeSelectVirtual AND (sll.stockLocation.company.id = :companyId OR COALESCE(:companyId,0) = 0 ) )"
+            + " OR (sll.stockLocation.id in (%2$s) AND 0 NOT IN (%2$s) ))"
```
ale-axelor, line 118: what is wrong with using a parameter here, why do we have to use a string format ?

ale-axelor (changes requested):

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationServiceImpl.java
oma-axelor: `getResultList` returns a list of a generic type of our choice. so it's natural to use a list here.
What do you suggest i do otherwise ?

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationServiceImpl.java
oma-axelor: I couldn't get it to work, i don't remember the exact reason but i suspect it was related to the `(0 IN (%2$s))` part.
The only way i could get this code snippet to work was to format the list element manually.

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockLocationServiceImpl.java
ale-axelor: I thought a map could be used, but after checking existing code, I did not find something better (we can do a List of Map, but it would not simplify the code)

PR #11077, opened 2023-08-25
RM#65712 : fix compute new average price for stockLocationLine in the case of a unit conversion

oma-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -722,6 +714,13 @@ public BigDecimal computeNewAveragePriceLocationLine(
               stockMoveLine.getProduct());
     }
 
+    BigDecimal oldQty = stockLocationLine.getCurrentQty().subtract(newQty);
+
+    if (oldAvgPrice == null || oldAvgPrice.signum() == 0 || oldQty.signum() == 0) {
+      oldAvgPrice = BigDecimal.ZERO;
+      oldQty = BigDecimal.ZERO;
+    }
```
ale-axelor, line 722: I think we can remove this check, as oldAvgPrice cannot be null and, if the fields are equals to 0, I don't see the point of setting them to 0

PR #11163, opened 2023-09-14
RM#69001 : reduce execution time for getContentStockLocationIds

oma-axelor:


oma-axelor:
When reviewing please treat each commit separately.

PR #11218, opened 2023-09-25
RM#69351 ProjectTask : realCost not computed properly when timeUnit is empty

dar-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/ProjectTaskReportingValuesComputingServiceImpl.java
```diff
@@ -322,33 +322,36 @@ protected void computeLandingValues(
...
-                              purchaseOrderLine.getPurchaseOrder().getStatusSelect();
-                          return purchaseOrderStatus == PurchaseOrderRepository.STATUS_VALIDATED
-                              || purchaseOrderStatus == PurchaseOrderRepository.STATUS_FINISHED;
-                        })
-                    .map(PurchaseOrderLine::getExTaxTotal)
-                    .reduce(BigDecimal::add)
-                    .orElse(BigDecimal.ZERO));
+    BigDecimal realCost = BigDecimal.ZERO;
+
+    if (unitIsTimeUnit) {
```
ale-axelor, line 334: I think it's better to call again "isTimeUnitValid" than to add a boolean as argument here

Issue #11364, opened 2023-10-13
Circular dependency when starting operation order in Manufacturing -> Manufacturing Order


avasilev:
I am using:
 - Axelor Open Suite Version: 7.0.6
 - Axelor SDK Version: 6.1.5

When trying to click start on Operation order in manufacturing order screen:
![image](https://github.com/axelor/axelor-open-suite/assets/4996846/51d5c7b6-158e-4123-984d-c3fed714947d)

 I've got following error:
[2;39m2023-10-13 10:56:59.567[m [31mERROR[m [35m1[m [2;39m---[m [2;39m[nio-8080-exec-6][m [36mc.a.a.b.s.exception.TraceBackService    [m [2;39m:[m com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) [Guice/CanNotProxyClass]: Tried proxying OperationOrderWorkflowService to support a circular dependency, but it is not an interface.
  while locating OperationOrderWorkflowServiceBusinessImpl
  at ManufOrderWorkflowMaintenanceServiceImpl.<init>(ManufOrderWorkflowMaintenanceServiceImpl.java:60)
      \_ for 1st parameter
  while locating ManufOrderWorkflowMaintenanceServiceImpl
  while locating ManufOrderWorkflowServiceImpl
  at OperationOrderWorkflowServiceBusinessImpl.<init>(OperationOrderWorkflowServiceBusinessImpl.java:51)
      \_ for 8th parameter
  while locating OperationOrderWorkflowServiceBusinessImpl
  while locating OperationOrderWorkflowService

Learn more:
  https://github.com/google/guice/wiki/CAN_NOT_PROXY_CLASS

1 error

======================
Full classname legend:
======================
ManufOrderWorkflowMaintenanceServiceImpl:  "com.axelor.apps.maintenance.service.ManufOrderWorkflowMaintenanceServiceImpl"
ManufOrderWorkflowServiceImpl:             "com.axelor.apps.production.service.manuforder.ManufOrderWorkflowServiceImpl"
OperationOrderWorkflowService:             "com.axelor.apps.production.service.operationorder.OperationOrderWorkflowService"
OperationOrderWorkflowServiceBusinessImpl: "com.axelor.apps.businessproduction.service.OperationOrderWorkflowServiceBusinessImpl"
========================
End of classname legend:
========================

	at com.google.inject.internal.InternalProvisionException.toProvisionException(InternalProvisionException.java:251)
[...]

PR #11371, opened 2023-10-16
Anomaly RM#70390: [Github 11364] - Circular dependency when starting …

bsu-axelor:
…operation order in Manufacturing Order


ale-axelor:
fixes issue #11364

PR #11405, opened 2023-10-19
70658 - Hide company without account config on company list when we c…

paa-axelor:
…hange move company


axelor-account/src/main/java/com/axelor/apps/account/service/move/attributes/MoveAttrsServiceImpl.java
```diff
@@ -423,4 +432,33 @@ protected boolean isPartnerRequired(Journal journal) {
...
+        CollectionUtils.isEmpty(companyList)
+            ? "0"
+            : companyWithAccountConfig.stream()
+                .map(Company::getId)
+                .map(Objects::toString)
+                .collect(Collectors.joining(","));
+
+    String domain = String.format("self.id IN (%s)", companyIds);
+
+    this.addAttr("company", "domain", domain, attrsMap);
```
aol-axelor, line 462: I don't think it should be done like this, you could just do something like `companyRepository.all().filter("self.accountConfig IS NOT NULL").fetch()`

paa-axelor: Resolved

PR #11411, opened 2023-10-20
Feature RM#70522 [SOP] Fill real values

pda-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/batch/BatchFillSopRealValues.java
```diff
@@ -0,0 +1,180 @@
...
+    this.appBaseService = appBaseService;
+  }
+
+  @Override
+  protected void process() {
+    ProductionBatch productionBatch = batch.getProductionBatch();
+    computeSopRealValues(productionBatch);
+  }
+
+  @Transactional
```
kra-axelor, line 75: Why is this method transactional ?

axelor-production/src/main/java/com/axelor/apps/production/service/batch/BatchFillSopRealValues.java
```diff
@@ -0,0 +1,180 @@
...
+  protected void computeSopRealValues(ProductionBatch productionBatch) {
+    Company company = productionBatch.getCompany();
+    for (Sop sop : productionBatch.getSopSet()) {
+      computeValuesPerSOP(sop, company);
```
kra-axelor, line 79: Please clear Jpa after every computing or every 10 or 20 computation, will need to do adaptions in somes method

axelor-production/src/main/java/com/axelor/apps/production/service/batch/BatchFillSopRealValues.java
```diff
@@ -0,0 +1,180 @@
...
+            .filter(
+                "self.statusSelect = :statusSelect AND self.product.productCategory = :productCategory AND self.company = :company AND self.realEndDateT BETWEEN :startDate AND :endDate")
+            .bind("statusSelect", ManufOrderRepository.STATUS_FINISHED)
+            .bind("company", company)
+            .bind("productCategory", productCategory)
+            .bind("startDate", period.getFromDate())
+            .bind("endDate", period.getToDate())
+            .fetchStream()
+            .map(ManufOrder::getQty)
+            .reduce(BigDecimal.ZERO, BigDecimal::add));
```
kra-axelor, line 133: Please, if it is not too difficult, use directly a query (createQuery method) to compute the sum instead of using a fetchStream

axelor-production/src/main/java/com/axelor/apps/production/service/batch/BatchFillSopRealValues.java
```diff
@@ -0,0 +1,180 @@
...
+        stockLocationLineHistoryRepo
+            .all()
+            .filter(
+                "self.stockLocationLine.stockLocation.company = :company AND self.stockLocationLine.product.productCategory = :productCategory AND self.dateT < :endDate")
+            .bind("company", company)
+            .bind("productCategory", productCategory)
+            .bind("endDate", period.getToDate())
+            .fetchStream()
+            .map(history -> history.getQty().multiply(history.getWap()))
+            .reduce(BigDecimal.ZERO, BigDecimal::add));
```
kra-axelor, line 149: Same as above, use directly a query (createQuery method) to compute the sum instead of using a fetchStream

axelor-production/src/main/java/com/axelor/apps/production/service/batch/BatchFillSopRealValues.java
```diff
@@ -0,0 +1,180 @@
...
+            .all()
+            .filter(
+                "self.saleOrder.statusSelect IN (3,4) AND self.product.productCategory = :productCategory AND self.saleOrder.company = :company AND self.saleOrder.confirmationDateTime BETWEEN :startDate AND :endDate")
+            .bind("productCategory", productCategory)
+            .bind("company", company)
+            .bind("startDate", period.getFromDate())
+            .bind("endDate", period.getToDate())
+            .fetchStream()
+            .map(SaleOrderLine::getInTaxTotal)
+            .reduce(BigDecimal.ZERO, BigDecimal::add));
```
kra-axelor, line 112: I think there is a functionnal error here, you can check with Mathis but saleOrderLine.getInTaxTotal is not always in the same currency so doing a sum without taking into account the currency is not correct. So instead you may use companyInTaxTotal
Also please if it is not too difficult, use directly a query (createQuery method) to compute the sum instead of using a fetchStream.

changelogs/unreleased/feature-sop-fill-real-values.yml
```diff
@@ -0,0 +1,3 @@
+---
+title : "[SOP] Fill real values"
```
kra-axelor, line 2: Please change to [SOP]: New batch that compute real sales, real manuf and real stock on SOP

kra-axelor (changes requested):

PR #11454, opened 2023-10-26
71038 -  Add journal when we check duplicate origins in move

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveToolServiceImpl.java
```diff
@@ -647,20 +647,22 @@ public void exceptionOnGenerateCounterpart(Move move) throws AxelorException {
   public List<Move> getMovesWithDuplicatedOrigin(Move move) {
     List<Move> moveList = null;
     StringBuilder query =
-        new StringBuilder("self.origin = :origin AND self.period.year = :periodYear");
+        new StringBuilder(
+            "self.origin = :origin AND self.period.year = :periodYear AND self.journal = :journal ");
     Map<String, Object> params = new HashMap<>();
 
-    if (!ObjectUtils.isEmpty(move.getOrigin()) && !ObjectUtils.isEmpty(move.getPeriod())) {
+    if (!move.getOrigin().isEmpty() && move.getPeriod() != null) {
```
jmi-axelor, line 654: Origin is a String, we'll get an NPE if it's null
Keep ObjectUtils.isEmpty() or use Strings.isNullOrEmpty(), as you prefer

jmi-axelor (changes requested):

Issue #11451, opened 2023-10-26
Error on Sale module install


claudiucotan:
Error while importing sale_saleConfig.csv
com.axelor.data.ImportException: Constraint violation error:

- [company] - ne doit pas être nul

there are important changes between version 7.1.6 and 7.2.0 on file "sale_saleConfig.csv".
The issue may come from there.

FYI i tried to install version 7.2.0 from scratch, while i faced this issue.

[sale_saleConfig_after_(version_7.2.0).csv](https://github.com/axelor/axelor-open-suite/files/13178369/sale_saleConfig_after_.version_7.2.0.csv)
[sale_saleConfig_before_(version_7.1.6).csv](https://github.com/axelor/axelor-open-suite/files/13178370/sale_saleConfig_before_.version_7.1.6.csv)

claudiucotan:
@axeloradmin please, this is blocking the install, is there a workaround?

claudiucotan:
![2023-10-27 14_50_41-Courrier](https://github.com/axelor/axelor-open-suite/assets/3496906/57faa50d-7e10-45ab-b6e1-4c9515fffef4)
![2023-10-27 14_53_06-(1) Dashboard — Webmin 2 101 on webhost03 (Ubuntu Linux 22 04 3)](https://github.com/axelor/axelor-open-suite/assets/3496906/beb44ad1-1565-460c-8148-a18e19f568f9)

claudiucotan:
![2023-10-27 14_43_27-sale_saleConfig_before_ version_7 1 6 csv - Excel](https://github.com/axelor/axelor-open-suite/assets/3496906/b376e62c-821e-467f-9843-c450e1b98c7e)
![2023-10-27 14_43_51-sale_saleConfig_after_ version_7 2 0 csv - Excel](https://github.com/axelor/axelor-open-suite/assets/3496906/197050d0-1507-4b74-9331-f6341baee021)

claudiucotan:
same error for the following modules:

- Accounting
- Stock
- Quality (production)

![2023-10-27 15_07_27-(1) Dashboard — Webmin 2 101 on webhost03 (Ubuntu Linux 22 04 3)](https://github.com/axelor/axelor-open-suite/assets/3496906/0583c562-59c7-4130-8183-94db2a29f650)
[...]

PR #11471, opened 2023-10-31
Anomaly RM#71122 [Github 11451] Error when installing the application…

pda-axelor:
… without demo data


ale-axelor:
fixes #11451

PR #11569, opened 2023-11-13
71678 InvoiceTerm : fix rounding issue when we unreconcile invoice pa…

rth-axelor:
…yment


axelor-account/src/main/java/com/axelor/apps/account/service/payment/invoice/payment/InvoicePaymentToolServiceImpl.java
```diff
@@ -107,6 +107,17 @@ public void updateAmountPaid(Invoice invoice) throws AxelorException {
 
   protected BigDecimal getAmountInInvoiceCompanyCurrency(BigDecimal amount, Invoice invoice)
```
ale-axelor, line 108: the method signature is strange, please change it to only have an invoice as parameter, and replace `amount` by `invoice.getAmountRemaining()`. Maybe we should also rename the method to `computeCompanyAmountRemaining`

ale-axelor (changes requested):

PR #11573, opened 2023-11-13
RM#71064: Stock : add button to split reception move lines into fulfi…

oma-axelor:
…lled ones and unfulfilled ones


axelor-stock/src/main/resources/i18n/messages_fr.csv
```diff
@@ -692,6 +692,7 @@
 "Split by unit","Scinder par unité",,
 "Split in Qty of","Scinder par qté de",,
 "Split into 2","Scinder en 2",,
+"Split into fulfilled line and unfulfilled one","Scinder en ligne remplies et lignes non remplie",,
```
kra-axelor, line 695: Please fix french translation

PR #11582, opened 2023-11-14
71280 Pricing : change the pricing query to use a groovy formula instead of product/product category fields

rth-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/pricing/PricingServiceImpl.java
```diff
@@ -150,44 +146,27 @@ protected void pricingFetchFilter(
...
-      if (productCategory != null) {
-        productFilter.append("self.productCategory = :productCategory ");
-        bindings.put("productCategory", productCategory);
+  protected List<Pricing> appendFormulaFilter(List<Pricing> pricings, Model model) {
+    Context scriptContext = new Context(Mapper.toMap(model), EntityHelper.getEntityClass(model));
+    ScriptHelper scriptHelper = new GroovyScriptHelper(scriptContext);
+    List<Pricing> filteredPricings = new ArrayList<>();
+    for (Pricing pricing : pricings) {
+      try {
+        if ((Boolean) scriptHelper.eval(pricing.getFormula())) {
```
ale-axelor, line 161: if the user writes a script that evaluates to null, I think this will hrow a NPE, we should avoid this by writing `Boolean.TRUE.equals(...)`

ale-axelor (changes requested):

PR #11711, opened 2023-12-04
 Feature #68564 APPCONFIG / COUNTRY / ADDRESS : Introduce the usage of string and groovy template method to compute addresses based on a template defined by country. 

dka-axelor:
Feature #68564 APPCONFIG / COUNTRY / ADDRESS : Introduce the usage of string and groovy template method to compute addresses based on a template defined by country.


axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -270,31 +276,35 @@ public String computeFullName(Address address) {
 
   @Override
   public String computeAddressStr(Address address) {
```
jmi-axelor, line 278: we don't need to recompute anything in this method, just return address.getFormattedFullName()

axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -341,4 +351,60 @@ public String getZipCode(Address address) {
     }
     return null;
   }
+
+  @Override
+  @Transactional(rollbackOn = {Exception.class})
+  public void setFormattedFullName(Address address) {
+        computeStringAddressTemplate(address);
+  }
```
jmi-axelor, line 359: I understand why you did that but I think it's ok to just delete computeStringAddressTemplate and put everything directly in this method

axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -341,4 +351,60 @@ public String getZipCode(Address address) {
...
+
+  @Override
+  @Transactional(rollbackOn = {Exception.class})
+  public int computeFormattedAddressForCountries(List<Long> countryIds) {
+        int totalAddressFormatted = 0;
+        AddressRepository addressRepository = Beans.get(AddressRepository.class);
```
jmi-axelor, line 365: don't inject the AddressRepository again, just use the addressRepo global variable

axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -341,4 +351,60 @@ public String getZipCode(Address address) {
...
+            totalAddressFormatted = totalAddressFormatted + addressList.size();
+            generateFormattedAddressForAddress(addressList);
+
+                    JPA.clear();
+
+                    offset = addressList.size();
+            addressList = query.fetch(FETCH_LIMIT, offset);
+          }
+
+                return totalAddressFormatted;
```
jmi-axelor, line 382: you should use a try catch and generate TraceBacks for each exception in this while loop
Also count the number of exceptions and return it with the number of formatted addresses. The user needs to know if errors happened.

axelor-base/src/main/java/com/axelor/apps/base/service/AddressServiceImpl.java
```diff
@@ -341,4 +351,60 @@ public String getZipCode(Address address) {
...
+          }
+
+        Map<String, Object> templatesContext = Maps.newHashMap();
+
+        Class<?> klass = EntityHelper.getEntityClass(address);
+        Context context = new Context(Mapper.toMap(address), klass);
+
+        templatesContext.put(klass.getSimpleName(), context.asType(klass));
+
+        String fullFormattedString = templates.fromText(content).make(templatesContext).render();
```
jmi-axelor, line 407: You should throw an AxelorException if an exception is throw during this compute. 
The AxelorException should inform that the AddressTemplate used has some errors

PR #11746, opened 2023-12-07
72664 - Fix NPE when we try to load no one payment line

paa-axelor:


ale-axelor:
merged in dev e7c68fb8473b4c9a8f9a0782911fa2f5c987e69c

PR #11776, opened 2023-12-15
RM#71027 Refacto BankStatementLineService

alo-axelor:


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankstatementline/afb120/BankStatementLineFetchAFB120ServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+      Optional<BankStatementLineAFB120> id =
+          bankPaymentBankStatementLineAFB120Repository
+              .all()
+              .filter(predicate)
+              .fetchStream()
+              .sorted(Comparator.comparing(BankStatementLineAFB120::getOperationDate))
+              .findFirst();
+      return id.isPresent()
+          ? bankPaymentBankStatementLineAFB120Repository.find(id.get().getId())
+          : null;
```
ale-axelor, line 59: please improve this using methods from Optional: you can use `map` on an optional to call `find` only if the line is present and either directly return the optional or `orElse(null)` to return null if empty.

ale-axelor (changes requested):

PR #11781, opened 2023-12-15
RM#72468 : Fix custom accounting report group by account

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/custom/AccountingReportValueAbstractService.java
```diff
@@ -198,7 +216,7 @@ protected int getColumnStyleSelect(
 
   protected List<String> getAccountFilters(
       Set<AccountType> accountTypeSet, String groupColumnFilter, boolean areAllAccountSetsEmpty) {
-    List<String> queryList = Arrays.asList("self.isRegulatoryAccount IS FALSE");
+    List<String> queryList = new ArrayList<>(Arrays.asList("self.isRegulatoryAccount IS FALSE"));
```
jmi-axelor, line 219: Hum why do you need to add this? It's exactly what is done in Arrays.asList

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/custom/AccountingReportValueAbstractService.java
aol-axelor: As stated in java doc, `Arrays.asList` will `Returns a fixed-size list backed by the specified array.`, so I need this to add new elements

PR #11839, opened 2024-01-08
Anomaly RM#72569:NPE at the manuf order generation from a sale order …

vva-axelor:
…on AOS 7.2.3


changelogs/unreleased/fix-npe-at-the-manuf-order-generation-from-sale-order.yml
kra-axelor, line 1: Please fix the changelog format, also may you change the title to "SaleOrder: Fixed a NPE that occured at the manuf order generation"

PR #11854, opened 2024-01-10
73581 - Fix debt recovery error when no invoice term was found

paa-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/debtrecovery/DebtRecoveryService.java
```diff
@@ -467,89 +468,94 @@ public boolean debtRecoveryGenerate(Partner partner, Company company, TradingNam
         debtRecovery.setBalanceDue(balanceDue);
 
         List<InvoiceTerm> invoiceTermList = this.getInvoiceTerms(partner, company, tradingName);
```
jmi-axelor, line 470: Instead of putting everything in another if and indent everything, can you just put after this line a : 
```
if (ObjectUtils.isEmpty(invoiceTermList)) {
    debtRecoverySessionService.debtRecoveryInitialization(debtRecovery);
}
```

And also put this line and the if before the `if (debtRecovery == null) {` because we create a debtRecovery for nothing if the invoiceTermList is empty

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/service/debtrecovery/DebtRecoveryService.java
paa-axelor: Resolved

paa-axelor (commented):
Resolved

jmi-axelor (approved):

PR #11860, opened 2024-01-10
72565 - Fix journal and company selection when user has no active com…

paa-axelor:
…pany


jmi-axelor (approved):

axelor-account/src/main/resources/views/Move.xml
```diff
@@ -426,11 +426,11 @@
         </field>
         <panel name="moveContentPanel" colSpan="12">
           <field name="company" canEdit="false" widget="SuggestBox"
-            readonlyIf="$validatePeriod || moveLineMassEntryList.length &gt; 0 || statusSelect == 3"
+            readonlyIf="moveLineMassEntryList.length &gt; 0 || statusSelect == 3"
             onChange="action-move-method-company-onchange" form-view="company-form"
             grid-view="company-grid" colSpan="3" onSelect="action-move-method-company-onselect"/>
           <field name="journal" domain="self.statusSelect = 1 AND self.company = :company"
-            readonlyIf="$validatePeriod || moveLineMassEntryList.length &gt; 0 || statusSelect == 3"
+            readonlyIf="($validatePeriod &amp;&amp; moveLineMassEntryList.length &gt; 0) || statusSelect == 3"
```
jmi-axelor, line 433: I don't think we have to change this, in move-form it's the same as before and the behavior is ok
Because now, when I have an active company and I add new lines, the journal is never in readonly

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #11868, opened 2024-01-10
RM#73948: LOCALIZATION : remove the Company.locale and use the localization instead

tyu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/CompanyDateService.java
```diff
@@ -48,7 +48,8 @@ public DateTimeFormatter getDateTimeFormat(Company company) throws AxelorExcepti
 
   protected Locale checkLocale(Company company) throws AxelorException {
     Locale locale = null;
-    String localeStr = Optional.ofNullable(company).map(Company::getLocale).orElse(null);
+    Localization localization = company.getLocalization();
+    String localeStr = localization.getCode();
```
jmi-axelor, line 52: you don't need to change that right?

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #11878, opened 2024-01-11
RM#74005: LOCALIZATION : language is still used to print an invoice

tyu-axelor:


axelor-account/src/main/resources/domains/AccountConfig.xml
```diff
@@ -179,8 +179,8 @@
       default="true"/>
     <boolean name="fixedAssetCatReqOnInvoice"
       title="Fixed asset category required on invoices"/>
-    <boolean name="isPrintInvoicesInCompanyLanguage"
-      title="Print invoices in company language" default="true"/>
+    <boolean name="isPrintInvoicesInCompanyLocalization"
+      title="Print invoices in company localization" default="true"/>
```
jmi-axelor, line 183: Just change the title, not the fields name, this is supposed to be an anomaly, we don't do schema changes in anomalies.

jmi-axelor (changes requested):

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceController.java
```diff
@@ -500,18 +500,22 @@ public void showInvoice(ActionRequest request, ActionResponse response) {
...
+        String localizationCode =
+            localizationMap != null && localizationMap.get("id") != null
+                ? Beans.get(LocalizationRepository.class)
+                    .find(Long.parseLong(localizationMap.get("id").toString()))
                     .getCode()
                 : null;
+        String locale = "en";
+        if (localizationCode != null) {
+          locale = localizationCode.substring(0, 2);
+        }
```
jmi-axelor, line 518: Just remove this code
We need the full locale and don't put a default value if it's null, if it's null it's already managed in InvoicePrintService.printInvoice method

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #11897, opened 2024-01-17
GlobalBudget/AdvancedExport : new management of advanced exports

rth-axelor:


axelor-budget/src/main/java/com/axelor/apps/budget/export/ExportBudgetCallableService.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.xssf.usermodel.XSSFWorkbook;
+
+public class ExportBudgetCallableService implements Callable<MetaFile> {
```
ale-axelor, line 56: this service is still injected, so you can inject dependency with the constructor instead of using `Beans.get`

ale-axelor (changes requested):

PR #11919, opened 2024-01-19
RM-73551 Adjust price and quantity precision

mea-axelor:


axelor-supplychain/src/main/resources/views/SaleOrderInvoicingWizard.xml
```diff
@@ -42,8 +42,9 @@
     <field name="product.code" width="120"/>
     <field name="productName" readonly="true"/>
     <field name="qty" readonly="true" aggregate="sum" x-scale="2"/>
-    <field name="unit" readonly="true" form-view="unit-form" grid-view="unit-grid"/>
-    <field name="price" readonly="true" x-scale="2"/>
+    <field name="unit" readonly="true" form-view="unit-form" grid-view="unit-grid"
+      x-scale="$nbDecimalDigitForQty"/>
+    <field name="price" readonly="true" x-scale="$nbDecimalDigitForUnitPrice"/>
```
paa-axelor, line 47: You already set a scale on `action-attrs-sale-order-invoicing-wizard-price-and-qty-precision`.
So we don't need a scale for those flieds.
Can you remove-it please

PR #11934, opened 2024-01-23
RM#74315 : Adding Holdbacks to Business Projects + holdbacks invoicing mechanism

mak-axelor:


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/HoldBackLineServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+            invoiceLines.add(invoiceLine);
+
+            return invoiceLines;
+          }
+        };
+
+    return invoiceLineGenerator.creates();
+  }
+
+  BigDecimal calculateHoldBackLinePrice(Invoice invoice, HoldBackLine holdBackLine) {
```
alo-axelor, line 79: This method should be protected.

PR #11936, opened 2024-01-23
74438 Bank Reconciliation: fix dates management in moveline query

rth-axelor:


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationService.java
```diff
@@ -925,29 +925,16 @@ public String getAccountDomain(BankReconciliation bankReconciliation) {
     return "self.id = 0";
   }
 
-  public String getRequestMoveLines(BankReconciliation bankReconciliation) {
+  public String getRequestMoveLines() {
```
jmi-axelor, line 928: Don't remove this parameter, you did not remove it in 7.2, we don't want to have differences
If we need to remove it, we'll do a ticket in 7.2

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #11943, opened 2024-01-24
RM#73909 : Add multi company support to custom account reports

aol-axelor:


axelor-account/src/main/resources/i18n/messages_fr.csv
```diff
@@ -1127,9 +1128,11 @@ e a été payée ou qu'elle est reliée à une session de paiement.</Badge></>",
 "Company currency code","Code devise société",,
 "Company currency decimals",,,
 "Company name",,,
+"Company number",,,
```
rth-axelor, line 1131: Missed translation

PR #11953, opened 2024-01-25
RM#74523 HR API - create new request to create an expense

dar-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/expense/ExpenseCreateServiceImpl.java
```diff
@@ -75,6 +78,28 @@ public Expense createExpense(
...
+        Optional.ofNullable(company).map(Company::getDefaultBankDetails).orElse(null);
+    Integer companyCbSelect =
+        Optional.ofNullable(employee).map(Employee::getCompanyCbSelect).orElse(null);
+
+    setExpenseInfo(company, employee, currency, bankDetails, null, companyCbSelect, expense);
+    expenseComputationService.compute(expense);
+
+    return expenseRepository.save(expense);
+  }
+
```
alo-axelor, line 102: You can create a new method using the other createExpense method, it will do more checks for bankDetails and Period.

PR #11963, opened 2024-01-26
RM#74618: INVOICE : printing feature doesn't work anymore

tyu-axelor:


axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceController.java
```diff
@@ -491,7 +491,7 @@ public void showInvoice(ActionRequest request, ActionResponse response) {
                         .collect(Collectors.toList()));
         fileLink = Beans.get(InvoicePrintService.class).printInvoices(ids);
         title = I18n.get("Invoices");
-      } else if (context.get("id") != null
+      } else if (context.get("_invoiceId") != null
```
jmi-axelor, line 494: this method is also called when we're in the invoices grid and we click on the button "showInvoiceBtn" on a line in the grid and now this button doesn't work. I think we need a OR to try to get "id" or "_invoiceId"

axelor-account/src/main/java/com/axelor/apps/account/web/InvoiceController.java
```diff
@@ -517,7 +517,7 @@ public void showInvoice(ActionRequest request, ActionResponse response) {
             Beans.get(InvoicePrintService.class)
                 .printInvoice(
                     Beans.get(InvoiceRepository.class)
-                        .find(Long.parseLong(context.get("id").toString())),
+                        .find(Long.parseLong(context.get("_invoiceId").toString())),
```
jmi-axelor, line 520: same here

jmi-axelor (changes requested):

jmi-axelor (approved):

PR #11998, opened 2024-01-31
RM#74709: Lead: sending mail error

tyu-axelor:


axelor-base/src/main/resources/views/GenerateMessageWizard.xml
```diff
@@ -3,6 +3,26 @@
...
 
+  <form name="generate-message-wizard-form" model="com.axelor.utils.db.Wizard"
+    id="base-generate-message-wizard-form" title="Generate message"
+    onNew="action-generate-message-group-on-new" extension="true">
+
+    <extend target="//panel[@name='mainPanel']/field[@name='language']">
+      <attribute name="name" value="localization"/>
+      <attribute name="title" value="Localization"/>
+      <attribute name="target" value="com.axelor.apps.base.db.Localization"/>
+      <attribute name="onChange" value="action-on-change-localization-template"/>
```
jmi-axelor, line 14: no just use replace tag and put directly the field localization, it's weird to do it that way

axelor-base/src/main/resources/views/GenerateMessageWizard.xml
```diff
@@ -3,6 +3,26 @@
...
+    </extend>
+    <extend target="//panel[@name='mainPanel']/spacer[@name='languageSpacer']">
+      <attribute name="name" value="localizationSpacer"/>
+    </extend>
```
jmi-axelor, line 18: you don't need to rename the spacer, it's not used anywhere else

jmi-axelor (changes requested):

PR #12000, opened 2024-02-01
Feature RM#74563: [STOCKLOCATIONLINEHISTORY] Add new fields

bsu-axelor:


.github/workflows/gradle.yml
kra-axelor, line 1: Please remove this change from the pr

kra-axelor (changes requested):

PR #12005, opened 2024-02-01
74776 change multi currency check in bank order management

rth-axelor:


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankorder/BankOrderLineService.java
```diff
@@ -408,7 +409,7 @@ public BigDecimal computeCompanyCurrencyAmount(BankOrder bankOrder, BankOrderLin
 
     Currency bankOrderCurrency = bankOrder.getBankOrderCurrency();
 
-    if (bankOrder.getIsMultiCurrency()) {
+    if (!Objects.equals(bankOrderCurrency, bankOrder.getCompanyCurrency())) {
```
aol-axelor, line 412: I think it should be replaced by a method `isMultiCurrency(BankOrder)` everytime this operation is done. It could be static inside a new `BankOrderToolService` like what we have for invoices with `InvoiceToolService.isMultiCurrency`

PR #12028, opened 2024-02-05
RM#74858: IMPORTEXPORTTRANSLATION : exported file is not found

tyu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/ImportExportTranslationServiceImpl.java
```diff
@@ -419,4 +426,19 @@ protected void saveNewTranslation(String key, String messageValue, String langua
...
+   * @throws IOException
+   */
+  @Override
+  public void copyFileFromUploadDirToExportDir(String uploadedFilePath) throws IOException {
+    Path EXPORT_PATH =
+        Paths.get(AppSettings.get().get(AvailableAppSettings.DATA_EXPORT_DIR, DEFAULT_EXPORT_PATH));
+    Path sourcePath = Paths.get(uploadedFilePath);
+    Path targetPath = EXPORT_PATH.resolve(sourcePath.getFileName());
+    Files.copy(sourcePath, targetPath, StandardCopyOption.REPLACE_EXISTING);
+  }
```
jmi-axelor, line 443: It would be better to put this method in FileExportTools because this is not a specific method for translations
Also it doesn't really copy a file from the upload directory, it just copies a file from the path in argument to the export directory. 
Just rename this method copyFileToExportDir

jmi-axelor (changes requested):

PR #12081, opened 2024-02-13
Anomaly RM#75335:TOUR:Hide validate button when everything is validated

cma-axelor:


alo-axelor (commented):
Since isValidated is a dummy field is please add '$' in front of the variable name, so we can't mistake it for a real variable.

PR #12106, opened 2024-02-16
RM#75415 Delete related mobile dashboard lines when deleting MobileDa…

alo-axelor:
…shboard


changelogs/unreleased/75415.yml
kra-axelor, line 1: Please fix the changelog format

PR #12115, opened 2024-02-16
75520 Pricing : add menu entry in application config

rth-axelor:


changelogs/unreleased/75520.yml
kra-axelor, line 1: Please remove the changelog

PR #12131, opened 2024-02-21
Anomaly RM#75712:[ABC CLASS] Everything is classed in the first class

vva-axelor:


changelogs/unreleased/fix-sequence-issue-in-abc-class.yml
kra-axelor, line 1: Please fix the changelog format of the ticket, also may you change the title to: "ABC CLASS: Fixed a bug where everything was classed in the first class"

PR #12136, opened 2024-02-22
Feature RM#75720:[COST CALCULATION] Add fields to export

vva-axelor:


changelogs/unreleased/feature-add-fields-to-cost-calculation-export.yml
kra-axelor, line 1: Please remove the changelog

PR #12159, opened 2024-02-23
RM#75857: ADDRESS : default country on a new address

tyu-axelor:


axelor-base/src/main/resources/views/Address.xml
```diff
@@ -191,10 +191,7 @@
   </action-method>
 
   <action-record name="default-address-record" model="com.axelor.apps.base.db.Address">
-    <field name="country" expr="eval: __user__?.activeCompany?.address?.country"
-      if="__user__?.activeCompany?.address"/>
```
jmi-axelor, line 195: please keep this

axelor-base/src/main/resources/views/Address.xml
```diff
@@ -191,10 +191,7 @@
...
-    <field name="country" expr="action:default-address-country-record"
-      if="!__user__?.activeCompany?.address"/>
```
jmi-axelor, line 197: Keep this but change the action default-address-country-record, it should not return France all the time but it should return the country on the user's localization

jmi-axelor (changes requested):

PR #12167, opened 2024-02-27
75067 FixedAsset : fix disposal wizard view and disposal sale move

rth-axelor:


paa-axelor (changes requested):
Also please update your PR to fix conflicts with dev branch

PR #12174, opened 2024-02-28
RM#65382: Add a supplier delivery time in supplier catalog

cce-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseOrderLineService.java
```diff
@@ -104,4 +105,6 @@ public void checkDifferentSupplier(
 
   public List<PurchaseOrderLine> updateLinesAfterFiscalPositionChange(PurchaseOrder purchaseOrder)
       throws AxelorException;
+
+  LocalDate getEstimatedReceiptDate(PurchaseOrderLine purchaseOrderLine, Partner supplierPartner);
```
kra-axelor, line 109: Hello,

Please may you return a Optional<LocalDate> which is better suited when the return value can be null

kra-axelor (changes requested):

PR #12184, opened 2024-02-29
RM#71777 : Add non deductible taxes

aol-axelor:


axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -493,133 +501,149 @@ public List<MoveLine> createMoveLines(
...
+      throws AxelorException {
+    int vatSystemSelect =
+        accountingSituationService.determineVatSystemSelect(accountingSituation, invoiceLineTax);
+
+    Account account =
+        taxAccountService.getAccount(
+            tax,
+            company,
+            move.getJournal(),
+            invoiceLineTax.getInvoiceLine().getAccount(),
```
jmi-axelor, line 587: There will be an NPE if I ventilate an old Invoice that has been created before the 8.1

jmi-axelor (changes requested):

PR #12231, opened 2024-03-07
RM#75078: MRP: take into account billOfMaterials used in propositions…

oma-axelor:
… for the MRP calculations

### DIFF
--- a/axelor-production/src/main/java/com/axelor/apps/production/service/MrpServiceProductionImpl.java
+++ b/axelor-production/src/main/java/com/axelor/apps/production/service/MrpServiceProductionImpl.java
@@ -74,6 +74,8 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.time.LocalDate;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
@@ -92,6 +94,7 @@ public class MrpServiceProductionImpl extends MrpServiceImpl {
   protected AppProductionService appProductionService;
 
   protected ProdProcessLineService prodProcessLineService;
+  protected final BillOfMaterialMrpLineService billOfMaterialMrpLineService;
 
   @Inject
   public MrpServiceProductionImpl(
@@ -120,7 +123,8 @@ public class MrpServiceProductionImpl extends MrpServiceImpl {
       ProductCompanyService productCompanyService,
       BillOfMaterialService billOfMaterialService,
       AppProductionService appProductionService,
-      ProdProcessLineService prodProcessLineService) {
+      ProdProcessLineService prodProcessLineService,
+      BillOfMaterialMrpLineService billOfMaterialMrpLineService) {
     super(
         mrpRepository,
         stockLocationRepository,
@@ -148,6 +152,7 @@ public class MrpServiceProductionImpl extends MrpServiceImpl {
     this.billOfMaterialService = billOfMaterialService;
     this.appProductionService = appProductionService;
     this.prodProcessLineService = prodProcessLineService;
+    this.billOfMaterialMrpLineService = billOfMaterialMrpLineService;
   }
 
   @Override
@@ -161,8 +166,21 @@ public class MrpServiceProductionImpl extends MrpServiceImpl {
     }
   }
 
-  // Manufacturing order AND manufacturing order need
+  @Override
+  protected void fillMrpLinesForProductMap(Map<Long, Integer> productMap) throws AxelorException {
+    super.fillMrpLinesForProductMap(productMap);
+    if (appProductionService.isApp("production")) {
+      this.createManufOrderMrpLines(productMap);
+      this.createMPSLines(productMap);
+    }
+  }
+
   protected void createManufOrderMrpLines() throws AxelorException {
+    this.createManufOrderMrpLines(this.productMap);
+  }
+
+  // Manufacturing order AND manufacturing order need
+  protected void createManufOrderMrpLines(Map<Long, Integer> productMap) throws AxelorException {
 
     MrpLineType manufOrderMrpLineType =
         mrpLineTypeService.getMrpLineType(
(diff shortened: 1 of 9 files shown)


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialServiceImpl.java
```diff
@@ -579,4 +582,32 @@ public List<BillOfMaterial> getSubBillOfMaterial(BillOfMaterial billOfMaterial)
 
     return new ArrayList<>();
   }
+
+  @Override
```
kra-axelor, line 586: Please return a Optional<BillOfMaterial>, it is even better since the method can return a null value

axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialService.java
```diff
@@ -94,4 +95,7 @@ List<BillOfMaterial> getAlternativesBOM(Product originalProduct, Company company
   int getPriority(BillOfMaterial billOfMaterial);
 
   List<BillOfMaterial> getSubBillOfMaterial(BillOfMaterial billOfMaterial);
+
+  BillOfMaterial getEligibleBillOfMaterialOfProductInMrpLine(MrpLine mrpLine, Product product)
```
kra-axelor, line 99: Please create a new service for BillOfMaterial operation related to MrpLine

axelor-production/src/main/java/com/axelor/apps/production/service/MrpServiceProductionImpl.java
```diff
@@ -443,60 +461,78 @@ protected void createProposalMrpLine(
       throws AxelorException {
 
     Company company = mrp.getStockLocation().getCompany();
-    BillOfMaterial defaultBillOfMaterial = billOfMaterialService.getDefaultBOM(product, company);
+
+    MrpLine originalMrpLine =
+        mrpLineOriginList.stream().findAny().map(MrpLineOrigin::getMrpLine).orElse(null);
+
+    BillOfMaterial billOfMaterial;
```
kra-axelor, line 468:
```java
 if (originalMrpLine != null) {
      if (originalMrpLine.getBillOfMaterial() != null) {
        billOfMaterial = originalMrpLine.getBillOfMaterial();
      } else {
        billOfMaterial =
            billOfMaterialService.getEligibleBillOfMaterialOfProductInMrpLine(
                originalMrpLine, product);
      }
    } else {
      billOfMaterial = billOfMaterialService.getDefaultBOM(product, company);
    }
```  
Isn't this equivalent of just calling` getEligibleBillOfMaterialOfProductInMrpLine `?

kra-axelor (changes requested):

axelor-production/src/main/java/com/axelor/apps/production/service/MrpServiceProductionImpl.java
```diff
@@ -443,60 +465,76 @@ protected void createProposalMrpLine(
       throws AxelorException {
 
     Company company = mrp.getStockLocation().getCompany();
-    BillOfMaterial defaultBillOfMaterial = billOfMaterialService.getDefaultBOM(product, company);
+
+    Optional<MrpLine> originalMrpLineOpt =
+        mrpLineOriginList.stream().findAny().map(MrpLineOrigin::getMrpLine);
+
+    BillOfMaterial billOfMaterial = null;
+    if (originalMrpLineOpt.map(MrpLine::getBillOfMaterial).isPresent()) {
```
kra-axelor, line 473: There is something weird with this bloc : 
```java
  if (originalMrpLineOpt.map(MrpLine::getBillOfMaterial).isPresent()) {
      billOfMaterial =
          billOfMaterialMrpLineService
              .getEligibleBillOfMaterialOfProductInMrpLine(originalMrpLineOpt.orElse(null), product)
              .orElse(null);
    }
    if (billOfMaterial == null) {
      billOfMaterial = billOfMaterialService.getDefaultBOM(product, company);
    }
```

You don't need to do this check `originalMrpLineOpt.map(MrpLine::getBillOfMaterial).isPresent()`, because `billOfMaterialMrpLineService
              .getEligibleBillOfMaterialOfProductInMrpLine` will already do the check and in this specific case  will always return an Optional.of(mrpLine.getBillOfMaterial) (since its present).
What is weird is that I don't understand why you call `getEligibleBillOfMaterialOfProductInMrpLine` in this case since this part of the method will never be called 

```java
    if (product == null) {
      return Optional.empty();
    }

    Optional<SaleOrderLine> saleOrderLineOpt =
        MrpLineTool.getOriginSaleOrderLineInMrpLineOrigin(mrpLine);
    Optional<BillOfMaterial> billOfMaterialOpt =
        saleOrderLineOpt.map(SaleOrderLine::getBillOfMaterial);

    if (billOfMaterialOpt.isPresent()
        && product.equals(billOfMaterialOpt.map(BillOfMaterial::getProduct).orElse(null))) {
      return billOfMaterialOpt;
    }
    return Optional.empty();
``` 
If this is intended then the method `getElgilibleBillOfMaterialOfProductInMrpLine` has a part of its code that will never be reached and therefore useless (since it seems it is the only place where the method is called)

In all cases, the bloc can be replaced with 
```java
    billOfMaterial = mrpLineOriginList
            .stream()
            .findAny()
            .map(MrpLineOrigin::getMrpLine)
            .flatMap(mrpLine -> billOfMaterialMrpLineService.getEligibleBillOfMaterialOfProductInMrpLine(mrpLine, product))
            .orElse(billOfMaterialService.getDefaultBOM(product, company));
``` 
(Need to remove the unreached part of the code)

However if you wish to keep the unreached code because you actually call the method somewhere elese then this can be replaced instead with

```java
    billOfMaterial = mrpLineOriginList
            .stream()
            .findAny()
            .map(MrpLineOrigin::getMrpLine)
            .map(MrpLine::getBillOfMaterial)
            .orElse(billOfMaterialService.getDefaultBOM(product, company));
```

kra-axelor (changes requested):

PR #12233, opened 2024-03-07
RM-75943: Add expense declaration

cce-axelor:


axelor-human-resource/src/main/resources/views/Expense.xml
```diff
@@ -198,7 +200,8 @@
         canNew="false" canSelect="false" canEdit="true" editable="true"
         form-view="employee-advance-usage-expense-form"
         grid-view="employee-advance-usage-expense-grid"/>
-      <panel name="accountInfoPanel" title="Account information">
+      <panel name="accountInfoPanel" title="Account information" hidden="true"
+        showIf="typeSelect=1">
```
ale-axelor, line 204: it is javascript, so it should be `==`

ale-axelor (changes requested):

axelor-human-resource/src/main/resources/views/Expense.xml
cce-axelor: Sorry, fixed !

PR #12253, opened 2024-03-12
76404 InvoicePayment : fix financial discount process when changing t…

rth-axelor:
…he payment date


paa-axelor (changes requested):
We need to wait the circular injection

PR #12254, opened 2024-03-13
Feature RM#76643: [PLANNING] Add a tool to generate events planning line

bsu-axelor:


axelor-base/src/main/resources/views/EventsPlanning.xml
```diff
@@ -10,11 +10,38 @@
...
       <field name="name"/>
     </panel>
     <panel-related name="eventsPlanningLineListPanel" field="eventsPlanningLineList"
       editable="true" grid-view="events-planning-line-grid" form-view="events-planning-line-form"/>
   </form>
 
+  <form name="add-period-wizard-form" model="com.axelor.utils.db.Wizard" title="Add a period"
+    onSave="action-method-generate-event-planning-lines">
+    <panel name="mainPanel">
+      <field name="$startDate" type="date" title="Start date" required="true"/>
```
kra-axelor, line 27: Hello,
Please do not use the old dirty check behavior `$` for dummy field, it is deprecated and will be removed in next major release of AOP. You may instead use `x-dirty`. (See https://redmine.axelor.com/news/3022)

PR #12259, opened 2024-03-14
Wip 76644

mah-axelor:


axelor-stock/src/main/resources/views/StockLocation.xml
```diff
@@ -125,11 +139,10 @@
           colSpan="3"/>
         <field name="isValued" showIf="typeSelect == 1 || typeSelect == 2" colSpan="3"/>
       </panel>
-
     </panel-tabs>
-
   </form>
 
+
   <form model="com.axelor.utils.db.Wizard" title="Select format to Export"
```
cce-axelor, line 146: Please remove restore original line breaks

PR #12286, opened 2024-03-19
76036: Tracking number configuration profile domains

kra-axelor:


axelor-stock/src/main/resources/views/Product.xml
```diff
@@ -140,8 +140,123 @@
...
+    <action if="__context__?._fromPurchase" name="action-product-attrs-readonly-purchasable"/>
+    <action name="action-product-record-set-barcode-type-config"/>
+    <action name="action-product-attrs-title-unit"/>
+    <action name="action-product-attrs-set-required-productCategory"/>
+    <action name="action-product-attrs-enable-product-company-list"/>
+    <action name="action-product-record-service-stock-managed"/>
+    <action name="action-product-attrs-set-cost-type-select"/>
+    <action name="action-product-attrs-set-required-procurement-method-select"/>
+    <action name="action-stock-product-attrs-set-scale-metric-mass-density"/>
+  </action-group>
```
ale-axelor, line 227: I think we should call a new action that calls the `action-group-base-product-onnew`, else this can become quickly difficult to maintain... same issue for onload

ale-axelor (changes requested):

PR #12289, opened 2024-03-20
RM#72990 Account management for YEB/YER contract

alo-axelor:


ale-axelor (changes requested):

axelor-contract/src/main/java/com/axelor/apps/contract/generator/InvoiceGeneratorContract.java
```diff
@@ -80,11 +79,25 @@ protected Invoice createInvoiceHeader() throws AxelorException {
...
+    int targetTypeSelect = contract.getTargetTypeSelect();
+
+    if (targetTypeSelect == ContractRepository.YEB_CUSTOMER_CONTRACT) {
+      invoice.setOperationTypeSelect(InvoiceRepository.OPERATION_TYPE_CLIENT_REFUND);
+    }
+
+    if (targetTypeSelect == ContractRepository.YEB_SUPPLIER_CONTRACT) {
+      invoice.setOperationTypeSelect(InvoiceRepository.OPERATION_TYPE_SUPPLIER_REFUND);
+    }
+  }
```
ale-axelor, line 102: I don't understand why we are replacing the type at the end, and not change directly when the operationTypeSelect is set line 38-40.

alo-axelor: I tried to use a method in line 38-40 but it was not possible. But I fixed it by directly accessing the operationType attribute.

PR #12294, opened 2024-03-20
wip-76388 Contract: CONTRACTLINE / CONSUMPTIONLINE : define max quant…

mah-axelor:


axelor-contract/src/main/java/com/axelor/apps/contract/service/ContractServiceImpl.java
```diff
@@ -898,4 +899,51 @@ public void setInitialPriceOnContractLines(Contract contract) {
...
+                            .getProduct()
+                            .getId()
+                            .equals(consumptionLineCtx.getProduct().getId())
+                        && !consumptionLine.getIsInvoiced())
+            .map(ConsumptionLine::getQty)
+            .reduce(BigDecimal.ZERO, BigDecimal::add);
+
+    sum = sum.subtract(initQt);
+    sum = sum.add(consumptionLineCtx.getQty());
+    return sum.compareTo(max) > 0;
```
cce-axelor, line 940: the case when there is no max quantity defined is not managed

mah-axelor: there is no need to manage it, since cosumptionMaxQuantity is defined in the domain with min=0, even if it's null and the request is done directly by a direct API call, the getter would return 0 not null, here is the getter:
public BigDecimal getConsumptionMaxQuantity() {
    return consumptionMaxQuantity == null ? BigDecimal.ZERO : consumptionMaxQuantity;
  }

cce-axelor: I meant functionnaly, we do not want to always have a max quantity

PR #12324, opened 2024-03-27
Wip 77126 StockLocation: improve automation on StockLocation forme

mah-axelor:


axelor-stock/src/main/resources/views/StockLocation.xml
```diff
@@ -240,6 +244,12 @@
       method="setStocklocationValue"/>
   </action-method>
 
+  <action-record name="action-stock-location-record-set-stock-location-value-Btn"
```
ale-axelor, line 247: please rename "action-stock-location-record-set-stock-location-value-Btn" => "action-stock-location-record-set-stock-location-value-btn" to follow our convention

axelor-stock/src/main/resources/views/StockLocation.xml
```diff
@@ -240,6 +244,12 @@
...
+    model="com.axelor.apps.stock.db.StockLocation">
+    <field name="$stockLocationValueBtn"
+      expr="eval: java.text.NumberFormat.getInstance(new Locale(__user__.localization?.code)).format(stockLocationValue)"/>
```
ale-axelor, line 250: the format does not work, please:

- to create the locale, use `Locale.forLanguageTag()` method and pass the `language` field to the method instead of the localization code
- Manage the case where the user language is empty by using Locale.US by default

maybe it would be interesting to create a method in a service annotated with `@CallMethod` to format a bigdecimal

axelor-stock/src/main/resources/views/StockLocation.xml
```diff
@@ -278,6 +289,51 @@
...
+
+  <action-record name="action-stock-location-record-empty-site"
+    model="com.axelor.apps.stock.db.StockLocation">
+    <field name="site" expr="eval: null"/>
+  </action-record>
+
+  <action-group name="action-stock-location-group-type-select-on-change">
+    <action name="action-stock-location-attrs-set-isOutsourcingLocation"/>
+    <action
+      name="action-stock-set-partner-domain-attrs,action-stock-location-attrs-set-is-valued"/>
```
ale-axelor, line 333: please separate the action in this action-group to improve readability

PR #12386, opened 2024-04-09
75855 - Rework Campaign service using axelor-message v3.0.0

paa-axelor:


libs.gradle
```diff
@@ -1,7 +1,7 @@
 def libs = ext.libs = [:]
 
 libs.axelor_studio = 'com.axelor.addons:axelor-studio:2.0.1'
-libs.axelor_message = 'com.axelor.addons:axelor-message:2.0.0'
+libs.axelor_message = 'com.axelor.addons:axelor-message:3.0.0'
 libs.axelor_utils = 'com.axelor.addons:axelor-utils:2.0.0'
```
dtr-axelor, line 5: shouldn't you update the 3 libs to 3.0.0? I think the major version are supposed to stay aligned

PR #12388, opened 2024-04-09
wip-74083 Partner: add a grid with linked contracts

mah-axelor:


axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -480,6 +480,10 @@
           onClick="save,action-partner-view-create-opportunity" link="" colSpan="3"/>
         <panel-include view="incl-opportunities-partner-form"/>
       </panel>
+      <panel name="contractsPanel" title="Contracts" showIf="id != null"
```
cce-axelor, line 483: you can use just showIf="id"

axelor-base/src/main/resources/views/Partner.xml
```diff
@@ -480,6 +480,10 @@
...
+        if="__config__.app.isApp('contract')">
+        <panel-include view="incl-contracts-partner-form"/>
+      </panel>
```
cce-axelor, line 486: Maybe it's not done like this for the other, but when it's possible, please use extension. You have no access to contract module in base so it's better to extend partner-form in Contract module.

PR #12390, opened 2024-04-09
Feature RM#40963: PRODUCTION : Manufacturing Order simulation for vis…

bsu-axelor:
…ualization of missing components

### DIFF
--- a/axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderServiceImpl.java
+++ b/axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderServiceImpl.java
@@ -94,6 +94,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.validation.ValidationException;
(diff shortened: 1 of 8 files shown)


axelor-production/src/main/java/com/axelor/apps/production/service/manuforder/ManufOrderServiceImpl.java
```diff
@@ -1323,4 +1324,48 @@ public void checkApplicableManufOrder(ManufOrder manufOrder) throws AxelorExcept
...
+                Collectors.groupingBy(
+                    BillOfMaterialLine::getProduct,
+                    Collectors.reducing(
+                        BigDecimal.ZERO, BillOfMaterialLine::getQty, BigDecimal::add)));
+
+    for (Entry<Product, BigDecimal> billOfMaterialLine : bomLineMap.entrySet()) {
+      Product product = billOfMaterialLine.getKey();
+      BigDecimal bomLineQty = billOfMaterialLine.getValue();
+      BigDecimal availableQty = productStockLocationService.getAvailableQty(product, company, null);
+      BigDecimal qtyNeeded = qty.multiply(bomLineQty).divide(bomQty);
```
ale-axelor, line 1357: I think we have to have a rounding for the `divide` call. It should work if we set the scale to appBaseService.getNbDecimalDigitForQty(), since it is used after in a subtraction.

ale-axelor (changes requested):

PR #12399, opened 2024-04-10
Wip 77317 SupplyChain: saleOrderLines Add a new menu entry

mah-axelor:


axelor-supplychain/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -295,6 +328,10 @@
       expr="eval: &quot; self.company.id = ${__parent__?.company.id} &quot;"/>
   </action-attrs>
 
+  <!-- ACTION VALIDATE -->
+
+
+
```
ale-axelor, line 334: can be removed

axelor-supplychain/src/main/resources/views/Menu.xml
```diff
@@ -277,5 +277,19 @@
...
+  <menuitem name="sc-root-sale-order-lines-to-invoice"
+    title="Mass sale order line invoicing" parent="sc-root-sale" order="601"
+    action="sc.root.sale.order.lines.to.invoice"/>
+
+  <action-view name="sc.root.sale.order.lines.to.invoice"
+    title="Mass sale order line invoicing" model="com.axelor.apps.sale.db.SaleOrderLine">
+    <view type="grid" name="sale-order-line-to-invoice-grid"/>
+    <view type="form" name="sale-order-line-form"/>
+    <domain>
+      (self.invoiced = false OR self.amountInvoiced &lt; self.qty * self.price) AND
```
ale-axelor, line 290: "invoiced" is used only for special process (it will probably be removed in current or next version), and directly checking the amountInvoiced does not manage all cases, so please instead use invoicingState
```suggestion
      (self.invoicingState = 1) AND
```

ale-axelor (changes requested):

PR #12449, opened 2024-04-18
77894: Manual tracking number

kra-axelor:


axelor-stock/src/main/resources/views/InventoryLine.xml
```diff
@@ -56,9 +56,11 @@
         onChange="action-inventory-line-group-update-inventory-line"/>
       <field name="price" hideIf="isPresentInStockLocation"/>
       <field name="isPresentInStockLocation" type="boolean" x-dirty="false" hidden="true"/>
+      <field name="inventory" readonly="true"/>
       <button name="showDetailStockProduct" title="See stock details" colSpan="3"
         onClick="action-inventory-line-view-for-a-product-from-product-view"
         if="__config__.app.isApp('supplychain')" widget="info-button" icon="fa-cubes"/>
+
```
ale-axelor, line 63: it looks like a change done by mistake ? please revert it if it is the case

ale-axelor (changes requested):

axelor-stock/src/main/resources/views/InventoryLine.xml
kra-axelor: The space was added by mistake, but not the field

PR #12451, opened 2024-04-18
RM#77605 PROJECTPLANNINGTIME new selection

dar-axelor:


axelor-human-resource/src/main/resources/domains/PlannedTimeValue.xml
```diff
@@ -0,0 +1,19 @@
...
+  xsi:schemaLocation="http://axelor.com/xml/ns/domain-models http://axelor.com/xml/ns/domain-models/domain-models_7.0.xsd">
+
+  <module name="project" package="com.axelor.apps.project.db"/>
+
+  <entity name="PlannedTimeValue" cacheable="true">
+    <string name="name" unique="true">
+      <![CDATA[
+		        String name = "";
+	            if(plannedTime != null) name = plannedTime.toString();
+	            return name;
```
ale-axelor, line 13: this code is not formatted and can be simplified.
```suggestion
    return plannedTime != null ? plannedTime.toString() : "";
```

ale-axelor (changes requested):

PR #12455, opened 2024-04-19
Feature RM#77838: Printing template: activate translatable on printin…

bsu-axelor:
…g template name


axelor-base/src/main/java/com/axelor/utils/service/TranslationServiceBaseImpl.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package com.axelor.utils.service;
+
+import com.axelor.apps.base.service.LocaleService;
+import com.axelor.apps.base.service.user.UserService;
+import com.axelor.meta.db.repo.MetaTranslationRepository;
+import com.google.inject.Inject;
+
+public class TranslationServiceBaseImpl extends TranslationServiceImpl
+    implements TranslationBaseService {
```
ale-axelor, line 27: extending a service from AOP could prevent other modules to extend it, so if we can avoid it, it's better:
- Please do not extend TranslationServiceImpl but instead inject it
- rename this class `TranslationBaseServiceImpl`

ale-axelor (changes requested):

PR #12498, opened 2024-04-25
Wip 78148 Base: File object to replace ContractFile and EmployeeFile

mah-axelor:


axelor-contract/src/main/resources/views/File.xml
```diff
@@ -0,0 +1,102 @@
...
+
+  <action-record name="action-contract-record-set-file-type-select"
+    model="com.axelor.apps.base.db.File">
+    <field name="fileTypeSelect" expr="eval: _fileTypeSelect"/>
+  </action-record>
+
+  <action-record name="action-contract-record-set-contract-version"
+    model="com.axelor.apps.base.db.File">
+    <field name="contractVersion"
+      expr="eval: __repo__(Contract).find(contract?.id).currentContractVersion"/>
```
ale-axelor, line 56: I got a NPE here because contract was null, please add a check on contract

axelor-contract/src/main/resources/views/File.xml
```diff
@@ -0,0 +1,102 @@
...
+  </action-record>
+
+  <action-group name="action-contract-file-group-file-onload">
+    <action name="action-contract-record-set-contract-on-load"/>
+    <action name="action-base-file-group-file-onload"/>
+  </action-group>
+
+  <action-record name="action-contract-record-set-contract-on-load"
+    model="com.axelor.apps.base.db.File">
+    <field name="$contract" expr="eval: contractVersion.contract"/>
```
ale-axelor, line 98: please manage the case where contractVersion is null to avoid a NPE

ale-axelor (changes requested):

PR #12499, opened 2024-04-25
Anomaly RM#77826: [PurchaseOrder From SaleOrder] - When Catlog does n…

bsu-axelor:
…ot have a code or name, message is not comprehensive when generating PO

### DIFF
--- a/axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/PurchaseOrderLineServiceSupplyChainImpl.java
+++ b/axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/PurchaseOrderLineServiceSupplyChainImpl.java
@@ -103,12 +103,7 @@ public class PurchaseOrderLineServiceSupplyChainImpl extends PurchaseOrderLineSe
 
     PurchaseOrderLine purchaseOrderLine =
         super.createPurchaseOrderLine(
-            purchaseOrder,
-            saleOrderLine.getProduct(),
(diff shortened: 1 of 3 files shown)


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseOrderLineService.java
```diff
@@ -104,9 +104,6 @@ public void checkMultipleQty(
       ActionResponse response)
       throws AxelorException;
 
-  public String[] getProductSupplierInfos(
```
kra-axelor, line 107: Is there a reason for removing the public method ? I can't see a direct link with the origin of the anomaly (https://github.com/axelor/axelor-open-suite/pull/10161/files)  

Even if it is not used anymore in dev, might be the case in previous versions so it should be deleted in futur minor versions. If the method need to be deleted in all cases and another method can be provided, please add developer section in the changelog to notify the devs of this change and the substitute method

kra-axelor (changes requested):

axelor-purchase/src/main/java/com/axelor/apps/purchase/service/PurchaseOrderLineService.java
bsu-axelor: Separate task is created in Redmine to remove this unused method: https://redmine.axelor.com/issues/79212

PR #12505, opened 2024-04-26
Anomaly RM#77367: Accounting Situation : Wrong scale on UsedCredit an…

bsu-axelor:
…d AcceptedCredit fields


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/SupplychainSaleConfigServiceImpl.java
```diff
@@ -30,6 +31,7 @@ public class SupplychainSaleConfigServiceImpl extends SaleConfigServiceImpl
     implements SupplychainSaleConfigService {
 
   @Inject private AccountingSituationRepository accountingSituationRepo;
+  @Inject private CurrencyScaleService currencyScaleService;
```
kra-axelor, line 34: May you change the visibility to protected please ?

kra-axelor (changes requested):

PR #12615, opened 2024-05-21
77334 - Fix NPE at move line creation when partner have financial dis…

paa-axelor:
…count setted on config


changelogs/unreleased/77334.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Move Line: fixed impossible to create a move line when partner have financial discount"
```
rth-axelor, line 2: Can you add in the changelog the fact that you changed the parameters of computeFinancialDiscount(MoveLine moveLine, Move move); please

PR #12623, opened 2024-05-22
wip-79260 BusinessProject: use priceDiscounted on contractFramework i…

mah-axelor:
…f discount exists


axelor-business-project/src/main/java/com/axelor/apps/businessproject/service/ProjectFrameworkContractServiceImpl.java
```diff
@@ -95,7 +95,7 @@ protected Map<String, Object> getProductDataFromCustomerContract(
     }
 
     ContractLine contractLine = contractLines.get(0);
-    productMap.put("unitPrice", contractLine.getPrice());
+    productMap.put("unitPrice", contractLine.getPriceDiscounted());
```
cce-axelor, line 98: You don't check if there is a discount ? Or is the priceDiscounted always set with the price is there is no discount ?

mah-axelor: yes, if there is no discount, priceDiscounted is equal to price

PR #12789, opened 2024-06-17
80248 add warning message at close/open accounting batch related to d…

rth-axelor:
…aybook moves


axelor-account/src/main/java/com/axelor/apps/account/service/batch/BatchCloseAnnualAccounts.java
```diff
@@ -544,4 +552,48 @@ public boolean equals(Object o) {
       }
     }
   }
+
+  public String checkDaybookMovesOnFiscalYear(AccountingBatch accountingBatch)
+      throws AxelorException {
+    String warning = "";
```
kra-axelor, line 558: Not a fan of returning a message as return value here as this is not exploitable elsewhere
I think it would be better to just return the daybookMoveCount or a boolean and set the warning message in the controller depending on the result

kra-axelor (changes requested):

PR #12899, opened 2024-07-05
81315 change project hold back from axelor-project to axelor-business…

rth-axelor:
…-project


axelor-sale/src/main/resources/views/SaleConfig.xml
```diff
@@ -17,7 +17,7 @@
       <field name="saleOrderInAtiSelect"/>
       <field name="defaultValidityDuration" domain="self.applicationTypeSelect = 1"
         form-view="duration-form" grid-view="duration-grid" colSpan="4"/>
-      <field name="defaultProjectHoldBackPercentage"/>
```
kra-axelor, line 20: Please do this change in another ticket, it is a anomaly that should be fixed in 8.1

kra-axelor (changes requested):

axelor-business-project/src/main/resources/views/Menu.xml
```diff
@@ -51,4 +51,15 @@
...
+  <!-- HOLDBACK -->
+  <menuitem name="project-project-hold-back" parent="project-configurations"
+    if="__config__.app.isApp('business-project')" if-module="axelor-business-project"
+    title="HoldBacks" action="project.root.project.hold.back.all" order="1400"/>
+
+  <action-view name="project.root.project.hold.back.all" title="HoldBacks"
+    model="com.axelor.apps.businessproject.db.ProjectHoldBack">
+    <view type="grid" name="project-hold-back-grid"/>
+    <view type="form" name="project-hold-back-form"/>
+  </action-view>
```
kra-axelor, line 63: Please rename menuitem and action view to match with the module name

kra-axelor (changes requested):

PR #13046, opened 2024-08-06
82765 manage tax by amount on saleOrder

rth-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorderline/SaleOrderLineCreateTaxLineServiceImpl.java
```diff
@@ -136,15 +144,61 @@ protected SaleOrderLineTax createSaleOrderLineTax(
...
+                    orderLineTaxService.isManageByAmount(saleOrderLineTax)
+                        && saleOrderLineTax
+                                .getTaxTotal()
+                                .compareTo(saleOrderLineTax.getPercentageTaxTotal())
+                            != 0)
+            .collect(Collectors.toList()));
+    return saleOrderLineTaxList;
+  }
+
+  protected SaleOrderLineTax isMapValueAlreadyInSaleOrderLineTaxList(
```
kra-axelor, line 187: Please rename this method, `is...()` or `has...()` method names suggest that it would return a `boolean` by convention (which is not the case here). In this case it could be `getSaleOrderLIneTaxFromList`

kra-axelor (changes requested):

PR #13048, opened 2024-08-06
82868 change isShowStatus into a new select

rth-axelor:


axelor-project/src/main/java/com/axelor/apps/project/service/ProjectToolServiceImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+  protected AppProjectService appProjectService;
+
+  @Inject
+  public ProjectToolServiceImpl(AppProjectService appProjectService) {
+    this.appProjectService = appProjectService;
+  }
+
+  @Override
+  public TaskStatus getCompletedTaskStatus(Project project) {
```
kra-axelor, line 24: Please return a Optional here which is easier to manage than null value

axelor-project/src/main/java/com/axelor/apps/project/service/ProjectTaskToolServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ProjectTaskToolServiceImpl implements ProjectTaskToolService {
+
+  public ProjectTaskToolServiceImpl() {}
+
+  @Override
+  public TaskStatus getCompletedTaskStatus(
```
kra-axelor, line 14: Please return a Optional here

axelor-project/src/main/java/com/axelor/apps/project/service/ProjectTaskToolServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+    if (!ObjectUtils.isEmpty(taskStatusSet)) {
+      completedTaskStatus = defaultTaskStatus;
+      if (completedTaskStatus == null || !taskStatusSet.contains(completedTaskStatus)) {
+        completedTaskStatus = null;
+      }
+
+      if (completedTaskStatus == null) {
+        List<TaskStatus> completedTaskStatusList =
+            taskStatusSet.stream().filter(TaskStatus::getIsCompleted).collect(Collectors.toList());
+        if (!ObjectUtils.isEmpty(completedTaskStatusList) && completedTaskStatusList.size() == 1) {
```
kra-axelor, line 27: Just to be sure, if there are more than 1 completed status, currently a null value would be return, is that intended ?  Shouldn't a random TaskStatus be returned, or a exception be thrown ?

kra-axelor (changes requested):

PR #13059, opened 2024-08-09
RM#82321 Loyalty account Manage company and trading names config

dar-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/loyalty/LoyaltyAccountServiceImpl.java
```diff
@@ -18,4 +43,21 @@ public Optional<LoyaltyAccount> getLoyaltyAccount(Partner partner, Company compa
...
+              .filter(account -> tradingName.equals(account.getTradingName()))
+              .findFirst();
+    }
+    return loyaltyAccount;
+  }
+
+  protected Optional<LoyaltyAccount> getLoyaltyAccount(Partner partner) {
+    return Optional.ofNullable(partner)
+        .map(Partner::getLoyaltyAccountList)
+        .map(account -> account.get(0));
```
ale-axelor, line 61: I think this method will not work if partner exists but has an empty loyalty account list.
Please check, but you should be able to replace this line by 

```suggestion
        .stream().findFirst();
```

ale-axelor (changes requested):

PR #13084, opened 2024-08-21
RM-83169 Manage exTaxes and inTaxes in API

mea-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/rest/dto/SaleOrderPostRequest.java
```diff
@@ -12,6 +12,8 @@ public class SaleOrderPostRequest extends RequestPostStructure {
   @Min(0)
   private Long clientPartnerId;
 
+  private String inAti;
```
alo-axelor, line 15: This should not be a String but a Boolean.

axelor-sale/src/main/java/com/axelor/apps/sale/service/SaleOrderGeneratorServiceImpl.java
```diff
@@ -94,4 +97,16 @@ protected void checkClientPartner(Partner clientPartner, SaleOrder saleOrder)
           company.getName());
     }
   }
+
+  protected void checkinAti(SaleOrder saleOrder, String inAti) throws AxelorException {
+    Boolean ati = Boolean.parseBoolean(inAti);
+    Company company = saleOrder.getCompany();
+    if (company.getSaleConfig().getSaleOrderInAtiSelect() == 1
```
alo-axelor, line 104: To fetch the company sale config, do not get it directly from the company. Use SaleConfigService#getSaleConfig. You should also check that inAti is not null since you should change it to a Boolean.

Instead of using hard values, use values from SaleConfigRepository.

axelor-sale/src/main/resources/i18n/messages_fr.csv
```diff
@@ -614,6 +614,7 @@
 "The client Partner is required and must be the same for all sale orders","Le client est requis et doit être le même pour toutes les cmdes clients",,
 "The client provided is not a customer of the given company: %s.","Le client fourni n'est pas un client de l'entreprise donnée: %s ",,
 "The company %s doesn't have any configured sequence for sale orders","La société %s n'a pas de séquence configurée pour les devis clients",,
+"The company configuration doesn't allow you to manage the exclusion or inclusion of taxes.","La configuration de l'entreprise ne permet pas de gérer l'exclusion ou l'inclusion des taxes.",,
```
alo-axelor, line 617: Change translation to :
La configuration de la société ne permet pas de gérer l'exclusion ou l'inclusion des taxes.

axelor-sale/src/main/java/com/axelor/apps/sale/service/SaleOrderGeneratorServiceImpl.java
```diff
@@ -94,4 +97,16 @@ protected void checkClientPartner(Partner clientPartner, SaleOrder saleOrder)
...
+        || company.getSaleConfig().getSaleOrderInAtiSelect() == 2) {
+      throw new AxelorException(
+          TraceBackRepository.CATEGORY_INCONSISTENCY,
+          I18n.get(SaleExceptionMessage.ATI_CHANGE_NOT_ALLOWED));
+    }
+    saleOrder.setInAti(ati);
```
alo-axelor, line 110: Do not set the inAti in the check method. Create another method that check if inAti is not null then you can set it.

axelor-sale/src/main/java/com/axelor/apps/sale/service/SaleOrderGeneratorServiceImpl.java
```diff
@@ -94,4 +104,29 @@ protected void checkClientPartner(Partner clientPartner, SaleOrder saleOrder)
...
+    Company company = saleOrder.getCompany();
+
+    if (saleConfigRepository
+                .find(saleConfigService.getSaleConfig(company).getId())
+                .getSaleOrderInAtiSelect()
+            == 1
+        || saleConfigRepository
+                .find(saleConfigService.getSaleConfig(company).getId())
+                .getSaleOrderInAtiSelect()
+            == 2) {
```
alo-axelor, line 126: Please check SaleOrderInitValueService#getInAti to see how to use SaleConfigService.

axelor-sale/src/main/java/com/axelor/apps/sale/rest/SaleOrderRestController.java
```diff
@@ -51,15 +51,23 @@ public Response createSaleOrder(SaleOrderPostRequest requestBody)
...
+                inAti);
 
     return ResponseConstructor.buildCreateResponse(saleOrder, new SaleOrderResponse(saleOrder));
   }
+
+  private static Boolean getInAti(String inAti) {
+    if (inAti.equals("true")) return true;
+    if (inAti.equals("false")) return false;
+    return null;
+  }
```
alo-axelor, line 72: NPE occurs if we don't specify inAti in the request.

This method should not be in the controller, please move it in a service, SaleOrderRestService for example.
Also it does not need to be static.

axelor-sale/src/main/java/com/axelor/apps/sale/rest/dto/SaleOrderPostRequest.java
```diff
@@ -30,6 +30,8 @@ public Long getCurrencyId() {
     return currencyId;
   }
 
+  private String inAti;
```
alo-axelor, line 33: We can force the value of inAti to be "true" or "false" with the Pattern annotation.
Look at ManufOrderProductPostRequest for an example.

PR #13125, opened 2024-08-28
RM#83397: NON DEDUCTIBLE TAX : invoices

tyu-axelor:
Improve tax computation (non deductible tax) on invoices.


axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceLineServiceImpl.java
```diff
@@ -740,4 +755,24 @@ public Map<String, Map<String, Object>> setScale(InvoiceLine invoiceLine, Invoic
 
     return attrsMap;
   }
+
+  @Override
+  public String computeInvoiceLineTaxLineSetDomain(int operationTypeSelect) {
+    LocalDate todayDate = LocalDate.now();
+    String domain = "(self.endDate > :date OR self.endDate IS NULL) ";
+    if (operationTypeSelect == 3 || operationTypeSelect == 4) {
```
bal-axelor, line 763: Here it is better to use repository variables, so instead of 3 and 4 you should use InvoiceRepository.OPERATION_TYPE_CLIENT_SALE and InvoiceRepository.OPERATION_TYPE_CLIENT_REFUND

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceLineServiceImpl.java
```diff
@@ -740,4 +755,24 @@ public Map<String, Map<String, Object>> setScale(InvoiceLine invoiceLine, Invoic
 
     return attrsMap;
   }
+
+  @Override
+  public String computeInvoiceLineTaxLineSetDomain(int operationTypeSelect) {
+    LocalDate todayDate = LocalDate.now();
```
bal-axelor, line 761: appBaseService.getTodayDate could be used instead

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/InvoiceLineServiceImpl.java
```diff
@@ -740,4 +755,24 @@ public Map<String, Map<String, Object>> setScale(InvoiceLine invoiceLine, Invoic
...
+      domain += " AND self.tax.isNonDeductibleTax = false ";
+    }
+
+    try {
+      List<Long> taxLineIds =
+          JPA.em()
+              .createQuery(" select self.id from TaxLine self where " + domain, Long.class)
+              .setParameter("date", todayDate)
+              .getResultList();
+      return "self.id in (" + Joiner.on(",").join(taxLineIds) + ")";
```
bal-axelor, line 773: It is not necessary to compute a list of ids, returning the where clause should be enough, so you could simply return "(self.endDate > :__date__ OR self.endDate IS NULL) " or "(self.endDate > :__date__ OR self.endDate IS NULL)  AND self.tax.isNonDeductibleTax = false " depending on the operationTypeSelect

axelor-account/src/main/java/com/axelor/apps/account/service/invoice/generator/tax/TaxInvoiceLine.java
```diff
@@ -274,22 +301,30 @@ protected List<InvoiceLineTax> finalizeInvoiceLineTaxes(
...
-                BigDecimal.valueOf(100), AppBaseService.COMPUTATION_SCALING, RoundingMode.HALF_UP);
-
+      List<InvoiceLineTax> invoiceLineTaxList,
+      BigDecimal sumOfAllDeductibleRateValue,
+      BigDecimal sumOfAllNonDeductibleRateValue) {
+
+    Boolean isNonDeductibleTax = invoiceLineTax.getTaxLine().getTax().getIsNonDeductibleTax();
+    BigDecimal originalTaxRateValue = invoiceLineTax.getTaxLine().getValue();
+    BigDecimal adjustedTaxValue =
+        Beans.get(TaxService.class)
```
bal-axelor, line 311: This injection should be done in the constructor to avoid doing it on each iteration of the loop

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -758,7 +760,15 @@ public MoveLine createMoveLineForAutoTax(
                 move.getJournal(), move.getOrigin(), move.getDescription())));
     moveLineToolService.setDecimals(newOrUpdatedMoveLine, move);
 
-    BigDecimal taxLineValue = taxLine.getValue();
+    Boolean isNonDeductibleTax = taxLine.getTax().getIsNonDeductibleTax();
+    BigDecimal originalTaxRateValue = taxLine.getValue();
+    BigDecimal adjustedTaxValue =
+        Beans.get(TaxService.class)
```
bal-axelor, line 766: TaxService is already injected in the constructor, you can use the variable taxService instead of doing another injection with Beans.get

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineCreateServiceImpl.java
```diff
@@ -767,10 +777,8 @@ public MoveLine createMoveLineForAutoTax(
     }
 
     int scale = currencyScaleService.getCompanyScale(move);
-    BigDecimal newMoveLineDebit =
-        debit.multiply(taxLineValue).divide(BigDecimal.valueOf(100), scale, RoundingMode.HALF_UP);
-    BigDecimal newMoveLineCredit =
-        credit.multiply(taxLineValue).divide(BigDecimal.valueOf(100), scale, RoundingMode.HALF_UP);
+    BigDecimal newMoveLineDebit = debit.multiply(adjustedTaxValue);
+    BigDecimal newMoveLineCredit = credit.multiply(adjustedTaxValue);
```
bal-axelor, line 782: These values should be scaled with the same scale and rounding as before

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
```diff
@@ -464,4 +470,28 @@ protected List<MoveLine> getMoveLinesFromIds(List<Integer> moveLineIds) {
     }
     return moveLineList;
   }
+
+  @Override
+  public String computeMoveLineTaxLineSetDomain(int functionalOriginSelect, String dateString) {
+    String[] split = dateString.split("-");
+    LocalDate date =
+        LocalDate.of(
+            Integer.parseInt(split[0]), Integer.parseInt(split[1]), Integer.parseInt(split[2]));
```
bal-axelor, line 479: It would be better not to use a string value for the date. You can get the LocalDate value directly from the parentContext of the moveLine by using context.getParent().asType(Move.class), and move.getDate().

And if you want to parse a string date, it is better to use something like LocalDate.parse(dateString) or LocalDate.parse(dateString, DateTimeFormatter.ofPattern(pattern)) (with "pattern" being "yyyy-MM-dd" or any other pattern)

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
```diff
@@ -464,4 +470,28 @@ protected List<MoveLine> getMoveLinesFromIds(List<Integer> moveLineIds) {
...
+
+    String domain = "(self.endDate > :date OR self.endDate IS NULL) ";
+    if (functionalOriginSelect == 3) {
```
bal-axelor, line 482: Same as before, the repository variable MoveRepository.FUNCTIONAL_ORIGIN_SALE should be used instead of 3

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
```diff
@@ -464,4 +470,28 @@ protected List<MoveLine> getMoveLinesFromIds(List<Integer> moveLineIds) {
...
+      domain += " AND self.tax.isNonDeductibleTax = false ";
+    }
+
+    try {
+      List<Long> taxLineIds =
+          JPA.em()
+              .createQuery(" select self.id from TaxLine self where " + domain, Long.class)
+              .setParameter("date", date)
+              .getResultList();
+      return "self.id in (" + Joiner.on(",").join(taxLineIds) + ")";
```
bal-axelor, line 492: Same as before, returning a list of ids works but is not necessary, returning there where clause is enough

axelor-base/src/main/java/com/axelor/apps/base/service/tax/TaxService.java
```diff
@@ -192,4 +199,102 @@ public Set<Tax> getTaxSet(Set<TaxLine> taxLineSet) {
...
+      if (isNonDeductibleTax) {
+        countNonDeductibleTaxes++;
+      } else {
+        countDeductibleTaxes++;
+      }
+    }
+    if (countDeductibleTaxes == 0 && countNonDeductibleTaxes > 0) {
+      return false;
+    }
+    return true;
```
bal-axelor, line 241: This would be easier to do it with something like 
!taxes.stream().allMatch(Tax::getIsNonDeductibleTax)

axelor-account/src/main/java/com/axelor/apps/account/db/repo/InvoiceManagementRepository.java
```diff
@@ -50,6 +55,8 @@ public Invoice copy(Invoice entity, boolean deep) {
   @Override
   public Invoice save(Invoice invoice) {
     try {
+      invoiceValidationService.checkNotOnlyNonDeductibleTaxes(invoice);
+      invoiceValidationService.checkSumOfNonDeductibleTaxes(invoice);
```
bal-axelor, line 59: I think it would be better to do these checks inside the method computeInvoice from InvoiceGenerator

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
tyu-axelor: This method is not used in this ticket, so there is no way to test it. I will put this modification in RM#83398.

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
tyu-axelor: This method is not used in this ticket, so there is no way to test it. I will put this modification in RM#83398.

axelor-account/src/main/java/com/axelor/apps/account/service/moveline/MoveLineServiceImpl.java
tyu-axelor: This method is not used in this ticket, so there is no way to test it. I will put this modification in RM#83398.

axelor-base/src/main/java/com/axelor/apps/base/service/tax/TaxService.java
```diff
@@ -192,4 +199,91 @@ public Set<Tax> getTaxSet(Set<TaxLine> taxLineSet) {
...
+    }
+    BigDecimal sumOfNonDeTaxes = BigDecimal.ZERO;
+    for (TaxLine taxLine : taxLines) {
+      if (Boolean.TRUE.equals(
+          Optional.of(taxLine).map(TaxLine::getTax).map(Tax::getIsNonDeductibleTax).orElse(null))) {
+        sumOfNonDeTaxes = sumOfNonDeTaxes.add(taxLine.getValue());
+        if (sumOfNonDeTaxes.compareTo(BigDecimal.valueOf(100)) > 0) {
+          throw new AxelorException(
+              TraceBackRepository.CATEGORY_INCONSISTENCY,
+              BaseExceptionMessage.SUM_OF_NON_DEDUCTIBLE_TAXES_EXCEEDS_ONE_HUNDRED);
```
jmi-axelor, line 284: I18n is missing, make sure it's the only place where it's missing

jmi-axelor (changes requested):

PR #13155, opened 2024-09-03
RM#82963: ADDRESS : the City M2O field must come along with the townName as in some cases it might differ

tyu-axelor:
Manage the views of Address. When there is a tonwName, check if the country have cities and then decide whether to show city field.


axelor-base/src/main/java/com/axelor/apps/base/web/AddressController.java
```diff
@@ -353,8 +353,38 @@ public void getAddressMetaField(ActionRequest request, ActionResponse response)
...
+            response.setAttr("townNameAndCityPanel", "hidden", false);
+            response.setAttr("cityPanel", "hidden", true);
+            response.setAttr("city", "hidden", true);
+          } else {
+            // only show cityPanel and keep the previous setting.
+            response.setAttr("townNameAndCityPanel", "hidden", true);
+            response.setAttr("cityPanel", "hidden", false);
+            response.setAttr("city", "hidden", cityHidden);
+          }
+        }
```
jmi-axelor, line 386: put all this in a service

jmi-axelor (changes requested):

PR #13376, opened 2024-10-03
RM#84509: Improve Address creation API (Country)

dar-axelor:


axelor-base/src/main/java/com/axelor/apps/base/service/address/CountryServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        country.setAlpha2Code(code);
+        break;
+      case 3:
+        if (!StringHelper.isDigital(code)) {
+          country.setAlpha3Code(code);
+        } else {
+          country.setNumericCode(code);
+        }
+    }
+    country.setAddressTemplate(addressTemplateRepository.findByName("DEFAULT"));
```
ale-axelor, line 38: please replace this by a config in app base

ale-axelor (changes requested):

PR #13540, opened 2024-11-04
Anomaly RM#85903: BankReconciliationService createDomainForMoveLine m…

bsu-axelor:
…ethod must not fetch directly Move line


axelor-bank-payment/src/main/java/com/axelor/apps/bankpayment/service/bankreconciliation/BankReconciliationDomainServiceImpl.java
```diff
@@ -168,15 +169,18 @@ public String getCashAccountDomain(BankReconciliation bankReconciliation) {
...
       throws AxelorException {
     String domain = "";
-    List<MoveLine> authorizedMoveLines =
+    String idList =
         moveLineRepository
             .all()
             .filter(bankReconciliationQueryService.getRequestMoveLines())
             .bind(bankReconciliationQueryService.getBindRequestMoveLine(bankReconciliation))
-            .fetch();
+            .fetch(0, 0)
```
kra-axelor, line 177: Hello,
Problem here is that the move lines are still being fetched entirely (then mappped to get the Id), we want to avoid that by directly fetching the id  using the Query.select method

PR #13545, opened 2024-11-05
Anomaly RM#83829: AnalyticMoveLineQuery : doesn't work with high data…

bsu-axelor:
… volumes


axelor-account/src/main/java/com/axelor/apps/account/service/analytic/AnalyticMoveLineQueryServiceImpl.java
```diff
@@ -108,82 +106,55 @@ public String getAnalyticMoveLineQuery(AnalyticMoveLineQuery analyticMoveLineQue
...
+                    paramMap.entrySet().stream()
+                        .allMatch(
+                            param ->
+                                entry.getValue().stream()
+                                    .anyMatch(
+                                        aml ->
+                                            aml.getAnalyticAxis().equals(param.getKey())
+                                                && param
+                                                    .getValue()
+                                                    .contains(aml.getAnalyticAccount()))))
```
ale-axelor, line 149: please improve readability for this code, this is too nested.

Also check if you can move checks from the java filter to the JPQL filter, it will improve performance.

ale-axelor (changes requested):

PR #13578, opened 2024-11-13
RM#86121 HR API : Request to change status

alo-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/leave/LeaveRequestMailService.java
```diff
@@ -25,9 +25,17 @@
...
 
+  void sendConfirmationEmailAndTrace(LeaveRequest leaveRequest) throws AxelorException;
+
   Message sendValidationEmail(LeaveRequest leaveRequest) throws AxelorException;
 
+  void sendValidationEmailAndTrace(LeaveRequest leaveRequest) throws AxelorException;
+
   Message sendRefusalEmail(LeaveRequest leaveRequest) throws AxelorException;
+
+  void sendRefusalEmailAndTrace(LeaveRequest leaveRequest) throws AxelorException;
```
ale-axelor, line 40: I don't understand why do we need to create these methods, can't we just catch the exception in LeaveRequestRestController to trace it ? Either way please refactor this, as it is duplicating code.

axelor-human-resource/src/main/resources/i18n/messages_fr.csv
```diff
@@ -625,6 +625,7 @@
 "Leave reason and day number have to be defined","Le motif de congé et le nombre de jour(s) pris doivent être définis",,
 "Leave request","Demande de congés",,
 "Leave request filters","Filtres demandes de congés",,
+"Leave request successfully updated.","Demande de congé mise à jour avec succès.",,
```
ale-axelor, line 628: (only to be consistent with existing translations)
```suggestion
"Leave request successfully updated.","Demande de congés mise à jour avec succès.",,

```

ale-axelor (changes requested):

PR #13580, opened 2024-11-14
86590 Move: reduce the time of mass reversing via the move's grid view

rth-axelor:


ale-axelor:
merged by mistake, reverted

PR #13611, opened 2024-11-19
RM#85224 Display tree grid in business project

alo-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorderline/SaleOrderLineContextHelper.java
```diff
@@ -19,15 +19,71 @@
...
+    saleOrder = saleOrderLine1.getSaleOrder();
+    if (saleOrder == null) {
+      saleOrder = getParentOrder(saleOrderLine1);
+    }
+    return saleOrder;
+  }
+
+  protected static SaleOrderLine getParentSol(Context context) {
+    SaleOrderLine saleOrderLine = context.asType(SaleOrderLine.class);
+    Long id = saleOrderLine.getId();
```
ale-axelor, line 61: since this method is recursive, if this method is called and no lines are persisted, we will get a NPE here when `context.getParent()` returns null

ale-axelor (changes requested):

PR #13727, opened 2024-12-18
Feature RM#88283:Retrieve partner information from the SIRET Number u…

nbu-axelor:
…sing the Sirene API


axelor-base/src/main/java/com/axelor/apps/base/service/partner/api/GenericApiFetchService.java
```diff
@@ -0,0 +1,76 @@
...
+public abstract class GenericApiFetchService {
+
+  public abstract String fetch(PartnerApiConfiguration configuration, String identifier)
+      throws AxelorException;
+
+  protected String getUrl(PartnerApiConfiguration configuration, String identifier) {
+    return configuration.getApiUrl() + "/" + identifier;
+  }
+
+  protected String treatResponse(HttpResponse<String> response, String identifier)
```
ale-axelor, line 30: I think this method is useless, as the response will depend on which API is used. So this method should be abstract.

axelor-base/src/main/java/com/axelor/apps/base/service/partner/api/PartnerApiFetchServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  protected Map<String, String> getHeaders(PartnerApiConfiguration configuration) {
+    Map<String, String> headers = new HashMap<>();
+    headers.put(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON);
+    headers.put(HttpHeaders.AUTHORIZATION, "Bearer " + configuration.getApiKey());
+    return headers;
+  }
+
+  protected String getUrl(PartnerApiConfiguration partnerApiConfiguration, String siretNumber) {
```
ale-axelor, line 50: either use the parent method, or set the parent method as abstract.

also please use `@Override` on methods whenever it's possible.

ale-axelor (changes requested):

PR #13782, opened 2025-01-02
Feature RM#88283: Update Partner form with data retrieved from Sirene API

nbu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/web/PartnerApiConfigurationController.java
```diff
@@ -26,11 +31,32 @@ public void fetchData(ActionRequest request, ActionResponse response) throws Axe
...
+    } else {
+      partner = new Partner();
+    }
+
     String result = Beans.get(PartnerApiFetchService.class).fetch(partnerApiConfiguration, siret);
 
     if (!StringUtils.isEmpty(result)) {
-      response.setInfo(result);
-      response.setCanClose(true);
+      Beans.get(PartnerApiCreateServiceImpl.class).setData(partner, result);
```
ale-axelor, line 45: * The interface should be injected, not the implementation.
* the controller should call a single method. so `PartnerApiCreateService` must have `PartnerApiFetchService` as a dependency.
* `PartnerApiCreateService` and the method `setData` should probably be renamed.

ale-axelor (changes requested):
please apply spotless

PR #13787, opened 2025-01-03
88552: Add new Type for purchase order for standard and subcontract

kra-axelor:


axelor-production/src/main/resources/views/PurchaseOrder.xml
```diff
@@ -43,4 +49,75 @@
...
+    <action name="action-purchase-order-method-fill-company-bank-details"/>
+    <action name="action-purchase-order-method-fill-price-list"/>
+    <action name="action-purchase-order-attrs-interco"/>
+    <action name="action-purchase-order-record-interco"/>
+    <action name="action-purchase-order-record-set-group-product-on-printing"/>
+    <action name="action-purchase-order-method-empty-fiscal-position-if-not-compatible"/>
+    <action name="action-purchase-order-method-fiscal-position-on-change"/>
+    <action name="action-purchase-order-method-compute"/>
+    <action name="action-purchase-order-record-set-type-select"/>
+  </action-group>
```
ale-axelor, line 69: the action group is production should call the the action group in purchase, else we are duplicating which actions must be called on partner change, which is difficult to maintain.

axelor-production/src/main/resources/views/PurchaseOrder.xml
```diff
@@ -43,4 +49,75 @@
...
+
+  <action-record name="action-purchase-order-record-set-type-select"
+    model="com.axelor.apps.purchase.db.PurchaseOrder">
+    <field name="typeSelect" expr="eval: 1"
+      if="!supplierPartner || !supplierPartner.isSubcontractor"/>
+    <field name="typeSelect" expr="eval: 2"
+      if="supplierPartner &amp;&amp; supplierPartner.isSubcontractor"/>
+  </action-record>
+
+  <search-filters name="purchase-order-filters"
```
ale-axelor, line 80: is there a way to avoid code duplication with search filters ? It would be nice because it is duplicating a lot...

axelor-purchase/src/main/resources/views/Menu.xml
```diff
@@ -67,13 +67,14 @@
...
 
+
   <action-view name="sc.root.purchase.quotations" title="Purchase quotations"
     model="com.axelor.apps.purchase.db.PurchaseOrder">
     <view type="cards" name="purchase-quotations-order-cards"/>
     <view type="grid" name="purchase-order-quotation-grid"/>
     <view type="form" name="purchase-order-form"/>
     <view-param name="search-filters" value="purchase-order-filters"/>
-    <domain>self.statusSelect &lt; 3</domain>
+    <domain>self.statusSelect &lt; 3 </domain>
```
ale-axelor, line 77: please remove all changes made to this file

ale-axelor (changes requested):

axelor-production/src/main/resources/views/PurchaseOrder.xml
kra-axelor: Removed duplication and override of this action, instead I directly override `action-purchase-order-record-set-type-select` which is what I needed (Don't know why I didn't do it from starter)

axelor-production/src/main/resources/views/PurchaseOrder.xml
kra-axelor: I don't know if there is a way, I tried to only add what I needed instead of duplicating but it does override everything so I can't do that.. Search-filter property on action view looks like it can only take 1 value also

PR #13789, opened 2025-01-06
RM#89178 Fix leave reason domain in leave request assistant

alo-axelor:


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/leavereason/LeaveReasonDomainServiceImpl.java
```diff
@@ -0,0 +1,63 @@
...
+    this.leaveReasonRepository = leaveReasonRepository;
+  }
+
+  @Override
+  public String getLeaveReasonDomain(LeaveReason leaveReason, Employee employee) {
+    return "self.id IN (" + StringHelper.getIdListString(getLeaveReasons(employee)) + ")";
+  }
+
+  protected Set<LeaveReason> getLeaveReasons(Employee employee) {
+    StringBuilder filter = new StringBuilder("self.leaveReasonTypeSelect = 2");
```
ale-axelor, line 32: I don't understand why we have to fetch leave reasons, find the ids, then build a domain instead of directly returning the base JPQL filter as a domain ? It seems like an unnecessary step.

Also since it is in java, we should use static variable for select values.

ale-axelor (changes requested):

PR #13847, opened 2025-01-16
89472 add kanban view in project for tasks by sprint

rth-axelor:


axelor-project/src/main/java/com/axelor/apps/project/service/roadmap/SprintServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public String getSprintIdsToExclude(List<Sprint> sprintList) {
+    String sprintIdsStr =
+        sprintList.stream()
+            .map(Sprint::getId)
+            .map(Object::toString)
+            .collect(Collectors.joining(","));
+    return sprintRepository
+        .all()
+        .filter(String.format("self.id NOT IN (%s)", sprintIdsStr))
```
ale-axelor, line 100: shouldn't it be filtered by project here ?

axelor-project/src/main/java/com/axelor/apps/project/service/roadmap/SprintServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+            ProjectRepository.SPRINT_MANAGEMENT_VERSION, project.getSprintManagementSelect())
+        && ObjectUtils.notEmpty(project.getRoadmapSet())) {
+      sprintList.addAll(
+          filterSprintsWithCurrentDate(
+              project.getRoadmapSet().stream()
+                  .map(ProjectVersion::getSprintList)
+                  .flatMap(Collection::stream)
+                  .collect(Collectors.toList()),
+              project.getCompany()));
+    }
```
ale-axelor, line 86: I think it's more clear here to create a method returning a sprint list from the project, managing both cases.

axelor-project/src/main/java/com/axelor/apps/project/service/roadmap/SprintService.java
```diff
@@ -16,14 +16,20 @@
...
 
 public interface SprintService {
 
   void generateBacklogSprint(Project project);
 
   void generateBacklogSprint(ProjectVersion projectVersion);
+
+  List<Sprint> getSprintToDisplay(Project project);
+
+  String getSprintIdsToExclude(List<Sprint> sprintList);
```
ale-axelor, line 34: these method should be moved in a separate service, they have nothing to do with the method to generateBacklogSprint.

Also they are made to be called one after the other. It would be simpler to only create a single method in the interface that return directly the right String.

axelor-project/src/main/java/com/axelor/apps/project/service/roadmap/SprintServiceImpl.java
rth-axelor: The filter by project is done before and the filter seems too big for a filter

PR #13854, opened 2025-01-17
89423: Configurator modify SOL

kra-axelor:


axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -57,14 +61,19 @@
...
   </grid>
 
   <grid name="sale-order-line-editable-grid" title="SO lines"
     model="com.axelor.apps.sale.db.SaleOrderLine" orderBy="sequence" editable="true"
     onNew="action-sale-order-line-attrs-set-compute-with-sol">
 
+
     <field name="sequence" if="__config__.app.getApp('sale')?.isDisplaySaleOrderLineNumber"
       width="51"/>
+    <button name="configuratorEditBtn" onClick="action-sale-order-line-view-regenerate-line"
```
ale-axelor, line 74: the form should be saved before the action is called

ale-axelor (changes requested):

PR #13857, opened 2025-01-20
89368: duplicate sale order line

kra-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/web/SaleOrderLineController.java
```diff
@@ -358,4 +359,18 @@ public void addToCart(ActionRequest request, ActionResponse response) {
       TraceBackService.trace(response, e);
     }
   }
+
+  public void configuratorDuplicateSaleOrderLine(ActionRequest request, ActionResponse response)
+      throws AxelorException {
+    Context context = request.getContext();
+    SaleOrderLine saleOrderLine =
+        Optional.ofNullable(context.asType(SaleOrderLine.class))
+            // .map(solCtx -> Beans.get(SaleOrderLineRepository.class).find(solCtx.getId()))
```
ale-axelor, line 368: I'm not sure about this being commented, it should save and get the sale order line from the db

axelor-sale/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -76,6 +80,10 @@
       onClick="save,action-sale-order-line-view-regenerate-line"
       if="__config__.app.getApp('sale')?.getEnableConfigurator()" readonlyIf="configurator == null"
       icon="fa-pen-to-square"/>
+    <button name="configuratorDuplicateBtn"
+      onClick="action-sale-order-line-method-configurator-duplicate"
```
ale-axelor, line 84: it seems that a save is missing here ? else when we call `setReload(true)` in the controller, it will remove every unsaved change in the form view.

ale-axelor (changes requested):

PR #13868, opened 2025-01-21
Feature RM#88824:Sprint - Avoid date overlap on sprint in the same pr…

vva-axelor:
…oject


axelor-project/src/main/java/com/axelor/apps/project/service/SprintServiceImpl.java
```diff
@@ -39,4 +41,31 @@ public void generateBacklogSprint(ProjectVersion projectVersion) {
...
+
+  protected boolean isOverlapping(Sprint sprint, Sprint otherSprint) {
+    if (sprint.getFromDate() == null
+        || sprint.getToDate() == null
+        || otherSprint.getFromDate() == null
+        || otherSprint.getToDate() == null) {
+      return false;
+    }
+    return sprint.getFromDate().isBefore(otherSprint.getToDate())
+        && sprint.getToDate().isAfter(otherSprint.getFromDate());
```
rth-axelor, line 69: It will only say that it's overlapping if sprint.fromDate < otherSprint.fromDate < otherSprint.toDate < sprint.toDate. 
We need to check that sprint.fromDate is between other sprint dates OR sprint.toDate is between other sprint dates indeed

PR #13916, opened 2025-01-30
RM#88925 Add production status in SoL form view

alo-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/SaleOrderLineDummyProductionServiceImpl.java
```diff
@@ -0,0 +1,121 @@
...
+      dummyFields.put(DUMMY_PRODUCTION_STATUS, PRODUCTION_STATUS_STANDBY);
+      return dummyFields;
+    }
+    if (statusSelectList.stream().anyMatch(status -> status == ManufOrderRepository.STATUS_DRAFT)) {
+      dummyFields.put(DUMMY_PRODUCTION_STATUS, PRODUCTION_STATUS_DRAFT);
+      return dummyFields;
+    }
+
+    return dummyFields;
+  }
```
ale-axelor, line 120: this service would be easier to read if we just had a protected method returning the correct PRODUCTION_STATUS_*. Also all PRODUCTION_STATUS variable could be defined inside an enum

ale-axelor (changes requested):

PR #14067, opened 2025-03-05
RM#91850 Prevent BoM edition if used in a SO

alo-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/BillOfMaterialDummyServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+      if (CollectionUtils.isNotEmpty(
+          saleOrderLineRepository
+              .all()
+              .filter(
+                  "self.billOfMaterialLine IS NOT NULL AND self.billOfMaterialLine = :billOfMaterialLine")
+              .bind("billOfMaterialLine", billOfMaterialLine)
+              .fetch())) {
+        return true;
+      }
+    }
```
ale-axelor, line 36: please run a single sql query with all bom ids instead of making one query per line

ale-axelor (changes requested):

PR #14142, opened 2025-03-24
RM-92365 Sale Order Line: Add the produced quantity on the sale order…

mea-axelor:
… line


axelor-production/src/main/resources/views/SaleOrderLine.xml
```diff
@@ -112,5 +121,8 @@
     <call class="com.axelor.apps.production.web.SaleOrderLineController"
       method="setProdProcessDomain"/>
   </action-method>
-
+  <action-method name="action-sale-order-line-method-display-qtyProduced">
+    <call class="com.axelor.apps.production.web.SaleOrderLineController"
+      method="showQtyProduced"/>
+  </action-method>
```
alo-axelor, line 127: This is not needed.

axelor-production/src/main/java/com/axelor/apps/production/service/StockMoveServiceProductionImpl.java
```diff
@@ -142,4 +149,23 @@ public void cancelFromManufOrder(StockMove stockMove, CancelReason cancelReason)
...
+              if (sml.getProducedManufOrder() != null
+                  && sml.getProducedManufOrder().getSaleOrderLine() != null) {
+                SaleOrderLine saleOrderLine = sml.getProducedManufOrder().getSaleOrderLine();
+                if (saleOrderLine.getProduct().equals(sml.getProduct()))
+                  saleOrderLine.setQtyProduced(sml.getQty());
+                saleOrderLineRepository.save(saleOrderLine);
+              }
+            });
+    return newStockSeq;
+  }
```
alo-axelor, line 170: You should not override this method but instead StockMoveServiceImpl#realizeStockMove which contains only the process linked to the realization. We had issues with transactions and mail services that is why we had to split the method in two.

axelor-production/src/main/java/com/axelor/apps/production/service/SaleOrderLineBomServiceImpl.java
```diff
@@ -98,4 +104,20 @@ public List<SaleOrderLine> createSaleOrderLinesFromBom(
...
+        manufOrderRepository
+            .all()
+            .filter("self.product.id=:productId and :saleOrderLineId =self.saleOrderLine.id")
+            .bind("saleOrderLineId", saleOrderLine.getId())
+            .bind("productId", saleOrderLine.getProduct().getId())
+            .fetchOne();
+
+    return saleOrder.getStatusSelect() >= SaleOrderRepository.STATUS_ORDER_CONFIRMED
+        && manufOrder != null;
+  }
```
alo-axelor, line 122: As said in the SaleOrderLine.xml comment, this code should be in a service SaleOrderViewProductionServiceImpl. 

Also there is a mistake in the method name, first letter should not be in uppercase.

ale-axelor (changes requested):

PR #14171, opened 2025-03-31
RM#92306 Fix PO generation from sale order

alo-axelor:


axelor-production/src/main/java/com/axelor/apps/production/service/productionorder/ProductionOrderSaleOrderServiceImpl.java
```diff
@@ -59,38 +66,127 @@ public ProductionOrderSaleOrderServiceImpl(
...
+      productionOrderIds.addAll(
+          productionOrderRepo
+              .all()
+              .filter("self.saleOrder = :saleOrder")
+              .bind("saleOrder", saleOrder)
+              .fetch()
+              .stream()
+              .map(ProductionOrder::getId)
+              .collect(Collectors.toList()));
+    }
```
ale-axelor, line 99: It looks like a lot of methods are starting with a check on `oneProdOrderPerSO`: maybe it would be easier to have different methods for when `oneProdOrderPerSO` is true or not, and check only once ?

PR #14190, opened 2025-04-02
92967: Call for tender init

kra-axelor:


axelor-purchase/src/main/resources/domains/CallTenderNeed.xml
```diff
@@ -0,0 +1,29 @@
...
+      scale="10"/>
+    <many-to-one name="unit" ref="com.axelor.apps.base.db.Unit" title="Unit"/>
+    <date name="requestedDate" title="Date"/>
+    <integer name="typeSelect" selection="purchase.call.tender.need.type.select"
+      title="Type"/>
+    <many-to-one name="callTender" ref="com.axelor.apps.purchase.db.CallTender"/>
+
+
+    <track on="UPDATE">
+      <field name="callTenderNeedList"/>
```
ale-axelor, line 22: it should be
```suggestion
      <field name="callTender"/>
```

I think

ale-axelor (changes requested):

PR #14192, opened 2025-04-02
Feature RM#91728 BANKRECONCILIATION : allow to select an analytic model

nbu-axelor:


axelor-bank-payment/src/main/resources/views/BankReconciliationLine.xml
```diff
@@ -107,41 +107,52 @@
       onChange="action-bank-reconciliation-line-unreconcile-unselected-reconcile-selected"
       onSelect="action-bank-reconciliation-line-method-move-line-domain" form-view="move-line-form"
       grid-view="move-line-grid"/>
+    <field name="analyticDistributionTemplate"/>
```
jmi-axelor, line 110: can you add a requiredIf="account.analyticDistributionRequiredOnMoveLines" and a readonlyIf="!account.analyticDistributionAuthorized" here? Everything is perfect after that

jmi-axelor (changes requested):

PR #14238, opened 2025-04-15
RM-93204 Added a sprint form for easier navigation and to disable sav…

mea-axelor:
…ing.


axelor-project/src/main/java/com/axelor/apps/project/web/SprintController.java
```diff
@@ -171,8 +171,7 @@ public void showAllSprints(ActionRequest request, ActionResponse response) {
     ActionView.ActionViewBuilder actionViewBuilder = ActionView.define(I18n.get("Sprints"));
     actionViewBuilder.model(Sprint.class.getName());
     actionViewBuilder.add("grid", "sprint-grid");
```
rth-axelor, line 173: Call another grid view that doesn't have the toolbar, can't create and can't remove please

PR #14376, opened 2025-05-21
Feature RM#95361: Move: single-select with color

bsu-axelor:


axelor-account/src/main/resources/views/Move.xml
```diff
@@ -33,21 +33,7 @@
...
-    <field name="reference" width="200"/>
-    <field name="journal" form-view="journal-form" grid-view="journal-grid" width="200"/>
-    <field name="functionalOriginSelect"/>
-    <field name="period" form-view="period-form" grid-view="period-grid" width="120"
-      canEdit="false" canView="false"/>
-    <field name="partner" form-view="partner-form" grid-view="partner-grid"/>
-    <field name="company" form-view="company-form" grid-view="company-grid" width="140"/>
-    <field name="invoice" form-view="invoice-form" grid-view="invoice-grid" width="140"/>
-    <field name="statusSelect" width="90"/>
+    <field name="statusSelect" width="90" widget="single-select"/>
```
ssb-axelor, line 36: I think you deleted this part without noticing. Could you check please ?

ssb-axelor (changes requested):

axelor-account/src/main/resources/views/Move.xml
bsu-axelor: I deleted this part because grid is not being used anywhere.

PR #14417, opened 2025-05-28
RM#95797 Manage sequence for quotation when quotation split

alo-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderSequenceServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+          company.getName());
+    }
+    return seq;
+  }
+
+  protected String computeChildrenOrderSequence(SaleOrder saleOrder, long counter) {
+    SaleOrder originSaleQuotation = saleOrder.getOriginSaleQuotation();
+    counter++;
+    return originSaleQuotation.getSaleOrderSeq() + "_" + counter;
+  }
```
ale-axelor: I don't get the purpose of this submethod, it is too small, seems to have side effect because it's incrementing a counter (but it's not the case)...

we can either:

- replace computeChildrenOrderSequence directly by the line originSaleQuotation.getSaleOrderSeq() + "_" + (childrenCounter + 1)
- use a submethod for the full case where `getIsQuotationAndOrderSplitEnabled` (so, line 38 to 50)

PR #14493, opened 2025-06-16
Feature RM#96543:Bank payment: Add check in XML import

vva-axelor:


axelor-bank-payment/src/main/resources/apps/demo-data/bank-payment-config.xml
```diff
@@ -5,14 +5,72 @@
 
   <input file="bankpayment_bankOrderEconomicReason.csv" separator=";"
     type="com.axelor.apps.bankpayment.db.BankOrderEconomicReason">
+    <bind column="code" to="code"
+      check="com.axelor.apps.base.service.imports.importer.ImportHelper.isValuePresent(name)"
```
cce-axelor: should be code instead of name

PR #14518, opened 2025-06-19
RM#96212 Add a order all button in wizard

alo-axelor:


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderSplitServiceImpl.java
```diff
@@ -144,6 +146,23 @@ public BigDecimal getQtyToOrderLeft(SaleOrderLine saleOrderLine) {
...
+      Map<String, Object> saleOrderLineMap =
+          Mapper.toMap(saleOrderLineRepository.copy(saleOrderLine, true));
+      saleOrderLineMap.put("$solId", saleOrderLine.getId());
+      saleOrderLineMap.put("$qtyToOrderLeft", getQtyToOrderLeft(saleOrderLine));
+      saleOrderLineMap.put("$nbDecimalDigitForUnitPrice", nbOfDecimalForPrice);
+      saleOrderLineMap.put("$nbDecimalDigitForQty", nbOfDecimalForQty);
+      saleOrderLineMapList.add(saleOrderLineMap);
+    }
+    return saleOrderLineMapList;
+  }
```
ale-axelor: I think it would be better to move this method to another service, since it's doing something related to the controller/views

PR #14519, opened 2025-06-19
96025: Hazard phrases

kra-axelor:


axelor-base/src/main/resources/views/Menu.xml
```diff
@@ -1227,4 +1227,13 @@
     <view type="form" name="base-hazard-pictogram-form"/>
   </action-view>
 
+  <menuitem name="base-hazard-phrase-all" parent="referential-root" title="Hazard phrases"
+    action="base.hazard.phrase.all" order="100"/>
```
alo-axelor, line 1231: The order "100" is already used for children of "referential-root", we should use a order after 600. Same for "base-icpe-all", I didn't notice it.

PR #14609, opened 2025-07-03
97254: Option to activate CFT

kra-axelor:


axelor-purchase/src/main/resources/views/Menu.xml
```diff
@@ -178,7 +178,8 @@
   </action-view>
 
   <menuitem name="purchase-root-call-for-tenders" title="Call for tenders" order="1200"
-    action="purchase.root.call.for.tenders" parent="sc-root-purchase"/>
+    action="purchase.root.call.for.tenders" parent="sc-root-purchase"
+    if="__config__.app.getApp('purchase')?.manageCallForTender"/>
```
gdu-axelor, line 182: no need to check the module ?

gdu-axelor (changes requested):

PR #14615, opened 2025-07-03
Anomaly RM#97047:Cost Calculation batch:Wrong calculation when Calcul…

cma-axelor:
…ation quantity from the BOM is > 1


axelor-production/src/main/java/com/axelor/apps/production/service/costsheet/UnitCostCalcLineService.java
```diff
@@ -23,9 +23,11 @@
 import com.axelor.apps.base.db.Product;
 import com.axelor.apps.production.db.CostSheet;
 import com.axelor.apps.production.db.UnitCostCalcLine;
+import java.math.BigDecimal;
 
 public interface UnitCostCalcLineService {
 
-  public UnitCostCalcLine createUnitCostCalcLine(
```
kra-axelor, line 29: Hello, please document the change of signature in the change log (developer part)

kra-axelor (changes requested):

PR #14640, opened 2025-07-09
Feature RM#91271:Company:Dark logo

cma-axelor:


axelor-base/src/main/resources/domains/MetaTheme.xml
```diff
@@ -0,0 +1,14 @@
...
+<domain-models xmlns="http://axelor.com/xml/ns/domain-models"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://axelor.com/xml/ns/domain-models http://axelor.com/xml/ns/domain-models/domain-models_7.4.xsd">
+
+  <module name="meta" package="com.axelor.meta.db"/>
+
+  <entity name="MetaTheme" table="META_THEME">
+
+    <integer name="logoModeSelect" title="Logo mode" selection="meta.theme.logo.mode.select"
+      default="1"/>
```
dtr-axelor: can't we use a real enumeration instead?

dtr-axelor: the list of logo modes are not supposed to be extensible imho

cma-axelor: I can add new Enum like ThemeLogoMode 
With 2 values LIGHT and DARK
What do you think ?

dtr-axelor: Yes that would be great 🙂

cma-axelor: `ThemeLogoMode` enum added with 3 values 
`NONE` to manage empty case

axelor-base/src/main/resources/views/Selects.xml
```diff
@@ -461,6 +461,11 @@
     <option value="com.axelor.apps.base.service.batch.ImportBatchService">com.axelor.apps.base.service.batch.ImportBatchService</option>
   </selection>
 
+  <selection name="meta.theme.logo.mode.select">
+    <option value="1">Light</option>
+    <option value="2">Dark</option>
+  </selection>
+
```
dtr-axelor: I think we can safely remove this selection now that we rely on a real enum

axelor-base/src/main/resources/domains/MetaTheme.xml
```diff
@@ -0,0 +1,18 @@
...
+
+  <entity name="MetaTheme" table="META_THEME">
+    <enum name="logoMode" ref="ThemeLogoMode" title="Logo mode" default="NONE"/>
+  </entity>
+
+  <enum name="ThemeLogoMode" numeric="true">
+    <item name="NONE" value="0"/>
+    <item name="LIGHT" value="1"/>
+    <item name="DARK" value="2"/>
+  </enum>
```
dtr-axelor: Why `numeric="true"`? It would make sense for an enum representing a month, a week day, a status, etc but for `ThemeLogoMode` I think we can use a string-enum and avoid these arbitrary numeric values.

PR #14683, opened 2025-07-21
Feature RM#84323: Base API - add request to create a Partner

bsu-axelor:


axelor-base/src/main/java/com/axelor/apps/base/rest/PartnerRestService.java
```diff
@@ -18,10 +18,24 @@
...
+      Integer titleSelect,
+      String firstName,
+      String name,
+      Partner mainPartner,
+      String description,
+      Boolean isContact,
+      Boolean isCustomer,
+      Boolean isSupplier,
+      Boolean isProspect)
+      throws AxelorException;
```
alo-axelor: I think this method could be moved in a new service like PartnerCreationService.

axelor-base/src/main/java/com/axelor/apps/base/rest/dto/PartnerPostRequest.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private String description;
+
+  private Boolean isContact = Boolean.FALSE;
+
+  private Boolean isCustomer = Boolean.FALSE;
+
+  private Boolean isSupplier = Boolean.FALSE;
+
+  private Boolean isProspect = Boolean.FALSE;
```
alo-axelor: You can use the boolean primitive here instead of the wrapper. I don't think there will be a case where it will be useful. Also, it will prevent having NPE. Don't forget to change every wrapper call when changing.

axelor-base/src/main/java/com/axelor/apps/base/rest/PartnerRestServiceImpl.java
```diff
@@ -43,4 +58,96 @@ public void addPartnerAddress(Partner partner, Address address) {
...
+
+    partner.setIsContact(isContact);
+    partner.setIsCustomer(isCustomer);
+    partner.setIsSupplier(isSupplier);
+    partner.setIsProspect(isProspect);
+  }
+
+  protected boolean getBooleanValue(Boolean value) {
+    return Boolean.TRUE.equals(value);
+  }
```
alo-axelor: By using the primitive boolean, this won't be necessary.

PR #14772, opened 2025-08-13
98828: Characteristique type

kra-axelor:


axelor-quality/src/main/resources/i18n/messages_fr.csv
```diff
@@ -46,6 +47,11 @@
 "Changing this value will change value of product, continue ?","Changer cette valeur changera la valeur de produit. Continuer ?",,
 "Characteristic","Caractéristique",,
 "Characteristic or sample does not belong to this control entry.","La caractéristique ou l'échantillon n'appartient pas à ce relevé de contrôle.",,
+"Characteristic properties","Propriété caractéristiques ",,
```
alo-axelor: I think this line shouldn't exist and "Characteristics properties" should be the only one for the plural form.

PR #14821, opened 2025-08-22
Feature RM#97784:Control Entry:add button to create quality improveme…

cma-axelor:
…nt from control entry


axelor-quality/src/main/java/com/axelor/apps/quality/service/QualityImprovementCreateServiceImpl.java
```diff
@@ -79,4 +95,9 @@ public QualityImprovement createQualityImprovement(
     }
     return qualityImprovement;
   }
+
+  public void setDefaultValues(QualityImprovement qualityImprovement) throws AxelorException {
```
kra-axelor: If public it needs to be declared in the Interface and add Overrides

If it is not intended, to make it protected

PR #14845, opened 2025-08-27
feat: add map viewer react application

lme-axelor:
RM#99945


axelor-base/src/main/java/com/axelor/apps/base/web/MapViewController.java
```diff
@@ -0,0 +1,43 @@
...
+import com.axelor.apps.base.db.MapView;
+import com.axelor.apps.base.db.repo.MapViewRepository;
+import com.axelor.apps.base.service.mapConfigurator.MapViewService;
+import com.axelor.inject.Beans;
+import com.axelor.rpc.ActionRequest;
+import com.axelor.rpc.ActionResponse;
+import com.google.inject.Singleton;
+import java.util.Objects;
+
+@Singleton
```
alo-axelor: I don't think the Singleton annotation is needed here.

PR #14927, opened 2025-09-16
feat: add legend management inside the header filter

gca-axelor:
RM#100758


axelor-base/map-viewer/src/components/MarkerGroup/MarkerGroup.tsx
```diff
@@ -42,12 +42,47 @@ const MarkerGroup = ({ id, name }: { id: number; name: string }) => {
...
 
+  const overlayName = useMemo(() => {
+    return `
+      <span data-map-group-id="${id}" style="display:none"></span>
+      <span>${name}</span>
+    `;
+  }, [name, id]);
+
+  useEffect(() => {
+    const color = config?.color ?? "#457896";
```
lme-axelor: Create a useMemo to compute color and use it in both the icon and checkbox config to avoid duplication

Also please create a const at the top to define default color : 
```
const DEFAULT_COLOR = "#457896";
```

PR #14934, opened 2025-09-17
90844 : Deposit slip: change condition for hiding a panel in deposit slip view and fix depositBankDetails field in payment voucher list grid

mou-axelor:


axelor-account/src/main/resources/views/DepositSlip.xml
```diff
@@ -228,7 +228,7 @@
     <field name="ref" readonly="true"/>
     <field name="valueForCollectionMove.journal" hidden="true"/>
     <field name="valueForCollectionMove.date" hidden="true"/>
-    <field name="depositBankDetails" readonly="true"/>
+    <field name="depositBankDetails" required="true"/>
```
ssb-axelor: Please also add this domain : 
`domain="self.active = true AND self.company = :company"`

ssb-axelor (changes requested):

PR #14943, opened 2025-09-17
RM-98626: add new default stock location feature

cce-axelor:


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineStockLocationServiceImpl.java
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * get default stock location by site (linked to stock move stock location)
+   *
+   * @param defaultStockLocationBySiteList
+   * @param stockMoveStockLocation
+   * @param mapDefaultStockLocation
+   * @return
+   */
+  public StockLocation getDefaultStockLocation(
```
alo-axelor: I think you wanted to set the method protected instead of public.

cce-axelor: Yes sorry !

PR #14953, opened 2025-09-19
Feature RM#99972: Production API : add new request to create a new co…

bsu-axelor:
…nsumed product on a operation order


axelor-production/src/main/java/com/axelor/apps/production/rest/dto/OperationOrderProductPostRequest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public TrackingNumber fetchTrackingNumber() {
+    if (trackingNumberId == null) {
+      return null;
+    }
+    return ObjectFinder.find(TrackingNumber.class, trackingNumberId, ObjectFinder.NO_VERSION);
+  }
+
+  public Product fetchProduct() {
+    return ObjectFinder.find(Product.class, productId, ObjectFinder.NO_VERSION);
```
alo-axelor, line 74: Add null check like you did in fetchTrackingNumber

bsu-axelor: productId is required field, so process will not go furthur if it's null during body validation.

PR #14970, opened 2025-09-23
Anomaly RM#94869:delivery panel missing for service in PO line

jak-axelor:


axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/purchaseorderline/view/PurchaseOrderLineViewServiceSupplychainImpl.java
```diff
@@ -0,0 +1,65 @@
...
+              .map(SupplyChainConfig::getHasInSmForStorableProduct)
+              .orElse(false);
+    }
+    if (productTypeSelect.equals(ProductRepository.PRODUCT_TYPE_SERVICE)) {
+      hidePanels =
+          !Optional.ofNullable(AuthUtils.getUser().getActiveCompany())
+              .map(Company::getSupplyChainConfig)
+              .map(SupplyChainConfig::getHasInSmForNonStorableProduct)
+              .orElse(false);
+    }
```
alo-axelor: I don't think there should be the same condition for storable product.

I think the condition should be more simple. We hide if product.productTypeSelect == 'service' && ! supplychainConfig.hasInSmForNonStorableProduct. Every other case we should not hide the panel.

axelor-supplychain/src/main/java/com/axelor/apps/supplychain/service/purchaseorderline/view/PurchaseOrderLineViewServiceSupplychainImpl.java
```diff
@@ -0,0 +1,58 @@
...
+import com.axelor.apps.purchase.db.PurchaseOrderLine;
+import com.axelor.apps.purchase.service.purchaseorderline.view.PurchaseOrderLineViewServiceImpl;
+import com.axelor.apps.supplychain.db.SupplyChainConfig;
+import com.axelor.auth.AuthUtils;
+import com.google.inject.Singleton;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+@Singleton
```
alo-axelor: This Singleton annotation is not necessary, please remove it.

PR #14979, opened 2025-09-25
100986

sos-axelor:


axelor-talent/src/main/resources/views/JobApplication.xml
```diff
@@ -254,6 +252,8 @@
       if="eval: _jobPositionId"/>
     <field name="responsible" expr="eval:__repo__(Employee).find(_responsibleId)"
       if="eval: _responsibleId"/>
+    <field name="hiringStage"
+      expr="eval:__repo__(HiringStage).all().filter('self.sequence = 1').fetchOne()"/>
```
alo-axelor: A config should be created for the default received hiring stage. What happens if the sequence is not correctly set ?

axelor-talent/src/main/java/com/axelor/apps/talent/db/repo/JobApplicationTalentRepository.java
```diff
@@ -39,4 +41,18 @@ public JobApplication save(JobApplication entity) {
...
+    super.remove(entity);
+
+    Beans.get(EventRepository.class)
+        .all()
+        .filter(
+            "relatedToSelect = ?1 AND self.relatedToSelectId = ?2",
+            JobApplication.class.getName(),
+            entity.getId())
+        .remove();
+  }
```
alo-axelor: Might be better to use a entity listener with the PreRemove annotation to not forget any case.  Just be careful to use the autoflush at false when using repository in listener.

PR #14996, opened 2025-09-29
RM#101175 CallTender PO Generation

alo-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/CallTenderPurchaseOrderServiceImpl.java
```diff
@@ -0,0 +1,146 @@
...
+            .collect(Collectors.toList());
+    for (Partner partner : partnerList) {
+      purchaseOrderList.add(
+          createPurchaseOrderFromOffer(callTender, partner, company, selectedCallTenderOfferList));
+    }
+
+    return purchaseOrderList;
+  }
+
+  protected PurchaseOrder createPurchaseOrderFromOffer(
```
cce-axelor, line 86: Maybe you can add a @Transactional here, it will be easier if an other method call it later

PR #15005, opened 2025-09-30
RM#101420 Fix QI buttons FR translation

alo-axelor:


axelor-production/src/main/resources/apps/demo-data/en/meta_metaTranslation.csv
```diff
@@ -0,0 +1,3 @@
+"key";"message";"language"
+"ProdProcessLine";"Phase";"fr"
+"OperationOrder";"Opération";"fr"
```
cce-axelor, line 3: Isn't it supposed to be in english ?

axelor-stock/src/main/resources/apps/demo-data/en/meta_metaTranslation.csv
```diff
@@ -0,0 +1,2 @@
+"key";"message";"language"
+"StockMoveLine";"Ligne de mouvement de stock";"fr"
```
cce-axelor, line 2: same here

cce-axelor: ok, normal

PR #15930, opened 2026-03-23
Anomaly RM#99916: FECIMPORT : Use company currency as default when Id…

cma-axelor:
…evise column is empty


axelor-account/src/main/java/com/axelor/apps/account/service/fecimport/FECImporter.java
```diff
@@ -144,7 +144,7 @@ protected void saveFecImport() {
   }
 
   protected void addMoveFromMoveLine(Model bean) {
-    if (bean.getClass().equals(MoveLine.class)) {
+    if (bean != null && bean.getClass() == MoveLine.class) {
```
rth-axelor: Can you let equals here, not just "==" if possible (the bean != null is OK)

PR #16038, opened 2026-04-07
RM-96909 [Maintenance] Maintenance Bills of Materials — Component Man…

mea-axelor:
…agement and Material Consumption from maintenance order


axelor-maintenance/src/main/java/com/axelor/apps/maintenance/service/CostSheetServiceMaintenanceImpl.java
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Override of CostSheetService for maintenance orders (typeSelect = 3).
+ *
+ * <p>For maintenance BOMs and ManufOrders, the "produced product" line is excluded from the cost
+ * sheet since maintenance does not produce a finished product. Only component, human resource, and
+ * machine cost lines are kept.
+ */
+public class CostSheetServiceMaintenanceImpl extends CostSheetServiceImpl {
+
+  protected boolean maintenanceContext;
```
alo-axelor: I think this boolean is not necessary and the process could be done in another way. Maybe refacto some methods for the parent class to avoid some duplication also.

axelor-maintenance/src/main/java/com/axelor/apps/maintenance/service/CostSheetServiceMaintenanceImpl.java
```diff
@@ -0,0 +1,332 @@
...
+      for (CostSheetLine child : new ArrayList<>(tempRoot.getCostSheetLineList())) {
+        child.setParentCostSheetLine(null);
+        costSheet.addCostSheetLineListItem(child);
+      }
+    }
+
+    BigDecimal costPrice = this.computeCostPrice(costSheet);
+    costSheet.setCostPrice(costPrice);
+    manufOrder.setCostPrice(costPrice);
+    Beans.get(ManufOrderRepository.class).save(manufOrder);
```
alo-axelor: I think you can add ManufOrderRepository as a parameter constructor in the parent class and use it here aswell. Or maybe it will be gone in the refacto to avoid duplication.

version.txt
alo-axelor, line 1: Remove the change on this file

PR #16073, opened 2026-04-15
Anomaly RM#111159:ACCOUNTING BATCH : Close/Open Accounts batch throws…

vva-axelor:
… SemanticException on run


axelor-account/src/main/java/com/axelor/apps/account/service/batch/BatchCloseAnnualAccounts.java
rth-axelor: Need an id here too

PR #16094, opened 2026-04-17
Feature RM#111107: HR : Add button to show accounting moves linked to…

cma-axelor:
… expense


axelor-human-resource/src/main/java/com/axelor/apps/hr/service/expense/ExpenseFetchMoveServiceImpl.java
```diff
@@ -48,4 +49,13 @@ public Move getExpensePaymentMove(Expense expense) {
         .bind("expenseId", expense.getId())
         .fetchOne();
   }
+
+  @Override
+  public List<Move> findAllMovesByExpense(Expense expense) {
+    return moveRepository
+        .all()
+        .filter("self.expense.id = :expenseId")
```
rth-axelor, line 57: Please add expensePayment too (payment moves are on this field)

axelor-human-resource/src/main/java/com/axelor/apps/hr/web/expense/ExpenseController.java
```diff
@@ -781,4 +782,37 @@ public void updateCurrencyOnLine(ActionRequest request, ActionResponse response)
...
+  public void showRelatedMoves(ActionRequest request, ActionResponse response) {
+    try {
+      Expense expense = request.getContext().asType(Expense.class);
+      expense = Beans.get(ExpenseRepository.class).find(expense.getId());
+      List<Move> moves = Beans.get(ExpenseFetchMoveService.class).findAllMovesByExpense(expense);
+
+      if (moves.isEmpty()) {
+        response.setInfo(I18n.get(HumanResourceExceptionMessage.EXPENSE_NO_MOVE_LINKED));
+      } else if (moves.size() == 1) {
+        response.setView(
```
rth-axelor, line 795: In the else if with the size, just put the context and the title then after it, set the response.setView with the same domain, context and showRecord only if size == 1. 
Maybe move it in a new service too, like ExpenseViewService
Then response.setView(expenseViewService.showMoves(expense.getId))
And in a try catch, throw an alert if catched

PR #16179, opened 2026-05-01
RM#110724 Call for tenders offers comparisons : charts

rch-axelor:


axelor-purchase/src/main/java/com/axelor/apps/purchase/service/CallTenderReportServiceImpl.java
```diff
@@ -0,0 +1,505 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+public class CallTenderReportServiceImpl implements CallTenderReportService {
+
+  public static final Inflector INFLECTOR = Inflector.getInstance();
+  private static final Map<Model, FullContext> FULL_CONTEXT_CACHE = new ConcurrentHashMap<>();
```
rth-axelor: Three problems stacked:

  1. Unbounded memory leak. The map is static, keyed by entity instances, and never evicted. Every CallTenderNeed/CallTenderOffer ever passed through this report stays referenced for the JVM
  lifetime — pinning detached Hibernate entities (and their object graphs) in the heap forever.
  2. Cross-request staleness. Axelor Model.equals/hashCode is id-based. Two different requests (different users, different transactions) loading the same CallTenderNeed id will hit a cached
  FullContext built from an entity belonging to an earlier, now-detached session → stale custom-field values served to other users.
  3. Negligible benefit. FullContextHelper.create is a thin wrapper; caching it across requests buys little and isn't worth the risk.

  Recommendation: remove the static cache. If you want to avoid re-wrapping the same entity within a single getReportData call, use a local Map created inside that method and passed down (or
  just wrap inline). Don't hold entities in static state.

axelor-purchase/src/main/java/com/axelor/apps/purchase/service/CallTenderReportServiceImpl.java
```diff
@@ -0,0 +1,505 @@
...
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("dd/MM/yyyy");
```
rth-axelor: Change it using the default methods in AOS for date format please (DateService.getDateFormat() / getDateTimeFormat())

axelor-purchase/src/main/java/com/axelor/apps/purchase/service/CallTenderReportServiceImpl.java
```diff
@@ -0,0 +1,505 @@
...
+        ObjectUtils.notEmpty(offer) && ObjectUtils.notEmpty(offer.getProposedUnit())
+            ? offer.getProposedUnit().getName()
+            : null);
+    BigDecimal proposedQty = ObjectUtils.notEmpty(offer) ? offer.getProposedQty() : null;
+    row.put(
+        "proposed_qty",
+        (ObjectUtils.notEmpty(proposedQty) ? proposedQty : BigDecimal.ZERO)
+            .setScale(appBaseService.getNbDecimalDigitForQty(), RoundingMode.HALF_UP));
+
+    LocalDate propDate = ObjectUtils.notEmpty(offer) ? offer.getRequestedDate() : null;
```
rth-axelor: Use Optional for it

axelor-purchase/src/main/java/com/axelor/apps/purchase/service/CallTenderReportServiceImpl.java
```diff
@@ -0,0 +1,505 @@
...
+
+  protected Map<String, CallTenderOffer> buildOfferMap(CallTender callTender) {
+
+    if (ObjectUtils.isEmpty(callTender.getCallTenderOfferList())) {
+      return Collections.emptyMap();
+    }
+
+    Map<String, CallTenderOffer> map = new HashMap<>();
+    for (CallTenderOffer offer : callTender.getCallTenderOfferList()) {
+      if (ObjectUtils.notEmpty(offer.getProduct())
```
rth-axelor: For M2O, use != instead of ObjectUtils.notEmpty in all this service

PR #16187, opened 2026-05-05
Anomaly RM#112056:Make all addControlEntryBtn use the same java process

vva-axelor:


changelogs/unreleased/112056.yml
```diff
@@ -0,0 +1,3 @@
+---
+title: "Control entry: unified the java process used by all add control entry buttons."
+module: axelor-quality
```
rth-axelor, line 3: Can you please explain the changes in the developer part of the changelog and give the migration script for all action view removal

PR #16194, opened 2026-05-06
Anomaly RM#112117: STOCKMOVE : Fix LazyInitializationException on Com…

cma-axelor:
…pany in conformity check and stop masking technical errors as conformity errors

### DIFF
--- a/axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
+++ b/axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
@@ -100,6 +100,7 @@ public class StockMoveLineServiceImpl implements StockMoveLineService {
   protected StockLocationLineFetchService stockLocationLineFetchService;
   protected TrackingNumberCreateService trackingNumberCreateService;
   protected StockMoveService stockMoveService;
+  protected StockMoveRepository stockMoveRepository;
 
   @Inject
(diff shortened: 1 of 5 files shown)


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -1113,6 +1113,9 @@ public void checkConformitySelection(StockMove stockMove) throws AxelorException
 
     final Set<String> productsWithErrors = new HashSet<>();
 
+    final long smId = stockMove.getId();
+    final StockMove[] smHolder = {Beans.get(StockMoveRepository.class).find(smId)};
```
cce-axelor: Can you please inject the repo in constructor ? and add the change in changelog, thanks

axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveLineServiceImpl.java
```diff
@@ -1128,17 +1131,19 @@ public void checkConformitySelection(StockMove stockMove) throws AxelorException
         sml -> {
           final Product product = sml.getProduct();
           Object type =
```
cce-axelor: You can add StockMove stockMove = stockMoveRepository.find(smId); here, it's more easy to read.

PR #16281, opened 2026-05-21
Anomaly RM#111488:Wrong title for product model tab

vva-axelor:


axelor-base/src/main/resources/views/Menu.xml
cce-axelor, line 314: Can't you use forceTitle in action view to avoid view duplication ?

PR #16311, opened 2026-06-02
Anomaly RM#113322: BUDGET : committed amount not updated if a budget …

bsu-axelor:
…distribution line is deleted

### DIFF
--- a/axelor-budget/src/main/java/com/axelor/apps/budget/service/purchaseorder/PurchaseOrderBudgetServiceImpl.java
+++ b/axelor-budget/src/main/java/com/axelor/apps/budget/service/purchaseorder/PurchaseOrderBudgetServiceImpl.java
@@ -43,6 +43,7 @@ import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


axelor-purchase/src/main/resources/views/PurchaseOrder.xml
```diff
@@ -358,7 +358,7 @@
         onClick="save,action-purchase-order-method-draft,save" icon="reply-all-fill"
         readonlyIf="!purchaseOrderLineList || purchaseOrderLineList.length == 0"/>
       <button name="newVersionBtn" title="New version" showIf="statusSelect == 2"
-        onClick="save,action-purchase-record-newVersion,save"
+        onClick="action-budget-purchase-order-group-new-version-btn"
```
rth-axelor: Hello, move this action in the budget module please. And in the new action group, call the action-purchase-record-newVersion and then the budget action if budget app is installed instead of move all the process in budget. If you have an application with only purchase, this will not make anything.

axelor-budget/src/main/java/com/axelor/apps/budget/service/purchaseorder/PurchaseOrderBudgetServiceImpl.java
```diff
@@ -337,4 +338,36 @@ public void fillBudgetStrOnLine(PurchaseOrder purchaseOrder) {
...
+    }
+    for (PurchaseOrderLine poLine : purchaseOrder.getPurchaseOrderLineList()) {
+      if (CollectionUtils.isNotEmpty(poLine.getBudgetDistributionList())) {
+        poLine.getBudgetDistributionList().forEach(bd -> budgets.add(bd.getBudget()));
+      }
+    }
+    return budgets;
+  }
+
+  protected void recomputeBudgets(Set<Budget> budgets) {
```
rth-axelor, line 356: call this in updateBudgetLinesFromPurchaseOrder too please

PR #16323, opened 2026-06-04
Anomaly RM#113515:STOCKMOVE : LazyException on realizing a stock move…

vva-axelor:
… with realStockMoveAutomaticMail enabled


axelor-stock/src/main/java/com/axelor/apps/stock/service/StockMoveServiceImpl.java
```diff
@@ -663,7 +667,8 @@ protected void sendMailForStockMove(StockMove stockMove, Template template)
           stockMove);
     }
     try {
-      Beans.get(TemplateMessageService.class).generateAndSendMessage(stockMove, template);
+      Beans.get(TemplateMessageService.class)
+          .generateAndSendMessage(stockMove, templateRepository.find(template.getId()));
```
cce-axelor: Please use JPA.find() to avoid inject repository and limit impacts

PR #16371, opened 2026-06-15
Feature RM#113861:New menu entry in order to manage ContractVersion

vva-axelor:


axelor-contract/src/main/resources/i18n/messages_fr.csv
```diff
@@ -74,6 +77,7 @@
 "Contract information","Informations du contrat",,
 "Contract invoice","Facture de contrat",,
 "Contract line","Ligne de contrat",,
+"Contract name",,,
```
cce-axelor: Cna you please add translations ? 
Contract ID => Id. contrat
Contract name => Nom contrat

PR #16465, opened 2026-06-30
Anomaly RM#114947:INVOICE : internal server error 'Move cannot be emp…

vva-axelor:
…ty' when recording a payment on a foreign currency invoice

### DIFF
--- a/axelor-account/src/main/java/com/axelor/apps/account/service/reconcile/ReconcileServiceImpl.java
+++ b/axelor-account/src/main/java/com/axelor/apps/account/service/reconcile/ReconcileServiceImpl.java
@@ -22,6 +22,7 @@ import com.axelor.apps.account.db.Account;
 import com.axelor.apps.account.db.AccountConfig;
 import com.axelor.apps.account.db.Invoice;
 import com.axelor.apps.account.db.InvoicePayment;
+import com.axelor.apps.account.db.InvoiceTerm;
 import com.axelor.apps.account.db.Journal;
 import com.axelor.apps.account.db.Move;
 import com.axelor.apps.account.db.MoveLine;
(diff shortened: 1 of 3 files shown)


axelor-account/src/main/java/com/axelor/apps/account/service/reconcile/ReconcileServiceImpl.java
```diff
@@ -563,6 +564,18 @@ public void balanceCredit(MoveLine creditMoveLine) throws AxelorException {
           MoveLine debitAdjustmentMoveLine =
               moveAdjustementService.createAdjustmentMove(creditMoveLine, creditAccount);
 
+          if (debitAdjustmentMoveLine == null) {
```
rth-axelor, line 568: Create a protected method for credit and debit side please, since it's the same code

axelor-account/src/main/java/com/axelor/apps/account/service/move/MoveLineInvoiceTermServiceImpl.java
```diff
@@ -102,7 +102,10 @@ public void generateDefaultInvoiceTerm(
     if (move == null
         || moveLine == null
         || moveLine.getAccount() == null
-        || !moveLine.getAccount().getUseForPartnerBalance()) {
+        || !moveLine.getAccount().getUseForPartnerBalance()
+        || (move.getInvoice() != null
+            && move.getInvoice().getPartnerAccount() != null
+            && !moveLine.getAccount().equals(move.getInvoice().getPartnerAccount()))) {
```
rth-axelor: For which case do we need this ? This method is used eveywhere and if we don't modify this method it should be better

PR #16582, opened 2026-08-07
Feature RM#116623: [UI] Allow opening the invoice from the move line …

bsu-axelor:
…origin only when move.invoice is set


axelor-account/src/main/resources/views/MoveLine.xml
```diff
@@ -676,6 +678,8 @@
     onLoad="action-move-line-method-onload" canArchive="false">
     <toolbar>
       <button name="viewMoveBtn" title="View entry" onClick="action-move-line-view-open-move"/>
+      <button name="viewInvoiceBtn" title="View invoice" showIf="move.invoice"
```
rth-axelor: Can you please add it after the viewMoveBtn in the movePanel please, like a spacer of colspan 6 and the button

PR #16595, opened 2026-08-11
Anomaly RM#116685: PRODUCTION : Fix tracking number recovery after ma…

cma-axelor:
…nual deletion

### DIFF
--- a/axelor-production/src/main/java/com/axelor/apps/production/service/ProductionTrackingPreservationServiceImpl.java
+++ b/axelor-production/src/main/java/com/axelor/apps/production/service/ProductionTrackingPreservationServiceImpl.java
@@ -24,6 +24,7 @@ import com.axelor.apps.base.db.Unit;
 import com.axelor.apps.base.service.ProductCompanyService;
 import com.axelor.apps.base.service.UnitConversionService;
 import com.axelor.apps.base.service.app.AppBaseService;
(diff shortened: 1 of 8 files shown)


axelor-production/src/main/java/com/axelor/apps/production/service/ProductionTrackingPreservationServiceImpl.java
```diff
@@ -351,6 +352,62 @@ public void drainRemainingPreservedTracking(
     }
   }
 
+  @Override
+  public PreservedTrackingNumbersByProduct reclaimOrphanedTrackingNumbers(
```
rth-axelor, line 356: One perf point on reclaimOrphanedTrackingNumbers: with many products to produce it does more work than needed, and it runs on every regeneration (createNewProducedStockMoveLineList + partialFinish):

  1. One query per product — a single query on originManufOrder + originStockMoveLine IS NULL, grouped by product in memory, is enough (no IN needed). → N queries down to 1.
  2. getReclaimQty recomputed per tracking number — it only depends on (prodProduct, manufOrder) and does a productCompanyService.get(...) each time. Memoize it per product. → once-per-orphan down to once-per-product.

  This would let findOrphanedByProductAndOriginManufOrder be replaced by a findOrphanedByOriginManufOrder finder (check it isn't used elsewhere).

PR #16608, opened 2026-08-14
Anomaly RM#116784: STOCK MOVE : Fix scheduled outgoing stock move inv…

cma-axelor:
…oicing


axelor-sale/src/main/java/com/axelor/apps/sale/service/saleorder/SaleOrderCreateServiceImpl.java
rth-axelor, line 180: You will get this npe here too, but need to get an exception I think for it, if company == null && salespersonUser.getActiveCompany is null

cma-axelor: In this execution path, the company is validated by controlSaleOrdersToMerge before generateSaleOrder is called. If the common company is null, checkErrors throws an AxelorException, so the reviewed company
 fallback is not reached with a null company. The reproduced case has a valid company but no authenticated user because it runs from a scheduled batch; the NPE occurs only while resolving the team. I would  therefore kept the fix scoped to the nullable salesperson/team case.

PR #16648, opened 2026-08-26
Anomaly RM#117276:Map: 'Customers/Prospects/Suppliers on the map' bro…

vva-axelor:
…ken - GET /ws/map/{object} returns 404 (JSON.parse error) after MapRest removal (RM#107895)


rth-axelor:
**Review — RM#117276 (Map REST endpoints restored)**

Note: mostly restored legacy code, so several points are inherited debt rather than regressions.

## 🔴 Critical

1. **Wrong JPQL operator precedence** in `getPartners` (`MapRest.java`): `"self.isCustomer = true OR self.isSupplier = true AND self.isContact=?"` is read as `isCustomer OR (isSupplier AND isContact=false)` → customer contacts are still returned, and the `archived` filter (present in `getCustomers`/`getProspects`) is missing. Parenthesize the `OR` and add the archived filter.
2. **No access control** on any endpoint — `/partner`, `/customer`, `/supplier`, `/lead`, `/opportunity`, `/geomap/turnover` expose names/phones/emails and turnover-per-country to any authenticated user, with no company filtering. Please confirm this matches the intended (pre-107895) behavior.
3. **Side effects inside GET requests**: `makeAddressString` → `AddressService.getOrUpdateLatLong` writes to DB and calls the geocoding API. `getOpportunities`/`getLeads` trigger this in a loop without checking `isValidLatLong` → slow GETs, per-row DB writes and rate-limited API calls.

## 🟠 Important

4. **Unguarded NPEs**: `getGeoMapData` (`so.getMainInvoicingAddress()` may be null — and this method has no try/catch → raw 500); `getTour` (`tourRepo.find(id)` null, `tourLine.getAddress()` null).
5. **`getGeoMapData` breaks the pattern** — the only endpoint without `try/catch` + `setData/setError`; it builds the JSON manually. Please align it.
6. **Full-table load + N+1**: `partnerRepo.all().fetch()` / `leadRepo.all().fetch()` fetch entire tables with no pagination, then `getInvoicingAddress` runs one query per record.
7. **Hardcoded / non-i18n labels**: `"Amount"`, `"Country"`, `"Turnover"`; the `if (!isNullOrEmpty(I18n.get("amount")))` guard is always true (`I18n.get` returns the key when missing), so it is dead code.

vva-axelor:
> 2. **No access control** on any endpoint — `/partner`, `/customer`, `/supplier`, `/lead`, `/opportunity`, `/geomap/turnover` expose names/phones/emails and turnover-per-country to any authenticated user, with no company filtering. Please confirm this matches the intended (pre-107895) behavior.

Confirmed - this matches the pre-107895 behavior exactly

PR #16709, opened 2026-09-10
Anomaly RM#118211:TICKET : display a useful message for too long subject

vva-axelor:


axelor-helpdesk/src/main/resources/domains/Ticket.xml
```diff
@@ -8,7 +8,7 @@
   <entity name="Ticket">
 
     <string name="ticketSeq" title="Ticket Number" readonly="true"/>
-    <string name="subject" title="Subject" required="true"/>
+    <string name="subject" title="Subject" required="true" max="255"/>
     <string name="fullName" title="Full name" namecolumn="true" search="ticketSeq,subject">
```
rth-axelor, line 12: Check this case too, since a subject of 255 and a ticket seq of 5 will raise the same error on the fullName

